PR #14, opened 2022-04-01
add wallets

mrval5:
### What does this do?
We added more wallets: WalletConnect, Coinbase, Fortmatic, Portis

### How do I test this?
Clicking on connect button


AdamDotCom:
Nice! That was quick. 

Presently when we click on the MetaMask option is opens up MetaMask and prompts for credentials. Can we do something similar for the Walletconnect, Coinbase, Fortmatic, and Portis? Our Wilder World Market app has the functionality we're looking for (click the 'Connect' button at https://app.wilderworld.com/market/).

AdamDotCom:
We're making use of [web3-react](https://github.com/NoahZinsmeister/web3-react) package. I suspect we'll need to add additional configuration values for the new providers in this file: https://github.com/zer0-os/zOS/blob/49db8361ea30991833d129623096c8ab0206cd8c/src/lib/web3/web3-react.tsx#L42-L42

.github/workflows/run-tests.yml
```diff
@@ -4,6 +4,8 @@ on:
     branches-ignore:
       - main
       - development
+env:
+  REACT_APP_INFURA_URL: https://kovan.infura.io/v3/number123
```
mrval5, line 8: this line is added to fix the unit test.

the lib is expecting an url:
https://github.com/NoahZinsmeister/web3-react/blob/v6/packages/network-connector/src/index.ts#L32

src/lib/web3/connectors.ts
```diff
@@ -0,0 +1,38 @@
...
+  appName: 'zOS',
+});
+
+export const fortmatic = new FortmaticConnector({
+  chainId: Chains.Kovan,
+  apiKey: 'api_key',
+});
+
+export const portis = new PortisConnector({
+  dAppId: config.rootDomainId,
```
mrval5, line 32: @AdamDotCom I'm not sure about the value of `dAppId `

AdamDotCom:
This looks great. Were you able to manually verify that the new wallet connections? We don't have a dedicated QA team so the verification testing falls on us.

AdamDotCom (approved):

mrval5:
> This looks great. Were you able to manually verify that the new wallet connections? We don't have a dedicated QA team so the verification testing falls on us.

I'm able to connect with wallets

AdamDotCom (approved):

PR #15, opened 2022-04-05
App menu

AdamDotCom:
Related PR: https://github.com/zer0-os/zOS-feed/pull/2

## Screen Capture

zOS-feed


src/store/apps/index.ts
```diff
@@ -0,0 +1,35 @@
...
+const initialState: AppsState = {
+  selectedApp: config.defaultApp,
+};
+
+const slice = createSlice({
+  name: 'apps',
+  initialState,
+  reducers: {
+    receive: (state, action: PayloadAction<string>) => {
+      state.selectedApp = action.payload;
```
AdamDotCom, line 28: I'm conflicted. The zns state containing the route also seems like an appropriate place to store the selected app. It seems unlikely that we'll access the ZNS route without the app. What are your thoughts?

src/zns-route-connect.tsx
```diff
@@ -19,24 +21,34 @@ export class Container extends React.Component<Properties> {
   }
 
   static mapActions(_props: Properties): Partial<Properties> {
-    return { setRoute };
+    return { setRoute, setSelectedApp };
   }
 
   componentDidMount() {
     this.props.setRoute(this.extractRouteFromProps());
+    this.props.setSelectedApp(this.extractAppFromProps());
```
AdamDotCom, line 29: Is this double saga call a performance issue?

src/components/address-bar/container.tsx
```diff
@@ -14,13 +14,14 @@ export interface Properties extends PublicProperties {
   history: History;
   route: string;
   deepestVisitedRoute: string;
+  app: string;
```
mrval5, line 17: If I'm not mistaken the type should be `Apps`

AdamDotCom: Thanks!

mrval5 (dismissed):

PR #25, opened 2022-04-11
Feature/wallet manager hook

mrval5:
### What does this do?
We changed the component from class based to hook
- extending from the class means adding more code in the build
- the lifecycle hooks from the class can be deprecated soon
- the community and React are encouraging to use hooks over class
- most of libraries are using hooks

- testing with Enzyme does not support `useEffect` method, which we have added a new package `jest-react-hooks-shallow`
- We have added also testing-library to help on testing
- we removed some of the unit test related to the component `WalletSelectModal` because this repo is not responsible of testing components from another lib

### Why the PR in draft
- we should merge https://github.com/zer0-os/zOS-component-library/pull/5, and bump it here


src/components/wallet-manager/index.test.tsx
```diff
@@ -162,46 +133,4 @@ describe('WalletManager', () => {
 
     expect(updateConnector).toHaveBeenCalledWith(Connectors.Metamask);
   });
-
-  describe('mapState', () => {
```
AdamDotCom, line 166: Where should the state / reducer / action related test coverage be moved to? Should they be move to the ../store directory or would they stay in this file?

src/components/wallet-manager/index.tsx
```diff
@@ -1,116 +1,132 @@
...
-}
-
-export const WalletManager = connectContainer<{}>(Container);
+        </div>
+      )}
+    </div>
+  );
+};
+
+const mapStateToProps = (state: RootState): Partial<Props> => {
```
AdamDotCom, line 110: What's the best way to test state mapping? It looks like [redux-mock-store](https://github.com/reduxjs/redux-mock-store) was an option. Is there a more modern alternative?

src/components/wallet-manager/index.test.tsx
```diff
@@ -81,41 +93,25 @@ describe('WalletManager', () => {
...
-    // assert pre-condition
-    expect(wrapper.find(WalletSelectModal).prop('isConnecting')).toBe(true);
-
-    wrapper.setProps({ connectionStatus: ConnectionStatus.Connected });
-
-    // re-open modal, as it will be closed at this point
-    wrapper.find('.wallet-manager__connect-button').simulate('click');
-
-    expect(wrapper.find(WalletSelectModal).prop('isConnecting')).toBe(false);
+    await waitFor(() => {
```
AdamDotCom, line 111: Cool, I've never encountered `waitFor` before.

AdamDotCom:
@mrval5 how should we attempt to test the state and actions? This article gives a couple options, but the article is two years old: https://jsramblings.com/3-ways-to-test-mapstatetoprops-and-mapdispatchtoprops/
This article demonstrates testing reducers: https://redux.js.org/usage/writing-tests Is there a better way? It'd be nice to see coverage for the state and actions.

src/components/wallet-manager/index.test.tsx
mrval5: Ideally we should test:
- action creators in store folder (unit test)
- reducers in store folder (unit test)
- testing sagas in store folder (unit test)

and if we want we can test the integrations between the component and the store.

jtulloch (changes requested):
hey! welcome @mrval5. hope you're settling in alright! i know this is a draft, so apologies if any comments are not relevant. feel free to ignore any that are! looking forward to meeting you soon.

jtulloch:
closing this for now until we get a chance to come back to it.

PR #32, opened 2022-04-21
day light mode enhancement

mrval5:
#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/address-bar/styles.scss
+++ b/src/components/address-bar/styles.scss
@@ -1,15 +1,16 @@
 @use '../../shared-components/theme-engine/theme' as theme;
 
 .address-bar {
-  background-color: theme.$background-color-primary;
+  background-color: theme.$address-bar-background-color-outer;
   border: 1px solid theme.$address-bar-border-color-outer;
   border-radius: 24px;
   padding: 3px;
-
+  
   display: flex;
   align-items: center;
-
+  
   &__inner {
+    background-color: theme.$address-bar-background-color-inner;
     border: 1px solid theme.$address-bar-border-color-inner;
     border-radius: 24px;
 
@@ -20,7 +21,7 @@
   
   &__protocol,
   &__route {
-    color: theme.$font-color-tertiary;
+    color: theme.$font-color-primary;
   }
 
   &__navigation {
--- a/src/shared-components/theme-engine/_theme.scss
+++ b/src/shared-components/theme-engine/_theme.scss
@@ -21,6 +21,9 @@ $secondary-border-color: var(--secondary-border-color, #3C3C3C);
 $address-bar-border-color-outer: var(--address-bar-border-color-outer, #23142F);
 $address-bar-border-color-inner: var(--address-bar-border-color-inner, #472E5B);
 
+$address-bar-background-color-outer: var(--address-bar-background-color-outer, #0A0A0A);
+$address-bar-background-color-inner: var(--address-bar-background-color-inner, #1A1022);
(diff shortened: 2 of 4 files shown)


src/components/address-bar/styles.scss
```diff
@@ -1,15 +1,16 @@
 @use '../../shared-components/theme-engine/theme' as theme;
 
 .address-bar {
-  background-color: theme.$background-color-primary;
+  background-color: theme.$address-bar-background-color-outer;
```
jtulloch, line 4: i know it's not always possible, but let's try to avoid having theme values tied to the app structure like this. if we need to name the colors in a more descriptive way than primary/secondary, we can do so, but the more the theme is tied to the actual implementation of the app, the harder it is to change things. both for us, and the designers.

if the designs are not organizing the colors well, we can definitely push back on them, and ask them to be a bit more consistent. it is better for everybody if the palette we are using is as small as possible, and is as consistent across the app as possible. eg. if they're using more than 3 font colors (aside from logos/branding, etc.), we should check in with them, as that will start to get pretty unwieldy. if they really want to use more, then we should at least work with them to define what the role of those additional colors will be.

src/shared-components/theme-engine/theme.json
```diff
@@ -29,6 +31,9 @@
     "fontColorTertiary": "#FFFFFF",
     "fontColorTertiaryHover": "#F5F5F5",
 
+    "fontColorAppMenu": "background: linear-gradient(0deg, #404040, #404040),linear-gradient(0deg, #808080, #808080)",
```
jtulloch, line 34: i find that it's easier with a theme to restrict it to css *values* rather than full statements. it becomes particularly important when you start allowing users/entities to customize the theme, but it also just makes things more flexible. if we can, let's actually try to restrict the theme to colors for now, and if we need gradients, etc., keep that in the css. (i know the existing theme might have some legacy stuff that breaks this, since we just started with a theme from the old app to get up and running, but since we're starting pretty fresh, let's try to avoid it if possible.)

src/shared-components/theme-engine/theme.json
```diff
@@ -88,6 +95,9 @@
     "fontColorTertiary": "#959595",
     "fontColorTertiaryHover": "#000000",
 
+    "fontColorAppMenu": "#E6E6E6",
+    "filterAppMenuImage": "brightness(100%)",
```
jtulloch, line 99: as with the above comment, i think filters, etc., don't need to be in the theme.

jtulloch (changes requested):

src/shared-components/theme-engine/_theme.scss
```diff
@@ -52,3 +52,6 @@ $underlay-background: var(--underlay-background, rgba(0,0,0,0.95));
 $glow-text-color: var(--glow-text-color, hsla(0,0%,100%,0.7));
 $glow-background-color: var(--glow-background-color, #6e10bb);
 $glow-transition-background-color: var(--glow-transition-background-color, #e0baff);
+$glow-transition-background-color: var(--glow-transition-background-color, #e0baff);
```
jtulloch, line 55: is this line actually duplicated, or just a merge artifact?

jtulloch (approved):

src/shared-components/theme-engine/_theme.scss
mrval5: nice catch, I've fixed it

PR #37, opened 2022-05-04
Fix close wallet connect modal after connection

hkhanouchi:
**### Actual** :

We get an error only when we connect to wallet (Metamask & Wallet Connect), this error blocks the closing of modal.

Web3React throws the error UnsupportedChainIdError, because the Chain 42 passed to the supportedChainIds param it's not supported.

**### Solution** :

The solution was to add another chain (MainNet = 1 ) to get : supportedChainIds: [Chains.Kovan, Chains.MainNet]


hkhanouchi:
**Actual :**

We get an error only when we connect  to wallet (Metamask & Wallet Connect), this error blocks the closing of modal.

Web3React throws the error UnsupportedChainIdError,  because the Chain 42 passed to the supportedChainIds param it's not supported.

**Solution :**

The solution was to add another chain (MainNet = 1 ) to get  :  supportedChainIds: [Chains.Kovan, Chains.MainNet]

mrval5 (approved):

PR #45, opened 2022-05-06
release osx app

mrval5:
### What does this do?
adding job for electron osx release, check on actions : https://github.com/zer0-os/zOS/runs/6321392819?check_suite_focus=true

#### Things to consider:
  1. Does this change any APIs?


jtulloch:
@mrval5 does this split where it is stored and/or name of the resulting app based on environment? if so, great! if not, i think it's worth nailing that down as part of this, so we don't run into conflicts when we do the production build.

jtulloch (approved):

PR #49, opened 2022-05-12
add public_url in env

mrval5:
### What does this do?
Fix white screen.

We get a white screen when we visit: https://optimistic-kare-3090a1.netlify.app/wilder/feed

The problem is the static files has the wrong url
Request URL: https://optimistic-kare-3090a1.netlify.app/wilder/static/js/main.0078f803.chunk.js

The fix: add PUBLIC_URL to env


mrval5:
the problem was in homepage (package.json)

PR #51, opened 2022-05-16
change address user after account switch

hkhanouchi:
#### Things to consider:
  1. Does this change any APIs?


src/components/web3-connect/index.tsx
```diff
@@ -110,6 +111,10 @@ export class Container extends React.Component<Properties, State> {
       this.syncGlobalsForConnectedStatus();
     }
 
+    if (web3.account && ( !previouslyAccount || ( web3.account !== previouslyAccount ))) {
+      this.syncGlobalsForConnectedStatus();
```
jtulloch, line 115: my main concern is that `syncGlobalsForConnectedStatus` may get called twice on initial connection, as well as if the user switches wallets. we should test that to make sure, and if it does, we should try to set this up so that it only gets called once in those situations.

jtulloch (changes requested):

jtulloch (approved):

PR #65, opened 2022-06-01
Add underlay when click search bar

hkhanouchi:
#### Things to consider:
  1. Does this change any APIs?


src/app-sandbox/index.tsx
```diff
@@ -9,23 +10,52 @@ export interface Properties {
...
+      </div>
+    );
   }
 
   render() {
     return (
-      <div className="app-sandbox">
+      <div className='app-sandbox'>
         {this.renderSelectedApp()}
+        {this.props.hasOverlay &&
```
AdamDotCom, line 56: Is it possible to use the  `isOpen` var in place of the `hasOverlay` var or vice versa? What scenario could we have an overlay that is not opened?

jtulloch (approved):

PR #94, opened 2022-06-22
add prettier

mrval5:
### What does this do?
- add prettier
- add pre commit hook to pretty only the staged files
- add pre push hook to validate code format
- pretty all files
Problem: each one of us has and prefer a style guide and when we push we find out a lot of change only because of indentation
solution: we should not car any more about indentation because we have pre commit hook who fix all the indention with the standard style.

#### Things to consider:
  1. Does this change any APIs?


src/store/apps/index.ts
```diff
@@ -1,8 +1,4 @@
-import {
-  createSlice,
-  createAction,
-  PayloadAction,
-} from '@reduxjs/toolkit';
+import { createSlice, createAction, PayloadAction } from '@reduxjs/toolkit';
```
jtulloch, line 1: as with arrays, i find import statements are easier to manage if each element is on a separate line.

src/components/app-menu/container.tsx
```diff
@@ -12,7 +12,14 @@ export interface Properties {
 
 export class Container extends React.Component<Properties, {}> {
   static mapState(state: RootState): Partial<Properties> {
-    const { zns: { value: { route } }, apps: { selectedApp: { type } } } = state;
```
AdamDotCom, line 15: I kind of like the single line destructuring version... but my eyes will readjust over time.

src/components/address-bar/util.tsx
```diff
@@ -1,3 +1,8 @@
 export const routeWithApp = (route, app) => {
-  return `/${[route, app].filter(Boolean).join('/')}`;
```
AdamDotCom, line 2: The initial version seems more readable.

src/components/autocomplete-dropdown/autocomplete-dropdown.test.tsx
```diff
@@ -65,10 +65,20 @@ describe('autocomplete-dropdown', () => {
   });
 
   it('it renders match suggestions', async () => {
-    findMatches = () => { return [
-      { id: 'result-first-id', value: 'result-first-value', route: 'result-first-route' },
-      { id: 'result-second-id', value: 'result-second-value', route: 'result-second-route' },
-    ]};
```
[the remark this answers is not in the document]
jtulloch: haha. yeah. .. i think with nested objects like this, it's less important that the inner objects are on separate lines, as you're mostly just going to be adding/removing single objects, rather than touching properties in every one of them. wonder if it can be based on number of properties.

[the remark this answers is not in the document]
jtulloch: @mrval5 is it possible to have it based on number of properties? eg. .. if 3 or less, allow single line, otherwise make it multiple lines?

src/components/autocomplete-dropdown/index.tsx
```diff
@@ -295,30 +272,18 @@ export class AutocompleteDropdown extends React.Component<Properties, State> {
   results(items, focusedItem) {
     let content;
     if (!items.length) {
-      content = (
-        <div className='autocomplete-dropdown-item autocomplete-dropdown__no-results'>
-          No results found
-        </div>
-      );
+      content = <div className='autocomplete-dropdown-item autocomplete-dropdown__no-results'>No results found</div>;
```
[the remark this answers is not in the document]
mrval5: I've added a new rule to make attrs in each line in case we have more than 1 attr

src/components/wallet-manager/index.tsx
```diff
@@ -85,12 +86,7 @@ export class Container extends React.Component<Properties, State> {
 
   get availableWallets(): WalletType[] {
     if (isElectron()) {
-      return [
-        WalletType.WalletConnect,
-        WalletType.Coinbase,
-        WalletType.Fortmatic,
-        WalletType.Portis,
-      ];  
+      return [WalletType.WalletConnect, WalletType.Coinbase, WalletType.Fortmatic, WalletType.Portis];
```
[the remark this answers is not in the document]
mrval5: that was from the rule `printWidth`.
I've added a plugin to make it in multiple lines

src/store/apps/index.ts
mrval5: prettier doesn't have a solution for it, we should find a plugin for it

src/components/address-bar/util.tsx
mrval5: because we introduced a plugin to make it array items in multiple lines

jtulloch: yeah. i think this one could be cleaned up just by not having it all inline in a string, probably.

src/store/apps/index.ts
jtulloch: no. let's just do the minimum with this PR, and get it mergted. this stuff is all auto-fixable, so let's just add/adjust the rules as we go.

PR #105, opened 2022-06-28
Feature/search animation

mrval5:
#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/autocomplete-dropdown/index.tsx
+++ b/src/components/autocomplete-dropdown/index.tsx
@@ -38,9 +38,9 @@ interface State {
 
 export class AutocompleteDropdown extends React.Component<Properties, State> {
   performSearch;
+  itemsElement;
   private _isMounted: boolean = false;
   private anchorElement: HTMLElement;
-  private itemsElement: React.RefObject<HTMLInputElement>;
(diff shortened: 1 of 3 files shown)


src/components/autocomplete-dropdown/index.tsx
```diff
@@ -72,6 +77,14 @@ export class AutocompleteDropdown extends React.Component<Properties, State> {
     }
   }
 
+  componentDidUpdate(prevProps, prevState) {
+    if (prevState.matches?.length !== this.state.matches?.length) {
+      this.setState({
+        dropdownHeight: this.itemsElement?.current?.getBoundingClientRect().height,
```
jtulloch, line 83: can we try to avoid using the `?.` operator in actual code? it does have some legitimate uses, but in general i find that it can gloss over some underlying problems, and can make them harder to diagnose. almost always, if an object can be null, it should be handled explicitly. preferably by making sure it *isn't* ever null, or in a case like this where null is reasonable/expected, adding an extra line or two that makes it clear what is actually happening when the height isn't set.

jtulloch (changes requested):

jtulloch (approved):

PR #106, opened 2022-06-30
Reconnect connector on reload page

hkhanouchi:
#### Things to consider:

1 - We need await after we activate connector in test.

### DIFF
--- a/src/components/web3-connect/index.tsx
+++ b/src/components/web3-connect/index.tsx
@@ -69,6 +69,22 @@ export class Container extends React.Component<Properties, State> {
     this.props.setConnectionStatus(ConnectionStatus.Disconnected);
   }
 
+  deactivateConnector() {
+    if (localStorage) {
+      localStorage.removeItem('previousConnector');
+    }
+    this.props.updateConnector(Connectors.Infura);
+    this.props.setAddress('');
+  }
+
+  reconnectPreviousConnector() {
+    const { currentConnector } = this.props;
+    if (localStorage.getItem('previousConnector') && currentConnector === Connectors.Infura) {
+      const previousConnector = localStorage.getItem('previousConnector');
+      this.props.updateConnector(Connectors[previousConnector.charAt(0).toUpperCase() + previousConnector.slice(1)]);
+    }
+  }
+
   async activateCurrentConnector() {
     const { web3, connectors, currentConnector } = this.props;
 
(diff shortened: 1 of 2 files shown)


src/components/web3-connect/index.tsx
```diff
@@ -122,8 +141,13 @@ export class Container extends React.Component<Properties, State> {
       this.syncGlobalsForConnectedStatus();
     }
 
+    if (!web3.account && web3.account !== previouslyAccount && !web3.active && !web3.library) {
```
jtulloch, line 144: can we just base this on the connection? do we need all of the other checks? eg. i think if the status goes from connected -> disconnected, we should reconnect infura. not sure if that just means checking web3.active, or if it'll be a bit more involved.

src/components/web3-connect/index.tsx
```diff
@@ -69,6 +69,22 @@ export class Container extends React.Component<Properties, State> {
     this.props.setConnectionStatus(ConnectionStatus.Disconnected);
   }
 
+  deactivateConnector() {
+    if (localStorage) {
+      localStorage.removeItem('previousConnector');
+    }
+    this.props.updateConnector(Connectors.Infura);
+    this.props.setAddress('');
```
jtulloch, line 77: i think we can simplify this. can we just pull the `setAddress` call out, and base it on account change?
eg. i think setAddress can always just happen like this:
```
    if (web3.account !== prevAccount) {
      this.props.setAddress(web3.account);
    }
```
that way we're not missing any calls by trying to call it in every place we think it might be needed.

jtulloch (changes requested):
sorry, i thought i had submitted this the other day, but it was still sitting in draft.

src/components/web3-connect/index.tsx
hkhanouchi: We can't just base on the connection, cuz the status here doesn't goes from connected to disconnected, and to guaranteed that the wallet was disconnected we should checking the web3.account .

src/components/web3-connect/index.tsx
```diff
@@ -84,6 +100,9 @@ export class Container extends React.Component<Properties, State> {
 
     try {
       await web3.activate(connector, null, true);
+      if (currentConnector && currentConnector !== Connectors.Infura && localStorage) {
```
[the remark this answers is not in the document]
hkhanouchi: I didn't understood exactly what did you mean it by that, but i changed the place of set localStorage to be more significant.

jtulloch (approved):

src/components/web3-connect/index.tsx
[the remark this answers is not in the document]
jtulloch: all i mean is that this component shouldn't be waiting for anything. all it should be doing is responding to changes that come from `web3`.

PR #115, opened 2022-07-07
Release/0.2.8

jtulloch:
#### Things to consider:
  1. Does this change any APIs?


src/components/autocomplete-dropdown/index.tsx
```diff
@@ -146,11 +159,15 @@ export class AutocompleteDropdown extends React.Component<Properties, State> {
     this.close();
   };
 
-  setAnchorElements = (ref: HTMLElement) => {
+  setAnchorElements = (ref: HTMLElement): void => {
     this.anchorElement = ref;
   };
 
-  onKeyDown = (e) => {
+  // setItemsElements = (ref: HTMLElement): void => {
```
AdamDotCom, line 166: Was this commented out block intended to be included?

src/store/zns/index.ts
```diff
@@ -19,8 +18,9 @@ export interface ZnsState {
 
 const initialState: ZnsState = {
   value: {
-    route: config.defaultZnsRoute,
-    deepestVisitedRoute: config.defaultZnsRoute,
```
AdamDotCom, line 23: Do we still need `defaultZnsRoute` in the config file or can we remove it entirely?

src/components/web3-connect/index.test.tsx
```diff
@@ -75,7 +149,7 @@ describe('Web3Connect', () => {
   });
 
   it('registers provider once when active is true', () => {
-    const library = { networkId: 3 };
+    const library = { networkId: Chains.Ropsten };
```
AdamDotCom, line 152: 🪄

AdamDotCom (approved):

src/store/zns/index.ts
jtulloch: it's still used. just not in this spot. this default actually never even ended up getting used anyway.

PR #121, opened 2022-07-15
Reset address barn navigation when switching apps

hkhanouchi:
#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/address-bar/container.tsx
+++ b/src/components/address-bar/container.tsx
@@ -9,7 +9,6 @@ import { routeWithApp } from './util';
 import { client, ZnsClientFactory } from '@zer0-os/zos-zns';
 import { ProviderService, inject as injectProviderService } from '../../lib/web3/provider-service';
 import { PlatformApp } from '../../lib/apps';
-import { setDeepestVisitedRoute } from '../../store/zns';
 interface PublicProperties {
   className?: string;
 }
@@ -23,7 +22,6 @@ export interface Properties extends PublicProperties {
   znsClient: ZnsClientFactory;
 
   app: PlatformApp;
-  setDeepestVisitedRoute: (route: string) => void;
 }
 export class Container extends React.Component<Properties> {
   static mapState(state: RootState): Partial<Properties> {
@@ -38,17 +36,7 @@ export class Container extends React.Component<Properties> {
   }
 
   static mapActions(_props: Properties): Partial<Properties> {
-    return { setDeepestVisitedRoute };
-  }
-
-  componentDidUpdate(prevProps: Properties) {
-    const {
-      app: { type },
-      route,
-    } = this.props;
-    if (prevProps.app.type !== type && prevProps.route !== route) {
-      this.props.setDeepestVisitedRoute(route);
-    }
+    return {};
   }
 
   znsClient() {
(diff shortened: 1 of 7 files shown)


src/components/address-bar/container.tsx
```diff
@@ -36,7 +38,17 @@ export class Container extends React.Component<Properties> {
...
+    return { setDeepestVisitedRoute };
+  }
+
+  componentDidUpdate(prevProps: Properties) {
+    const {
+      app: { type },
+      route,
+    } = this.props;
+    if (prevProps.app.type !== type && prevProps.route !== route) {
+      this.props.setDeepestVisitedRoute(route);
```
jtulloch, line 50: i don't think we should be setting this from this component. because this is essentially part of a state machine around the routing, if we enable setting it from arbitrary components it can get out of sync. there should really only be a single place that this is modified, and the component that is currently responsible for doing that is `src/zns-route-connect.tsx`. i think this task is likely going to require some refactoring of that component, as well as potentially some sagas, to ensure that this state is always mirroring what has happened at the level of the router.

let me know if you're running into issues. i can look a little closer at what this will need, if you'd like, or we can pair on it.

jtulloch (changes requested):

src/zns-route-connect.tsx
```diff
@@ -39,6 +40,7 @@ export class Container extends React.Component<Properties> {
 
     if (selectedApp !== this.extractAppFromProps(prevProps)) {
       this.props.setSelectedApp(selectedApp);
+      this.props.setDeepestVisitedRoute(currentRoute);
```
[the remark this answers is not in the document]
hkhanouchi: Ok, got it

src/zns-route-connect.tsx
```diff
@@ -25,20 +25,23 @@ export class Container extends React.Component<Properties> {
...
+    let routeApp: RouteApp = { route: currentRoute, hasAppChanged: false };
 
     if (selectedApp !== this.extractAppFromProps(prevProps)) {
       this.props.setSelectedApp(selectedApp);
+      routeApp.hasAppChanged = true;
+    }
+
+    if (currentRoute !== this.extractRouteFromProps(prevProps)) {
+      this.props.setRoute(routeApp);
     }
```
jtulloch, line 45: i *think* we probably still want this to be before `setSelectedApp`. this will work totally fine as it is, but my reasoning is that setting the route first will make it less likely that the app will load with the old route, and then have to immediately update. it's possible that having it this way will lead to more requests, or a flicker, etc. depending on how the app is set up.

src/store/zns/index.ts
```diff
@@ -4,7 +4,12 @@ export enum SagaActionTypes {
   UpdateRoute = 'zns/saga/updateRoute',
 }
 
-const setRoute = createAction<string>(SagaActionTypes.UpdateRoute);
+export interface RouteApp {
+  route: string;
+  hasAppChanged: boolean;
+}
```
jtulloch, line 10: i don't *think* we need to export this as a named type. typescript is really good at inferring types, and i find that if you try to strongly type too many things, it can actually be counterproductive. for this kind of thing, where all we're doing is defining a set of parameters that a function takes, we shouldn't need to pass the type around everywhere. let's either just define it inline `createAction<{ route: string; hasAppChanged: boolean }>` or get rid of the export, and just keep the definition local. either is fine, but the more types that get passed around, the more difficult it becomes to untangle things when you go to make a change.

jtulloch (changes requested):

jtulloch:
one more thing -- we need to test the actual change! :)

hkhanouchi:
> one more thing -- we need to test the actual change! :)

What test, :)

jtulloch (changes requested):
can we also add a saga test where we send { hasAppChanged: true }, that makes sure we actually update the state for that scenario?

PR #123, opened 2022-07-18
add connectWallet to app props

mrval5:
#### Things to consider:
  1. Does this change any APIs?


jtulloch (changes requested):
nice. just a couple of comments. i was picturing more of a refactoring, so that the modal isn't owned by the button anymore, but this is good. the button will always be at the root, for now at least, so this is perfect. we can adjust if needed.

jtulloch (approved):

PR #124, opened 2022-07-21
Get channels list from api zero

hkhanouchi:
#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/store/channels-list/api.ts
+++ b/src/store/channels-list/api.ts
@@ -1,13 +1,10 @@
-export const api = {
-  async fetch(_id: string) {
-    return [
-      'stuff',
-      'tacos',
-      'cats',
-      'work',
-    ].map((name, index) => ({
-      id: `channel-id-${index}`,
-      name,
-    }));
-  },
-};
+import { config } from '../../config';
+
+export async function fetchChannels(_id: string) {
+  const data = await fetch(`${config.ZERO_API_URL}/api/networks/${_id}/chatChannels/public`);
+  const channels = await data.json();
+  return channels.map((channel, index) => ({
+    id: `channel-id-${index}`,
+    ...channel,
+  }));
+}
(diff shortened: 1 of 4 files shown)


src/store/channels-list/saga.test.ts
```diff
@@ -1,13 +1,31 @@
...
+  global.fetch = () =>
+    Promise.resolve({
+      json: () => Promise.resolve(MOCK_CHANNELS),
+    });
+});
+
+afterAll(() => {
+  global.fetch = unmockedFetch;
+});
+
```
jtulloch, line 28: you shouldn't need to mock the global fetch like this. a good rule of thumb is if you find yourself mocking a global or an entire module, there's probably a better way to do it. feel free to reach out any time to ask. :)

react-saga-test-plan has mocking built-in. that's what the `provide` method does. so, for the first test below, all you have to do is: 
```
.provide([
        [
          matchers.call.fn(fetchChannels),
          MOCK_CHANNELS,
        ],
      ])
```

src/store/channels-list/api.ts
```diff
@@ -1,13 +1,10 @@
...
-};
+import { config } from '../../config';
+
+export async function fetchChannels(_id: string) {
+  const data = await fetch(`${config.ZERO_API_URL}/api/networks/${_id}/chatChannels/public`);
+  const channels = await data.json();
+  return channels.map((channel, index) => ({
+    id: `channel-id-${index}`,
+    ...channel,
+  }));
```
jtulloch, line 9: let's pull this into the saga so that we can easily test it. additionally, we should map the actual channel id, rather than generating one. double-check this, but i believe the id field on channels is `url`.

jtulloch (changes requested):

jtulloch (approved):

PR #133, opened 2022-08-03
infinite scroll channel view

mrval5:
#### Things to consider:
  1. Does this change any APIs?


src/store/messages/api.ts
```diff
@@ -1,9 +1,14 @@
...
+export interface MessagesFilter {
+  lastCreatedAt: string;
+}
 
-  return messages;
+export async function fetchMessagesByChannelId(channelId: string, filter?: MessagesFilter): Promise<MessagesResponse> {
+  const queryParams = JSON.stringify(filter || {});
+  const messagesResponse = await fetch(
+    `${config.ZERO_API_URL}/chatChannels/${channelId}/messages?filter=${encodeURIComponent(queryParams)}`
+  );
```
jtulloch, line 12: can we use superagent instead of using fetch directly here? makes using filters, etc., a lot cleaner, and reduces the chance of introducing hard to find errors with payload construction, etc.

src/platform-apps/channels/channel-view.tsx
```diff
@@ -1,27 +1,74 @@
...
+  }
+
+  setChannelViewRef = (ref) => {
+    this.channelViewRef = ref;
+  };
+
+  scrollToBottom = () => {
+    this.lastElementOnChannelViewRe.current?.scrollIntoView({ behavior: 'auto' });
+    this.setState({ hasScrolledToBottom: true });
+  };
```
jtulloch, line 39: instead of managing this directly, i think i would add a ScrollContainer component that works based on content size rather than specific elements. if we only had to scroll once, then this makes sense, but the problem with handling reverse scrolling is that you need to update the scroll position any time the content changes at all. that means it needs to be updated not only when new messages are loaded, but also any time a message itself changes size. eg. when an image is loaded, when the content is replaced (mentions, emojis, etc.), or for any other reason.

you can look at the one in zero for ideas, but it is more aggressive than we will need here, it is made to support changes that occur outside of virtual dom updates - which we don't want to do, and is also super messy so i wouldn't copy the entire thing directly.

jtulloch (changes requested):
alright, so two main suggestions:
- i think we can simplify the whole timestamp/fetch thing by just using a waypoint instead of trying to add/remove listeners from individual messages.
- i think the scroll management will end up being more complex than we want to handle as a one-off in this component, and i think it would be good to start building a ScrollContainer component that handles scrolling based on content size rather than individual components.

src/components/inverted-scroll/index.test.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+
+import { shallow } from 'enzyme';
+
+import InvertedScroll, { Properties } from '.';
+
+describe('feature-flag', () => {
```
jtulloch, line 7: two small picky things -- forgot to change the name of this. :)

jtulloch (changes requested):

jtulloch (approved):

jtulloch:
@mrval5 feel free to just resolve comments and merge for anything like this. it doesn't need a re-review unless the changes are reasonably complex and you're not certain it resolved what i was asking for.

PR #134, opened 2022-08-05
Grouped message by day

hkhanouchi:
Screen Shot 2022-08-05 at 02 09 43
Screen Shot 2022-08-05 at 02 09 23


package.json
```diff
@@ -28,6 +28,7 @@
     "history": "^4.10.1",
     "lodash.get": "^4.4.2",
     "lodash.kebabcase": "^4.1.1",
+    "moment": "^2.29.4",
```
jtulloch, line 31: can we see if we can use a different lib than moment.js? that is what we us in zero, but they have been recommending for a while that people use a different library. maybe pick something from this page? or pick a couple and present it in the chat room and we can decide? i haven't looked into it yet, so i'm not sure which would be best. 
https://momentjs.com/docs/#/-project-status/recommendations/

src/platform-apps/channels/channel-view.tsx
```diff
@@ -21,14 +55,21 @@ export class ChannelView extends React.Component<Properties> {
     );
   }
 
+  renderMessages() {
+    const messagesByDay = this.getMessagesByDay();
+    const groupDays = Object.keys(messagesByDay).sort((a, b) => (a > b ? 1 : -1));
+
+    return <div className='container'>{groupDays.map((day: string) => this.renderDay(day, messagesByDay))}</div>;
```
jtulloch, line 62: can we update this className to make it more specific? container has a lot of potential for conflicts as a class name. :)

src/platform-apps/channels/channel-view.test.tsx
```diff
@@ -40,10 +40,46 @@ describe('ChannelView', () => {
...
+    ];
+
+    const wrapper = subject({ messages });
+
+    expect(wrapper.find('.message__header-date').length).toStrictEqual(3);
+  });
+
+  it('renders a header Date contain Today', () => {
+    const messages = [
+      { id: 'message-one', message: 'what', createdAt: new Date() },
```
jtulloch, line 69: what is the type of createdAt from the server? let's make sure that we know what it is, and that we keep it consistent. above we're using timestamps, and here we're using a date object. let's pick one (either what the server returns, or convert it to something specific when we build the initial object) and use it consistently.

jtulloch (changes requested):

jtulloch (approved):
@mrval5 same as the other one. feel free to use your judgement and resolve comments and merge once they're resolved.

PR #135, opened 2022-08-05
Link preview

AdamDotCom:
Related PR: https://github.com/zer0-os/zOS-component-library/pull/26

link-preview


public/index.html
```diff
@@ -29,6 +29,19 @@
...
+        var id='embedly-platform', n = 'script';
+        if (!d.getElementById(id)){
+            w.embedly = w.embedly || function() {(w.embedly.q = w.embedly.q || []).push(arguments);};
+            var e = d.createElement(n); e.id = id; e.async=1;
+            e.src = ('https:' === document.location.protocol ? 'https' : 'http') + '://cdn.embedly.com/widgets/platform.js';
+            var s = d.getElementsByTagName(n)[0];
+            s.parentNode.insertBefore(e, s);
+        }
+      })(window, document);
+    </script>
```
jtulloch, line 44: do we need this? i don't think we should. getting rid of this stuff is the main reason we switched to the oembed version of their api.

src/components/link-preview/embed.tsx
```diff
@@ -0,0 +1,196 @@
+import React from 'react';
```
jtulloch, line 1: i think we can switch this file name to `index`. the only reason we had the alternate `embed` naming was to make it easier to transition from the old style of link previews to the new version.

.env.example
```diff
@@ -15,3 +15,5 @@ REACT_APP_INFURA_ID=
 REACT_APP_FORTMATIC_API_KEY=
 REACT_APP_PORTIS_DAPP_ID=
 
+# these are optional, but are required for link previews
+REACT_APP_EMBEDLY_KEY=
```
jtulloch, line 19: i don't think we should need any embedly references in the front end.

jtulloch (changes requested):

PR #137, opened 2022-08-10
add time to message

mrval5:
#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/platform-apps/channels/channel-view.tsx
+++ b/src/platform-apps/channels/channel-view.tsx
@@ -1,8 +1,9 @@
 import React from 'react';
 import { Waypoint } from 'react-waypoint';
+import classNames from 'classnames';
+import moment from 'moment';
 import { Message as MessageModel } from '../../store/messages';
 import { Message } from './message';
-import moment from 'moment';
 import InvertedScroll from '../../components/inverted-scroll';
 
 interface ChatMessageGroups {
@@ -56,7 +57,7 @@ export class ChannelView extends React.Component<Properties> {
 
           return (
             <Message
-              isFirstFromUser={isFirstFromUser}
+              className={classNames('messages__message', { 'messages__message--first-in-group': isFirstFromUser })}
               key={message.id}
               {...message}
             />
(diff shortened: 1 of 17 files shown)


src/platform-apps/channels/channel-view.tsx
```diff
@@ -11,12 +11,17 @@ export class ChannelView extends React.Component<Properties> {
...
-            key={message.id}
-            {...message}
-          />
-        ))}
+        {this.props.messages.map((message, index) => {
+          const isFirstFromUser = index === 0 || message.sender.userId !== this.props.messages[index - 1].sender.userId;
+
+          return (
+            <Message
+              isFirstFromUser={isFirstFromUser}
```
jtulloch, line 19: not totally sure it makes sense for the message to know whether it is the first message from a user. how the overall message list is displayed feels more like the responsibility of the parent than the message itself. instead of passing in a property here, can we just use a class?

something like this:
```
<Message
  className={classNames('messages__message', { isFirstFromUser: 'messages__message-first-in-group' })}
```
and then in the stylesheet:
```
.messages__message: {
  &-first-in-group: {
    .message__time: {
       visibility: hidden;

       &:hover {
         visibility: visible;
       }
    }
  }
}
```

src/platform-apps/channels/message.test.tsx
```diff
@@ -65,6 +65,24 @@ describe('message', () => {
...
+  it('renders time as first owner', () => {
+    const wrapper = subject({
+      message: 'the message',
+      createdAt: new Date('December 17, 1995 17:04:00').valueOf(),
+      isFirstFromUser: true,
+    });
+
+    expect(wrapper.find('.message__time').text()).toStrictEqual('17:04');
+
+    expect(wrapper.find('.message__time--hidden').exists()).toBe(false);
```
jtulloch, line 83: i find that the `hasClass` method makes it a bit clearer what we're asserting in situations like this:
```
expect(wrapper.find('.message__time').hasClass('.message__time--hidden')).toBe(false);
```

jtulloch (changes requested):

PR #141, opened 2022-08-21
add mention highlight

mrval5:
### What does this do?
- highlight to mentions

FYI: in zer0 when we click on mentions we open a modal. I'm not sure if this modal is part of this task @jtulloch .


src/platform-apps/channels/styles.scss
```diff
@@ -229,5 +230,21 @@
...
+    }
+  }
+}
+
+@keyframes fade-background-in {
+  from {
+    background-color: rgba(0, 0, 0, 0);
+  }
+  to {
+    background-color: #c9fd51;
```
jtulloch, line 248: let's make this color the 'accent color' from the theme.

src/platform-apps/channels/message.tsx
```diff
@@ -53,13 +62,35 @@ export class Message extends React.Component<Properties> {
...
+        if (profileId) {
+          return (
+            <ProfileLink
+              {...props}
+              id={profileId}
+            >
+              {mention}
+            </ProfileLink>
+          );
+        }
```
jtulloch, line 89: sorry, the task description was not very good for this one! let's simplify this for now. *all* we need to do is have it highlighted. we don't need it to be "clickable" at all for now. it should just be styled with the accent background color, and then the font should be readable. in zero we have accent color and text on accent color in the theme. if we don't have that in our theme here, let's add it and make sure it works in light and dark mode.

jtulloch (changes requested):
as mentioned in the comment. we can simplifiy this. i'll update the task now to make it more obvious, but this was my bad for not making the task detailed enough. sorry!

PR #143, opened 2022-08-23
feat: add zApp Staking to app container

colbr:
### What does this do?

- Adds a new external package: [`@zero-tech/zapp-staking`](https://www.npmjs.com/package/@zero-tech/zapp-staking) - this is the Staking zApp.
- Adds the `App.Staking` enum value to the list of active apps in `lib/apps/index.ts`.
- Renders Staking zApp when `selectedApp === App.Staking`.

Short answer: it adds zApp Staking to the list of apps.

To break ground on the relationship between zApps and zOS - this PR will act as validation of our process.

### How do I test this?

- Click "Staking" in the app selector.
- Check that the "Staking" app renders.

#### Things to consider:
  3. How will this affect performance?
  4. Does this change any APIs?


src/app-sandbox/index.tsx
```diff
@@ -4,6 +4,7 @@ import { Store } from 'redux';
 
 import { Apps } from '../lib/apps';
 import { App as FeedApp } from '@zer0-os/zos-feed';
+import StakingApp from '@zero-tech/zapp-staking';
```
jtulloch, line 7: maybe worth dropping the default export, but not the end of the world since this is the only usage.

jtulloch (dismissed):

src/app-sandbox/index.tsx
colbr: Good point - I'll drop it for consistency!

colbr:
All resolved @jtulloch!

PR #145, opened 2022-08-25
Add indicator when new messages arrive in real time

hkhanouchi:
Screen Shot 2022-08-25 at 14 19 02


src/platform-apps/channels/channel-view.tsx
```diff
@@ -79,13 +92,47 @@ export class ChannelView extends React.Component<Properties> {
   render() {
     return (
       <div className='channel-view'>
+        {this.props.hasNewMessage > 0 && (
```
jtulloch, line 95: can we extract this? this feels like a separate component to me.

jtulloch (changes requested):

src/platform-apps/channels/channel-view.tsx
hkhanouchi: OK

PR #153, opened 2022-08-30
Feature/indicator new messages

hkhanouchi:
Screen Shot 2022-08-25 at 14 19 02


src/platform-apps/channels/channel-view.tsx
```diff
@@ -79,13 +93,19 @@ export class ChannelView extends React.Component<Properties> {
   render() {
     return (
       <div className='channel-view'>
+        <IndicatorMessage
+          hasNewMessage={this.props.hasNewMessage}
+          closeIndicator={this.props.closeIndicator}
+          scrollToBottom={this.scrollToBottom}
```
jtulloch, line 99: i think we can get rid of one of these callbacks. let's keep the 'closeIndicator' one and remove the scrollToBottom one, as the Indicator component shouldn't need to know what the ultimate effect of closing it will be.

can we also add a test for this? it should make sure that we're scrolling the ref, as well as forwarding the callback to the parent.

jtulloch (changes requested):
nice! just one more thing. :)

src/components/indicator-message/index.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  it('it should not renders indicator when click on it', () => {
+    const closeIndicatorSpy = jest.fn();
+    const wrapper = subject({ countNewMessage: 3, closeIndicator: closeIndicatorSpy });
+
+    expect(wrapper.find('.indicator__new-message').exists()).toBe(true);
+
+    wrapper.find('.indicator__new-message-bar').simulate('click');
+
+    expect(wrapper.prop('closeIndicator')).toHaveBeenCalled;
```
jtulloch, line 40: pretty sure that `toHaveBeenCalled` is a function. this expectation probably isn't doing what we think it is, and might actually be causing problems. was this the branch that had timeout failures? this could be the reason! it might be waiting for an expectation, and timing out when none is there.

jtulloch (changes requested):

src/components/indicator-message/index.test.tsx
hkhanouchi: No, the reason was the setInterval function which caused the problem, I fixed the problem, i added a 

> `beforeEach(() => {

     jest.useFakeTimers();
   })` 
in the test file to fix the issue

jtulloch: ah. k. so, two takeaways:
- the first is just that. .. my main point was that this test probably isn't working. `toHaveBeenCalled`  is a function, and we're not calling it here. (we're also accessing it indirectly using `prop` instead of just having the expectation based on the spy itself. let's change that while we're at it.)
- second is that components are generally the wrong place for code that uses setInterval. there are exceptions, but if you find yourself using fakeTimers in a component test, there's probably a better way to organize the code. i'll add comments elsewhere to make it clearer what i'm talking about for this one.

src/platform-apps/channels/channel-view-container.tsx
```diff
@@ -32,22 +36,44 @@ export class Container extends React.Component<Properties> {
...
+  state = { countNewMessage: 0 };
+
   componentDidMount() {
     const { channelId } = this.props;
-
     if (channelId) {
       this.props.fetchMessages({ channelId });
+      this.fetchInterval = setInterval(() => {
+        this.props.fetchMessages({ channelId });
+      }, 20000);
```
jtulloch, line 47: sorry. i think maybe i didn't look closely enough at this PR the first time through. let's pull the setInterval out of the component. generally things that manage scheduling like this are best encapsulated in a way that nothing which uses it has to account for the fact that it is on a timer. ie. the thing that does the scheduling is a different thing than the part that is doing the work. the only time you should need to account for the fact that something is 'scheduled' or using a timer is when you are directly testing the scheduler itself.

instead of having an interval here, let's just make this a trigger for a scheduler. so, something like this:
```
this.props.startMessageSync({ channelId });
```

you can see an example of how we might want to do the actual scheduling here: https://github.com/m3m3n70/zero-web/blob/development/src/store/sagas/chat-channel-list.ts#L123

jtulloch (changes requested):

AdamDotCom (approved):

src/platform-apps/channels/channel-view-container.tsx
```diff
@@ -29,25 +33,43 @@ export class Container extends React.Component<Properties> {
...
   }
 
+  state = { countNewMessages: 0 };
+
   componentDidMount() {
     const { channelId } = this.props;
 
     if (channelId) {
       this.props.fetchMessages({ channelId });
+      this.props.startMessageSync({ channelId });
```
jtulloch, line 47: i think we probably also need to stop the message sync, when the user switches channels, or navigates away from channels entirely. it's possible that the saga just handles the first scenario (we should verify that if we haven't), but in the case of navigating away from channels, we need to have an action that stops the channel syncing, or it will continue in the background.

jtulloch (changes requested):

jtulloch (dismissed):

PR #154, opened 2022-08-31
build(craco): configure for linked ES6 modules 

colbr:
### What does this do?

`babel-loader`, by default, doesn't run on `node_modules`. This means we will get an error if we use ES6+ syntax (i.e. function components) in a packages symlinked through `npm link`. The error looks something like the below:

```
Module parse failed: Unexpected token (23:46)
File was processed with these loaders:
 * ./node_modules/@pmmmwh/react-refresh-webpack-plugin/loader/index.js
You may need an additional loader to handle the result of these loaders.

> export const Component = () => (
|       <>
```

[`craco-babel-loader`](https://www.npmjs.com/package/craco-babel-loader) is a CRACO plug-in which can run `babel-loader` on any folder. This means we can point it to our symlinked folders, and it will compile them at the same time as the host app (zOS).

As mentioned above, symlinked components using ES6+ syntax aren't being run through `babel-loader` and thus causing errors.

### How do I test this?

1. Create a function component in an external package.
2. Symlink (`npm link`) the external package into zOS.
3. Render the symlinked component in zOS.
(If you run zOS here, you'll see the error we're trying to solve!)
4. Add the relative link to your external package folder to the array parameter of `resolveApps` in `craco.config.js`.
5. Start zOS (or restart, if you tested at step 3)
6. The symlinked component should render, and not throw any errors.

#### Things to consider:
This should be configured to only run in a dev environment.
  2. How will this affect performance?
It won't affect production performance, and will only marginally affect dev performance.
  3. Does this change any APIs?
Nope!

### DIFF
--- a/craco.config.js
+++ b/craco.config.js
@@ -1,4 +1,21 @@
 const path = require('path');
+const fs = require('fs');
+const appDirectory = fs.realpathSync(process.cwd());
+
+/**
+ * Converts a relative path to an absolute path.
+ * e.g. '../zApp-Staking' -> '/users/zero/zApp-Staking'
+ */
+const resolveApp = (relativePath) => {
+  path.resolve(appDirectory, relativePath);
+};
+
+/**
+ * Maps a list of relative app locations to absolute paths
+ */
+const resolveApps = (appLocations) => {
+  return appLocations.map((app) => resolveApp(app));
+};
 
 module.exports = {
   webpack: {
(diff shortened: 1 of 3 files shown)


jtulloch:
@colbr hmm. curious why we want this to be the responsibility of zOS? unless there's a really good reason, i think that building apps should be the responsibility of the app itself. the reason the build doesn't run on node_modules is because the consumer shouldn't need to know what it takes to build each package it consumes. additionally, running the build from totally different places for development and production introduces quite a bit of risk, and delays it to a really late stage in the process.

if the app loads *after* the package is published, it should also work using npm link, so it's probably worth digging into what is actually going on, rather than trying to get around it. can you share a bit more detail about what exactly the issue is? is there any more context for that error? is your prod/development build set up the same way in your package? with the same targets, etc? (at a glance, it *looks* like there might still be JSX in that error output. is it possible that your build is set up differently when packaging vs. development?)

jtulloch:
closing for now, due to reasons mentioned above. feel free to re-open if you want to continue the discussion.

PR #159, opened 2022-09-06
Add Lightbox

AdamDotCom:
Related PR: https://github.com/zer0-os/zOS-component-library/pull/28
zOS-lightbox


src/platform-apps/channels/channel-view.test.tsx
```diff
@@ -116,4 +117,66 @@ describe('ChannelView', () => {
...
+        {
+          id: 'message-3',
+          message: 'audio message',
+          media: { url: 'video.mp3', type: MediaType.Audio },
+          sender: { userId: 1 },
+          createdAt: 1658776625732,
+        },
+      ];
+      const wrapper = subject({ messages });
+      wrapper.instance().openLightbox();
```
jtulloch, line 148: let's not call internal methods. we should be able to trigger this from the boundary so that we're not depending on implementation details. something like:
```
wrapper.find(Message).at(1).simulate('imageClicked');
```

src/platform-apps/channels/message.tsx
```diff
@@ -1,12 +1,13 @@
...
 import classNames from 'classnames';
 import moment from 'moment';
-import { Message as MessageModel } from '../../store/messages';
+import { Message as MessageModel, MediaType } from '../../store/messages';
 import { textToEmojis } from './utils';
 import { LinkPreview } from '../../components/link-preview/';
 import { CloudinaryProvider, provider } from '../../lib/cloudinary/provider';
 interface Properties extends MessageModel {
   className: string;
+  openLightbox: any;
```
jtulloch, line 10: i don't think the event name should reflect what the consumer is doing in response to the event. .. it should just reflect what has happened from the perspective of this component. as a message, i shouldn't need to know what my consumer plans to do when i am clicked. i should just tell the consumer that i was clicked.
something like:
```
onImageClick: () => void;
```

jtulloch (changes requested):

PR #160, opened 2022-09-07
Add categories to channels list

hkhanouchi:
Screen Shot 2022-09-07 at 18 36 17

### DIFF
--- a/src/platform-apps/channels/channel-list.test.tsx
+++ b/src/platform-apps/channels/channel-list.test.tsx
@@ -4,6 +4,7 @@ import { shallow } from 'enzyme';
 
 import { ChannelList } from './channel-list';
 import { ZnsLink } from '@zer0-os/zos-component-library';
+import Collapsible from 'react-collapsible';
 
 describe('ChannelList', () => {
   const subject = (props: any = {}) => {
@@ -16,10 +17,10 @@ describe('ChannelList', () => {
 
   it('renders each channel name', () => {
     const channels = [
-      { id: 'one', name: 'first channel' },
-      { id: 'two', name: 'second channel' },
-      { id: 'three', name: 'third channel' },
-      { id: 'four', name: 'fourth channel' },
+      { id: 'one', name: 'first channel', category: 'catg1' },
+      { id: 'two', name: 'second channel', category: 'catg1' },
+      { id: 'three', name: 'third channel', category: 'catg2' },
+      { id: 'four', name: 'fourth channel', category: 'catg2' },
     ];
 
     const wrapper = subject({ channels });
@@ -56,10 +57,10 @@ describe('ChannelList', () => {
 
   it('renders Link to channel id', () => {
     const channels = [
-      { id: 'one', name: 'first channel' },
-      { id: 'two', name: 'second channel' },
-      { id: 'three', name: 'third channel' },
-      { id: 'four', name: 'fourth channel' },
+      { id: 'one', name: 'first channel', category: 'catg1' },
+      { id: 'two', name: 'second channel', category: 'catg1' },
+      { id: 'three', name: 'third channel', category: 'catg2' },
+      { id: 'four', name: 'fourth channel', category: 'catg2' },
     ];
 
     const wrapper = subject({ channels });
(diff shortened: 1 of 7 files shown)


src/platform-apps/channels/channel-list.test.tsx
```diff
@@ -16,10 +17,10 @@ describe('ChannelList', () => {
...
   it('renders each channel name', () => {
     const channels = [
-      { id: 'one', name: 'first channel' },
-      { id: 'two', name: 'second channel' },
-      { id: 'three', name: 'third channel' },
-      { id: 'four', name: 'fourth channel' },
+      { id: 'one', name: 'first channel', category: 'catg1' },
+      { id: 'two', name: 'second channel', category: 'catg1' },
+      { id: 'three', name: 'third channel', category: 'catg2' },
+      { id: 'four', name: 'fourth channel', category: 'catg2' },
```
jtulloch, line 23: i don't think we need to add categories to tests that aren't using them. let's remove the category from these, since this test doesn't have anything to do with categories.

src/platform-apps/channels/channel-list.test.tsx
```diff
@@ -36,10 +37,10 @@ describe('ChannelList', () => {
...
   it('renders Link to channel id', () => {
     const channels = [
-      { id: 'one', name: 'first channel' },
-      { id: 'two', name: 'second channel' },
-      { id: 'three', name: 'third channel' },
-      { id: 'four', name: 'fourth channel' },
+      { id: 'one', name: 'first channel', category: 'catg1' },
+      { id: 'two', name: 'second channel', category: 'catg1' },
+      { id: 'three', name: 'third channel', category: 'catg2' },
+      { id: 'four', name: 'fourth channel', category: 'catg2' },
```
jtulloch, line 43: same as above. there's no reason to have categories on the channels in this test.

src/platform-apps/channels/channel-list.test.tsx
```diff
@@ -53,4 +54,33 @@ describe('ChannelList', () => {
...
+    const wrapper = subject({ channels });
+    expect(wrapper.find('.channel-list__category-channel-name').exists()).toBe(true);
+  });
+
+  it('renders grouped category', () => {
+    const channels = [
+      { id: 'one', name: 'first channel', category: 'catg1' },
+      { id: 'two', name: 'second channel', category: 'catg1' },
+      { id: 'three', name: 'third channel', category: 'catg2' },
+      { id: 'four', name: 'fourth channel', category: 'catg2' },
```
AdamDotCom, line 75: It looks like all these tests depend on the same test data. Rather than redefining channels for each test can we move channels into the `subject`? 

```
  const subject = (props: any = {}) => {
     const channels = [
        { id: 'one', name: 'first channel', category: 'catg1' },
        { id: 'two', name: 'second channel', category: 'catg1' },
        { id: 'three', name: 'third channel', category: 'catg2' },
        { id: 'four', name: 'fourth channel', category: 'catg2' },
     ];
    const allProps = {
      channels,
      ...props,
    };
    return shallow(<ChannelList {...allProps} />);
  };
```

src/platform-apps/channels/channel-list.test.tsx
```diff
@@ -53,4 +54,33 @@ describe('ChannelList', () => {
...
+    ];
+
+    const wrapper = subject({ channels });
+    const textes = wrapper.find(Collapsible).map((l) => l.prop('trigger'));
```
AdamDotCom, line 79: Typo `textes` should be `text`.

src/platform-apps/channels/channel-list.tsx
```diff
@@ -1,17 +1,30 @@
...
   channels: Channel[];
 }
 
+interface ChannelsCategory {
+  channels: Channel[];
+}
+
 export class ChannelList extends React.Component<Properties> {
-  render() {
+  getChannelsByCategpry(channels: Channel[]): ChannelsCategory[] {
```
AdamDotCom, line 16: Typo `getChannelsByCategpry`, `Categpry` should be `Category`

src/platform-apps/channels/channel-list.tsx
```diff
@@ -25,4 +38,35 @@ export class ChannelList extends React.Component<Properties> {
...
+          return (
+            <div
+              key={category}
+              className='channel-list__category-channel'
+            >
+              <Collapsible
+                className='channel-list__category-channel-name'
+                openedClassName='channel-list__category-channel-name'
+                trigger={category}
+                open={true}
```
AdamDotCom, line 58: I think `open` without the true assignment is equivalent `open={true}`

src/platform-apps/channels/styles.scss
```diff
@@ -45,6 +45,37 @@
       border-left: 3px solid #fb216e;
     }
   }
+
+  &-list__category-channel-name {
+    .Collapsible__trigger {
```
AdamDotCom, line 50: Is it possible to have the css class name lowercase. I don't think we have any uppercase at this point?

AdamDotCom: It looks like we can override it using `classParentString`: https://github.com/glennflanagan/react-collapsible#classparentstring--string--default-collapsible

AdamDotCom (changes requested):

AdamDotCom (approved):

jtulloch (approved):

PR #164, opened 2022-09-13
add authentication - login to zero

mrval5:
### What does this do?
the changes let the users connect to their wallet and connect to zero.

env variable added to github.

### How do I test this?
connect to meta mask
meta mask will ask to sign to connect to zero


src/components/wallet-manager/index.tsx
```diff
@@ -109,6 +115,41 @@ export class Container extends React.Component<Properties, State> {
...
   }
 
+  onConnect(): void {
+    this.closeModal();
+    this.setState({ walletSelected: false });
+
+    this.login();
+  }
+
+  login(): void {
```
mrval5, line 125: @jtulloch this method will always (each refresh of the page) will ask for a new sign to get a new token which I think is not correct - let's chat about it during the daily.

src/store/authentication/index.ts
```diff
@@ -0,0 +1,28 @@
...
+const initialState: AuthenticationState = {
+  accessToken: null,
+};
+
+const slice = createSlice({
+  name: 'authentication',
+  initialState,
+  reducers: {
+    setAccessToken: (state, action: PayloadAction<AuthenticationState['accessToken']>) => {
+      state.accessToken = action.payload;
```
AdamDotCom, line 21: Should we store the access token in a cookie or local storage? If it's in state will the user be prompted to login when opening a new tab?

mrval5: You are right, as far as I know the cookie should come in the response from the api

another comment which is related to the token cookie here: https://github.com/zer0-os/zOS/pull/164#discussion_r969335388

src/components/wallet-manager/index.tsx
jtulloch: the more i think about this, i think this piece is going to require some thought. .. let's ignore what i said about changing addresses for the moment. i'll make a new task, as i think it is a bigger question, and is also related to the account creation stuff. it will need at least some input from matt, as i don't think we can fully solve that problem without interacting with the user.

src/store/authentication/index.ts
jtulloch: (yeah. if we can use a cookie and avoid storing this, even better.)

jtulloch (approved):

PR #167, opened 2022-09-21
Add text box to send a message to the current channel

hkhanouchi:
### How do I test this?

Screen Shot 2022-09-21 at 13 53 17


src/store/messages/api.ts
```diff
@@ -11,3 +11,15 @@ export async function fetchMessagesByChannelId(channelId: string, lastCreatedAt?
...
 }
+
+export async function sendMessagesByChannelId(
+  channelId: string,
+  message: string,
+  mentionedUser: string
+): Promise<boolean> {
+  const response = await Request.post(`${config.ZERO_API_URL}/chatChannels/${channelId}/message`)
+    .withCredentials()
+    .send({ message, mentionedUsers: [mentionedUser] });
```
jtulloch, line 22: let's wrap the request object so that we don't need to provide the api url every time.
we should also not need to add `.withCredentials()`. that should just happen automatically.

src/store/messages/saga.test.ts
```diff
@@ -32,6 +32,28 @@ describe('messages saga', () => {
       .run();
   });
 
+  it('send message', async () => {
+    const channelId = '0x000000000000000000000000000000000000000A';
+    const message = 'hello';
+    const mentionedUser = '00000000000000A';
```
jtulloch, line 38: i don't think we need 'mentionedUser' here yet, as we don't have mentions at this point.

src/platform-apps/channels/channel-view.tsx
```diff
@@ -104,6 +106,15 @@ export class ChannelView extends React.Component<Properties, State> {
     );
   }
 
+  renderChatWindow() {
+    return (
+      <ChatWindow
+        isUserConnected={true}
+        onSubmit={this.props.sendMessage}
+      />
```
jtulloch, line 114: i'm not sure this is the right abstraction here. i think that ChannelView and ChannelWindow are actually the same thing. i think we can probably just pull the input into the channel view, instead of needing to also add the ChatWindow component?

src/platform-apps/channels/channel-view-container.test.tsx
```diff
@@ -98,6 +99,26 @@ describe('ChannelViewContainer', () => {
...
+
+    const wrapper = subject({
+      sendMessage,
+      channelId: 'the-channel-id',
+      channel: { hasMore: true, name: 'first channel' },
+    });
+
+    wrapper.find(ChannelView).first().prop('sendMessage')(message);
+
+    expect(sendMessage).toHaveBeenLastCalledWith({
```
jtulloch, line 115: i think that `toHaveBeenLastCalledWith` might be the wrong expectation here. i think in this case we want to make sure that it was called once and only once.

src/components/chat-window/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+  onSubmit = (message: string) => {
+    this.props.onSubmit(message);
+  };
+
+  render() {
+    return (
+      <div className={classNames('chat-window__input-wrapper', this.props.className)}>
+        {!this.props.isUserConnected && (
+          <div className='chat-window__input-disabled'>You do not have permission to send messages in this channel</div>
+        )}
```
jtulloch, line 24: for this task, let's just show or not show the input. we don't need to have a message in the case of not being connected.

src/store/messages/saga.ts
```diff
@@ -44,6 +50,15 @@ export function* fetch(action) {
   );
 }
 
+export function* send(action) {
+  const { channelId, message, mentionedUser } = action.payload;
+
+  yield call(sendMessagesByChannelId, channelId, message, mentionedUser);
+
+  yield call(fetch, { payload: { channelId } });
```
jtulloch, line 58: i think instead of re-fetching messages we probably want to just add the message directly to the channel state. we can follow that up by fetching messages, but i think just fetching will be too slow of a user experience, when we already have the message that we are adding.

jtulloch (changes requested):

src/store/messages/saga.test.ts
hkhanouchi: We need `'mentionedUser' ` here for the test, because the `sendMessagesByChannelId` function used. 
` yield call(sendMessagesByChannelId, channelId, message, mentionedUser)`

src/store/messages/saga.ts
hkhanouchi: I agree, but for our case we only have the message text, we miss  the other messages properties.

jtulloch (changes requested):

src/store/messages/saga.ts
hkhanouchi: Yes, the endpoint just return the status of request, i'll create those tasks for that.

jtulloch (approved):

PR #181, opened 2022-09-30
Add message directly to channel state after send

hkhanouchi:
#### Things to consider:
  1. Does this change any APIs?


src/store/messages/saga.ts
```diff
@@ -61,9 +61,24 @@ export function* fetch(action) {
 export function* send(action) {
   const { channelId, message } = action.payload;
 
-  yield call(sendMessagesByChannelId, channelId, message);
+  const messagesResponse = yield call(sendMessagesByChannelId, channelId, message);
+  if (messagesResponse.status === 200) {
```
jtulloch, line 65: for things like this where an entire block lives inside an `if` statement, i generally prefer either: setting it up as a guard rather than a block, so: 
`if (messagesResponse.status !== 200) return;`

or extracting to a method so it's only a one-liner inside the `if` block.
```
if (messagesResponse.status === 200) {
    yield sendMessage(messageResponse);
}
```

jtulloch (changes requested):

src/store/messages/saga.ts
jtulloch: also: tests!

PR #188, opened 2022-10-06
Send message instantly

mrval5:
#### Things to consider:
  1. Does this change any APIs?


mrval5:
I have to fix comments and only test

PR #190, opened 2022-10-06
Feature/channels textbox mentions

hkhanouchi:
Screen Shot 2022-10-06 at 16 56 00


src/store/users/index.ts
```diff
@@ -0,0 +1,37 @@
+import { createAction, createSlice, PayloadAction } from '@reduxjs/toolkit';
+
+import { FetchUsersPayload } from './saga';
+
+export interface UserMentions {
```
jtulloch, line 5: this interface shouldn't be at the level of the user, as this is a very local way of thinking about a user. it is tied directly to how users are considered from a very local component. if this interface is important, it shouldn't really bleed outside of the components that are

src/store/users/saga.ts
```diff
@@ -0,0 +1,21 @@
+import { takeLatest, put, call } from 'redux-saga/effects';
+import { SagaActionTypes, setUsers } from '.';
+
+import { searchMyNetworks } from './api';
+
+export interface FetchUsersPayload {
+  search: string;
+  networkIds?: string[];
+  feedItemId?: string;
```
jtulloch, line 9: what is this used for?

src/store/users/api.ts
```diff
@@ -0,0 +1,7 @@
+import { get } from '../../lib/api/rest';
+
+export async function searchMyNetworks(filter: string): Promise<any> {
+  const response = await get<any>('/api/users/searchInNetworks', { filter });
+
+  return response.body;
+}
```
jtulloch, line 7: as mentioned in discord, let's use a different endpoint for this.

jtulloch (changes requested):
added some comments, but i think we need to restructure this a bit.
a) let's use a different endpoint
b) let's trigger the fetch using an action on the channel, not the user.
c) let's add a relation to channels so that the user's are connected to the channel that they're fetched for.
d) let's access the users using the channel, as that will speed things up when re-accessing a channel, and will also make it more clear what it is we're doing. eg. `myChannel.users`

src/platform-apps/channels/channel-view-container.tsx
```diff
@@ -35,17 +39,20 @@ export class Container extends React.Component<Properties, State> {
     const channel = denormalize(props.channelId, state) || null;
     const {
       authentication: { user },
+      users: { users },
     } = state;
 
     return {
       channel,
       user,
+      users,
```
jtulloch, line 48: we shouldn't just be grabbing all users here. .. i think a couple of steps from my initial feedback might have gotten lost in the shuffle:
b) let's trigger the fetch using an action on the channel, not the user.
c) let's add a relation to channels so that the user's are connected to the channel that they're fetched for.
d) let's access the users using the channel, as that will speed things up when re-accessing a channel, and will also make it more clear what it is we're doing. eg. myChannel.users

essentially, at this point, since we've already denormalized the channel, you should *already* have the users. accessing them should be the same as for messages: `channel.users` or `channel.members`

jtulloch: https://github.com/zer0-os/zOS/blob/development/src/store/channels/index.ts
you'll need to add a relationship to the channel for members/users/whatever we want to call it, and it should be connected to users in the same way that channels are connected to messages.

jtulloch (changes requested):

src/store/authentication/types.ts
```diff
@@ -37,3 +37,13 @@ export interface User {
 export interface AuthenticationState {
   user: UserPayload;
 }
+
+export interface Member {
```
jtulloch, line 41: let's put this in either the channel store or the user. if we're using the terminology of 'user' globally in the app, then let's rename this, and keep it with the user interface. or else just keep it scoped to the channel if we just want it specifically to be for channels, in which case it can be called Member, if we're calling channel member 'Members'

src/store/users/api.ts
```diff
@@ -0,0 +1,7 @@
+import { get } from '../../lib/api/rest';
+
+export async function searchMyNetworks(channelId: string): Promise<any> {
```
jtulloch, line 3: this needs to be named based on what it's actually doing. let's also pull it into the channels api, as we're fetching user for a channel, not users for a network.

jtulloch (changes requested):

src/store/users/saga.ts
```diff
@@ -0,0 +1,32 @@
...
+    ...rest,
+  }));
+
+  yield put(
+    receive({
+      id: channelId,
+      users: formatUsers,
+    })
+  );
+}
```
jtulloch, line 28: i don't think this belongs in the user saga. this method is too generic, since this feature isn't about fetching users, it's about fetching channel members. the saga should be on the channel, not the user. the problem here is that in the future, once we need to fetch users for some other purpose, we need to change this method. let's pull this into the channel saga, and name it appropriately for what it's doing. eg. `loadUsers` or something.

src/components/message-input/index.tsx
```diff
@@ -1,47 +1,88 @@
...
 interface State {
   value: string;
+  userIds: string[];
 }
 
 export class MessageInput extends React.Component<Properties, State> {
-  state = { value: '' };
-  onSubmit = (e) => {
+  state = { value: '', userIds: [] };
+  onSubmit = (e): void => {
```
jtulloch, line 24: you don't have to change this now, but in the future let's not add types to things like this. adding too many types makes it harder to make changes. for things like this typescript already knows what the return type is.

jtulloch (changes requested):
pretty close, but i think we need to switch the responsibility over to the channel saga from the user saga.

src/store/channels-list/saga.ts
```diff
@@ -43,6 +47,24 @@ export function* unreadCountUpdated(action) {
   yield put(receive(channelsList));
 }
 
+export function* loadUsers(action) {
```
jtulloch, line 50: sooooooo. i *think* this is still the wrong place for this. haha. sorry. :)

the channel-list state/saga has a very specific purpose, which is to manage the main list of channels that is displayed in the context menu.

essentially there are (at least) two types of entities in global state: 
- there's the core domain objects like user, channel, message, etc.,
- and then there is higher-level state that manages specific pieces of the app or UI.

the channel-list piece of state is the second type, while the channels and users piece of state is the first. what we are doing here is loading the users for a given channel. it isn't something that relates to the list of channels, it relates to a specific channel. for that reason, it should live in the channel 'slice', and in a channels saga.

let me know if that makes sense or if you have any questions about it. i know this stuff is pretty complex for how many examples we have at this point. i'll try and work up a diagram in the next few days, but for now if you'd like me to take a pass and show you what i mean, i'd be happy to do that. otherwise i think we're pretty close. :)

jtulloch (changes requested):

src/store/channels-list/saga.ts
hkhanouchi: ok, no worries. I'll change the place for the user saga.

PR #195, opened 2022-10-07
Channel login prompt

AdamDotCom:
connect-button


mrval5 (approved):

src/platform-apps/channels/channel-view.test.tsx
```diff
@@ -29,15 +39,35 @@ describe('ChannelView', () => {
...
+      messages: [],
+      user: null,
+      countNewMessages: 0,
+      ...props,
+    };
+
+    const store = createStore(reducer);
+
+    return mount(
+      <Provider store={store}>
```
jtulloch, line 54: i think this is a sign that something is wrong. we shouldn't need to provide a store within our tests. what is the reason for adding this? in the off chance that we *do* need the store, we should make sure we're adding the global store, not creating one like this, as the data structure will be different in this case than it is for the global store.

src/platform-apps/channels/channel-view.test.tsx
```diff
@@ -29,15 +39,35 @@ describe('ChannelView', () => {
...
+        <AuthenticationContext.Provider value={{ ...context }}>
```
jtulloch, line 55: are we just adding this provider so that we can test that it shows/hides the connect button? since we've tested the internal logic elsewhere, it should be sufficient here to just test that we are wrapping the message input with the correct IfAuthenticated component.

src/platform-apps/channels/channel-view.tsx
```diff
@@ -164,7 +152,15 @@ export class ChannelView extends React.Component<Properties, State> {
...
-          {this.renderChatWindow()}
+          <IfAuthenticated show>
+            <MessageInput
+              placeholder='Speak your truth...'
+              onSubmit={this.props.sendMessage}
+            />
+          </IfAuthenticated>
+          <IfAuthenticated hide>
+            <ConnectButton />
+          </IfAuthenticated>
```
jtulloch, line 163: instead of needing to have multiple blocks, and add the ConnectButton everywhere you want to show it, i think that part can just be pulled into the IfAuthenticated component. something like we do in the blockchain restricted component:
https://github.com/m3m3n70/zero-web/blob/development/src/components/blockchain-restricted/container.tsx

i think it will be a pretty common pattern to either show content, or show the connect button. so adding a `includeConnectButton` property or something similar to the IfAuthenticated component would streamline that.

src/components/authentication/button.tsx
```diff
@@ -0,0 +1,21 @@
...
+  };
+
+  render() {
+    return (
+      <div onClick={this.openModal}>
+        <ComponentButton>Connect</ComponentButton>
+      </div>
+    );
+  }
+}
```
jtulloch, line 21: i don't think this needs a separate container component. it is simple enough that i would just make this the container.

src/components/authentication/if-authenticated.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+import { AuthenticationContext } from '../../context/authentication';
+
+export interface Properties {
+  show?: boolean;
+  hide?: boolean;
```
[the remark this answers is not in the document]
jtulloch: just noticed where we are using this. i think if the only reason we have this is to show the button, i'd just pull the button into here for now, as i described in the other comment, and get rid of the show/hide thing.

src/context/authentication-provider.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { AuthenticationContext } from './authentication';
+import { AuthenticationContextProvider } from './authentication-provider';
+
+describe('AuthenticationProvider', () => {
+  const ChildComponent = () => {
+    return <></>;
+  };
+
+  const subject = (state: any = {}) => {
+    const store = createStore(reducer, state);
```
jtulloch, line 20: as with the other place, we shouldn't need to create a store in tests. it's a lot of overhead that should be avoidable.

src/components/authentication/if-authenticated.tsx
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export class IfAuthenticated extends React.Component<Properties> {
+  static contextType = AuthenticationContext;
```
jtulloch, line 11: adding context in this way makes the structure of the context less discoverable, and it also makes it harder to test. instead, i would add an HOC so that the context is just injected through props. 

something like this: https://github.com/zer0-os/zOS-component-library/blob/main/src/components/app-layout/context.tsx#L17

and consumed like this: https://github.com/zer0-os/zOS-component-library/blob/main/src/components/app-layout/app-layout.tsx#L44

jtulloch (changes requested):
added a bunch of comments, but i *think* it mostly comes down to tweaking how the context is set up so that we can just test at the boundaries instead of needing to mount and test at a higher level.

PR #204, opened 2022-10-17
fix divider

mrval5:
#### Things to consider:
  1. Does this change any APIs?


jtulloch (approved):
you can just merge PRs like this without going through review. :)

PR #205, opened 2022-10-17
fade in channel view

mrval5:
#### Things to consider:
  1. Does this change any APIs?


jtulloch (approved):
assuming we've already added the 'fadein' animation, this looks fine.

PR #230, opened 2022-10-31
fix(app-sandbox): scrollbar width

colbr:
### What does this do?

Adds functionality to prevent the sandbox scrollbar from pushing sandbox content to the left.

https://user-images.githubusercontent.com/12437916/198920015-09c35b8a-1f88-4a4a-ac0f-3ae70a577de7.mov

**Note: this is not an issue on Firefox.**

- Added a `ref` to the `app-sandbox` element.
- Add a `ResizeObserver` to `app-sandbox` to handle size changes. I attached a `ResizeObserver` to `app-sandbox` instead of using a `window` event listener because the context menus opening/closing could cause the app to overflow (and add a scrollbar).
- On `app-sandbox` size change, calculate scroll bar width and apply scroll bar width as `paddingLeft` on `app-sandbox`. Scroll bar width should be calculated as browsers seem to all react differently to scrollbar styling, and because it allows for more robust code at the cost of a tiny computation (i.e. don't have to hardcode the scrollbar width).

Because the app shouldn't be pushed left when the `app-sandbox` overflows.

### How do I test this?

1. Open the current development build in any non-Firefox browser.
2. Open any app which overflows (I used `Trade`).
3. Zoom out until the content no longer overflows.
4. Resize your browser (vertical) until the scroll bar appears (and pushes the content left).

#### Things to consider:
-

  2. How will this affect performance?
`ResizeObserver` is about as fast as we're going to get this. The `app-sandbox` container only sets state when scrollbar width changes, so shouldn't happen in quick succession.

  3. Does this change any APIs?
Added an extra prop to `app-sandbox`, and an extra value to the `layout` context. Apart from that - no changes.


colbr:
Hey @AdamDotCom - creating this PR for a problem n3o had raised a couple weeks back.

I had a pretty hard time understanding how the codebase works, so I'm hoping this is all correct enough. I didn't add any tests as I wasn't sure how to approach them.

I'm pretty strapped for time at the moment - if this is 90% of the way there would your team be able to take this task from here and I can learn from what y'all add? If not, I can probably push it over the line, but it's going to take a while for me to get to it.

mrval5 (approved):

jtulloch:
hey @colbr -- i don't think this is how we should be handling this. scrolling should be the responsibility of each individual app. it isn't something we should be handling in the platform at all, as it requires too many assumptions about how individual apps will look and behave.

jtulloch (changes requested):
hey guys, please see my comment. i think this strategy is placing too much responsibility on the platform. thanks!

colbr:
Agreed @jtulloch - have solved it in each zApp instead

PR #235, opened 2022-11-01
Fix the sync connection state

hkhanouchi:
#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/wallet-manager/index.tsx
+++ b/src/components/wallet-manager/index.tsx
@@ -21,6 +21,7 @@ export interface Properties extends PublicProperties {
   updateConnector: (connector: WalletType | Connectors.None) => void;
   setWalletModalOpen: (isWalletModalOpen: boolean) => void;
   isWalletModalOpen: Web3State['isWalletModalOpen'];
+  isAuthenticated: boolean;
 }
 
 export interface State {
@@ -38,6 +39,7 @@ export class Container extends React.Component<Properties, State> {
       currentAddress: value.address,
       connectionStatus: status,
       isWalletModalOpen,
+      isAuthenticated: !!state.authentication.user?.data && !!value.address,
     };
   }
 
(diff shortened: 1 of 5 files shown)


AdamDotCom:
We have an existing Connect button component. I'm not sure, but is it possible to removed the showButton logic and the Connect button from the WalletManager and make use of the IfAuthenticated component in [Main.tsx](https://github.com/zer0-os/zOS/blob/f3c8302def930f295ed90260f207c0b8f8f1a33d/src/Main.tsx#L57-L57)? 
```
  <IfAuthenticated showChildren>
    <WalletManager className='main__wallet-manager' />
  </IfAuthenticated>
  <IfAuthenticated hideChildren>
    <ConnectButton />
  </IfAuthenticated>
```

src/components/wallet-manager/index.tsx
```diff
@@ -38,6 +39,7 @@ export class Container extends React.Component<Properties, State> {
       currentAddress: value.address,
       connectionStatus: status,
       isWalletModalOpen,
+      isAuthenticated: !!state.authentication.user?.data && !!value.address,
```
AdamDotCom, line 42: Is it possible to get `isAuthenticated` from the context rather than the state in this situation? [context: { isAuthenticated }](https://github.com/zer0-os/zOS/pull/195/files#diff-ba9e5f36913253ad75371769e2dcf1ca405991e16351de5ce97f7394ff8a756eR20)

AdamDotCom (changes requested):

src/components/wallet-manager/index.tsx
hkhanouchi: I tried to get the `isAuthenticated` from the context but it is not possible since in the channel the context is used from the `AuthenticationContextProvider` added in the container of the channels app.

mrval5 (approved):

PR #246, opened 2022-11-09
Save the current viewMode in localStorage

hkhanouchi:
#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/theme-engine/index.tsx
+++ b/src/components/theme-engine/index.tsx
@@ -4,7 +4,7 @@ import { connectContainer } from '../../store/redux-container';
 
 import { ThemeEngine as ThemeEngineComponent, ViewModes } from '../../shared-components/theme-engine';
 import { setViewMode } from '../../store/theme';
-import { savedViewMode } from '../../store/theme/saga';
+import { keyStorageLightMode } from '../../store/theme/saga';
 
 export interface Properties {
   viewMode: ViewModes;
@@ -29,12 +29,12 @@ export class Container extends React.Component<Properties> {
   }
 
   componentDidMount() {
-    const viewMode = this.getSavedViewMode(savedViewMode) === 'true' ? ViewModes.Light : ViewModes.Dark;
+    const viewMode = this.isLightModeIsSaved(keyStorageLightMode) === 'true' ? ViewModes.Light : ViewModes.Dark;
     this.props.setViewMode(viewMode);
   }
 
-  getSavedViewMode = (savedViewMode: string): string => {
-    return localStorage.getItem(savedViewMode);
+  isLightModeIsSaved = (keyStorageLightMode: string): string => {
+    return localStorage.getItem(keyStorageLightMode);
   };
 
   get isDarkMode() {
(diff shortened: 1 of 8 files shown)


src/zns-route-connect.test.tsx
```diff
@@ -4,6 +4,7 @@ import { Container } from './zns-route-connect';
 import { Main } from './Main';
 import { Web3Connect } from './components/web3-connect';
 import { Authentication } from './components/authentication';
+import { ViewModes } from './shared-components/theme-engine';
```
mrval5, line 7: can you remove this line

src/store/theme/saga.ts
```diff
@@ -0,0 +1,17 @@
+import { takeLatest, put } from 'redux-saga/effects';
+import { SagaActionTypes, receive } from '.';
+import { ViewModes } from '../../shared-components/theme-engine';
+
+export const savedViewMode = 'viewMode:isLight';
+
+export function* setViewMode(action) {
+  const viewMode = action.payload;
+
+  localStorage.setItem(savedViewMode, viewMode === ViewModes.Dark ? 'false' : 'true');
```
mrval5, line 10:
```suggestion
  localStorage.setItem(savedViewMode, viewMode === ViewModes.Dark);
```

src/store/theme/saga.test.ts
```diff
@@ -0,0 +1,30 @@
...
+      },
+    };
+  });
+
+  it('sets viewMode theme', async () => {
+    const {
+      storeState: { value },
+    } = await expectSaga(setViewMode, { payload: ViewModes.Light }).withReducer(reducer).run();
+
+    expect(value.viewMode).toEqual(ViewModes.Light);
```
mrval5, line 28: can we test on localStorage if it's called with the right params

src/components/theme-engine/index.tsx
```diff
@@ -22,9 +25,18 @@ export class Container extends React.Component<Properties> {
...
-    return {};
+    return { setViewMode };
   }
 
+  componentDidMount() {
+    const viewMode = this.getSavedViewMode(savedViewMode) === 'true' ? ViewModes.Light : ViewModes.Dark;
+    this.props.setViewMode(viewMode);
+  }
+
+  getSavedViewMode = (savedViewMode: string): string => {
```
mrval5, line 36: this method returns either true or false as string, which is not what the name of the function describe.

can wee change it to `isLightModeIsSaved` (this is just an example). also to get rid of the logic of `'true'` in `componentDidMount`

src/store/theme/saga.ts
```diff
@@ -0,0 +1,17 @@
+import { takeLatest, put } from 'redux-saga/effects';
+import { SagaActionTypes, receive } from '.';
+import { ViewModes } from '../../shared-components/theme-engine';
+
+export const savedViewMode = 'viewMode:isLight';
```
mrval5, line 5:
```suggestion
export const keyStorageLightMode = 'viewMode:isLight';
```

mrval5 (changes requested):

src/store/theme/saga.ts
hkhanouchi: Argument of type 'boolean' is not assignable to parameter of type 'string'

PR #250, opened 2022-11-16
Realtime chat

mrval5:
### What does this do?
- add realtime chat when user is authenticated.
- remove channel sync when user authenticated
- add loader indicator when chat is not connected

#### Things to consider:
  1. Does this change any APIs?


hkhanouchi (approved):

src/lib/chat/chat-message.ts
```diff
@@ -0,0 +1,125 @@
...
+    };
+  } catch (err) {}
+
+  return chatMember;
+};
+
+function getSender(sendbirdUser, isAdmin) {
+  if (sendbirdUser) {
+    return getChatMember(sendbirdUser);
+  } else if (isAdmin) {
```
AdamDotCom, line 45: Wast this `isAdmin` condition carried over from ZERO Web? I think we can remove it from zOS.

AdamDotCom (approved):

PR #252, opened 2022-11-17
add private channel

mrval5:
### What does this do?
Screenshot 2022-11-17 at 13 22 19
Screenshot 2022-11-17 at 13 09 07

#### Things to consider:
  1. Does this change any APIs?


AdamDotCom:
I think we're missing the channel refresh portion of this task. When I'm not connected and view the Channels it is expected that I do not see the private channel, when I am connected and view the Channels I should see the private channels without having to refresh the page (video attached).

[private-channel-requires-reload.webm](https://user-images.githubusercontent.com/31045/202560833-97f1de42-e001-4099-a638-ed6c1ac59517.webm)

PR #253, opened 2022-11-21
Add join channel button

hkhanouchi:
Screen Shot 2022-11-21 at 18 13 01


src/store/channels/api.ts
```diff
@@ -1,7 +1,13 @@
-import { get } from '../../lib/api/rest';
+import { get, post } from '../../lib/api/rest';
 
 export async function fetchUsersByChannelId(channelId: string): Promise<any> {
   const response = await get<any>(`/chatChannels/${channelId}/members`);
 
   return response.body;
 }
+
+export async function joinChannel(channelId: string): Promise<any> {
```
mrval5, line 9: can we change the any type, I know we have it already in the above method - let's try to get rid of it

AdamDotCom (approved):
Hi @hkhanouchi, I added a commit to update the channel's `hasJoined` state after a user has joined so that the message input displays without refreshing. 
[join-channel.webm](https://user-images.githubusercontent.com/31045/204338228-17433254-5b5a-4b40-8c41-a603afaaa4c4.webm)

PR #258, opened 2022-12-01
Prompt create account

mrval5:
### What does this do?
Screenshot 2022-12-01 at 08 44 09

Use can change the display name and image profile.

### How do I test this?
- disconnect all wallets that connected to the website - in the same browser
- create a new address
- connect
Expected : you should see the prompt

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/account/create.tsx
+++ b/src/components/account/create.tsx
@@ -11,6 +11,7 @@ import { ImageUpload } from '../image-upload';
 import { Button } from '@zer0-os/zos-component-library';
 
 import './styles.scss';
+import classNames from 'classnames';
 
 export interface Properties {
   currentAddress: string;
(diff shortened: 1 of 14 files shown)


hkhanouchi (approved):

AdamDotCom (approved):

src/components/account/create.tsx
```diff
@@ -41,42 +54,144 @@ export class Container extends React.Component<Properties> {
...
+      .then((response) => response)
+      .catch((error) => {
+        const {
+          response: {
+            body: { code },
+          },
+        } = error;
+
+        if (/USER_HANDLE_ALREADY_EXISTS/.test(code)) {
+          console.log('Sorry that user handle has already been taken');
```
AdamDotCom, line 131: I think we're going to need an error state on the `handle` field. A handle must be unique, the `createAndAuthorize` will fail if there's a duplicate handle.

AdamDotCom (approved):

AdamDotCom (approved):

PR #279, opened 2023-01-05
Add send link preview

hkhanouchi:
Screen Shot 2023-01-07 at 00 31 51

### DIFF
--- a/src/platform-apps/channels/message.tsx
+++ b/src/platform-apps/channels/message.tsx
@@ -175,9 +175,9 @@ export class Message extends React.Component<Properties, State> {
 
   renderMessageWithLinks(): React.ReactElement {
     const { message } = this.props;
-    const hasLinks = linkifyjs.test(message);
+    const hasLinks = linkifyjs.find(message);
 
-    if (hasLinks) {
+    if (hasLinks.length) {
       return (
         <Linkify
           options={{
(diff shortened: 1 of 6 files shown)


src/store/messages/api.ts
```diff
@@ -46,3 +47,11 @@ export async function uploadFileMessage(channelId: string, media: File): Promise
 
   return response.body;
 }
+
+export async function getLinkPreviews(link: string): Promise<LinkPreview> {
+  const filter: any = {};
```
mrval5, line 52:
```suggestion
  const filter: any = {
    url: link
  };
```

mrval5 (approved):

src/platform-apps/channels/message.tsx
```diff
@@ -175,9 +175,9 @@ export class Message extends React.Component<Properties, State> {
 
   renderMessageWithLinks(): React.ReactElement {
     const { message } = this.props;
-    const hasLinks = linkifyjs.test(message);
+    const hasLinks = linkifyjs.find(message);
 
-    if (hasLinks) {
+    if (hasLinks.length) {
```
ratik21, line 180: Just to confirm, `hasLinks` will always be an array - right? This code will break if `hasLinks` is `undefined`/`null` otherwise

ratik21 (approved):

src/platform-apps/channels/message.tsx
hkhanouchi: Yes, `hasLink` will always be an array, the `linkifyjs.find(message)` return always an array.

PR #285, opened 2023-01-09
Remove development electron deploy workflow

dalefukami:
### What does this do?

Removes the CI workflow for deploying the development electron app.

This step has been failing for months which means it isn't doing anything useful.

### How do I test this?

Merging to development should no longer run this step.

#### Things to consider:
  1. How will this affect security? No
  1. How will this affect performance? Release performance will improve as we don't have to wait for this failing step
  1. Does this change any APIs? No


mrval5 (commented):
we shouldn't remove it, we should fix it.

maybe it will get priority soon

mrval5:
> we shouldn't remove it, we should fix it.
> 
> maybe it will get priority soon

I just read in the chat that we can kill it.

mrval5 (approved):

dalefukami:
> maybe it will get priority soon

Yeah, I agree, but it only takes 30 seconds to bring it back if it does receive priority so this is the quickest path to alleviating the problem for now. It's also possible it won't be a priority for another year.

PR #290, opened 2023-01-12
Add padding to AppContent components

dalefukami:
### What does this do?

1. Adds left padding (64px) to the AppContent component
2. Adds a css variable for apps to use to add right padding as necessary
3. Updates the Channels app to use the right hand padding

Note: PR to add the right padding to the Feed app to follow.

https://www.notion.so/zerotech/9a4ed77a5ea14501a3e6ef7a33f5a546?v=21e3add7413040bfb3eaf77e8c689cec&p=a2badc2a9c354fee800342544f121f8c&pm=s

We decided that, due to apps potentially having scrollbars, we'd enforce the 64px padding on the left side of the AppContent node but provide a simple mechanism for the apps to define where in their structure the 64px right padding should go. While not ideal, it's a simple compromise between flexibiity and consistency. Otherwise, the scrollbars would be seen 64px inside the right edge of the browser and that's not likely what design wants.

### How do I test this?

Load up zOS and verify that the Channel app has 64px of padding on the left and right sides.

#### Things to consider:
  1. How will this affect security? No
  1. How will this affect performance? N/A
  1. Does this change any APIs? The layout conventions now include having to add right hand padding.

### DIFF
--- a/src/platform-apps/channels/channel-view.tsx
+++ b/src/platform-apps/channels/channel-view.tsx
@@ -186,27 +186,29 @@ export class ChannelView extends React.Component<Properties, State> {
           />
         )}
         <InvertedScroll className='channel-view__inverted-scroll'>
-          <div className='channel-view__name'>
-            <h1>Welcome to #{this.props.name}</h1>
-            <span>This is the start of the channel.</span>
+          <div className='channel-view__main'>
+            <div className='channel-view__name'>
+              <h1>Welcome to #{this.props.name}</h1>
+              <span>This is the start of the channel.</span>
+            </div>
+            {this.props.messages.length > 0 && <Waypoint onEnter={this.props.onFetchMore} />}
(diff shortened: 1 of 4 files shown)


src/platform-apps/channels/channel-view.tsx
```diff
@@ -186,27 +186,29 @@ export class ChannelView extends React.Component<Properties, State> {
           />
         )}
         <InvertedScroll className='channel-view__inverted-scroll'>
-          <div className='channel-view__name'>
-            <h1>Welcome to #{this.props.name}</h1>
-            <span>This is the start of the channel.</span>
+          <div className='channel-view__main'>
+            <div className='channel-view__name'>
```
dalefukami, line 190: The rest of this file is just indentation changes. View without whitespace changes for a nicer picture.

AdamDotCom (approved):

PR #301, opened 2023-01-16
Shallow render in autocomplete-dropdown tests

dalefukami:
### What does this do?

Improves the performance of the autocomplete-dropdown test by not fully rendering the complicated nested component.

Before:
before

After:
after

While having the fastest possible tests isn't an end goal in itself, it is a factor in the developer experience. This is a simple change that improves some of our slowest tests.

### How do I test this?

Run the automated tests.


ratik21 (approved):
lgtm

PR #302, opened 2023-01-16
Refactor: autocomplete-dropdown tests"

dalefukami:
### What does this do?

Refactoring of the autocomplete-dropdown tests.

I found the tests difficult to follow while I was working on the performance. This makes a few things a little easier to undestand, IMO.

### How do I test this?

Run the automated tests.

### DIFF
--- a/src/components/autocomplete-dropdown/autocomplete-dropdown.test.tsx
+++ b/src/components/autocomplete-dropdown/autocomplete-dropdown.test.tsx
@@ -159,14 +159,7 @@ describe('autocomplete-dropdown', () => {
 
     wrapper.update();
 
-    const option = wrapper
-      .find(Result)
-      .findWhere((n) => {
-        const match = n.prop('item') as any;
-        return match.value === valueToSelect;
-      })
-      .first();
-    option.shallow().find('.autocomplete-dropdown-item').simulate('mouseDown', inputEvent());
+    selectOption(wrapper, valueToSelect);
 
     expect(onSelect).toHaveBeenCalledWith(searchResults[0]);
   });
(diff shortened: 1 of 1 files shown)


src/components/autocomplete-dropdown/autocomplete-dropdown.test.tsx
```diff
@@ -49,59 +48,27 @@ describe('autocomplete-dropdown', () => {
...
-          route: 'result-second-route',
-        },
-      ];
-    };
-
-    const wrapper = subject({ findMatches });
+    const searchResults = [
+      stubResult('first'),
+      stubResult('second'),
+    ];
```
ratik21, line 54: maybe if this used widely we can do a function?
```js
const getSearchResults = (n) => {
 const r = []
 return forLoop1ToN( r.push( stubResult(..) )
}
 ```

File: src/components/autocomplete-dropdown/autocomplete-dropdown.test.tsx [hunk 2]
```diff
@@ -457,3 +287,49 @@ function inputEvent(attrs = {}) {
     ...attrs,
   };
 }
+
+async function performSearch(dropdown, searchString) {
+  const input = dropdown.find('input');
+  // Fake the timers because we debounce search requests
```
ratik21, line 293: thanks for adding comments 👍

ratik21 (commented):
lgtm.

src/components/autocomplete-dropdown/autocomplete-dropdown.test.tsx
dalefukami: Good idea. I was thinking about that too so I think you've given me a good enough reason to extract!

PR #305, opened 2023-01-17
Add a CI action to run the build on PRs

dalefukami:
### What does this do?

Adds a CI action to run the project build, lint, and type checking on PRs

1. We have Lint rules to keep our code consistent
2. We use type checking to try to reduce certain classes of bugs
3. If it can't build then we don't want to merge it

All of these things are important parts of maintaining the project so we want to prevent errors from reaching the main branches.

### How do I test this?

Each step has been verified in this PR as failing when appropriate and will be successful prior to merging.

### DIFF
--- a/.github/workflows/run-build.yml
+++ b/.github/workflows/run-build.yml
@@ -1,0 +1,18 @@
+name: run-build
+on:
+  pull_request: {}
+
+jobs:
+  run-build:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


dalefukami:
Fixes merged and now all checks are correctly passing

mrval5 (approved):

hkhanouchi (approved):

PR #309, opened 2023-01-18
Fix real time receive messages

hkhanouchi:
#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/store/messages/saga.ts
+++ b/src/store/messages/saga.ts
@@ -277,16 +277,19 @@ export function* stopSyncChannels(action) {
 
 export function* receiveNewMessage(action) {
   const { channelId, message } = action.payload;
+  const channelIdWithoutPrefix = channelId.replace('sendbird_group_channel_', '');
 
-  const cachedMessageIds = [...(yield select(getCachedMessageIds(channelId)))];
-  const currentMessages = yield select(rawMessagesSelector(channelId));
+  const cachedMessageIds = [...(yield select(getCachedMessageIds(channelIdWithoutPrefix)))];
+  const currentMessages = yield select(rawMessagesSelector(channelIdWithoutPrefix));
 
   let messages = [];
+  let cachedMessageLength = 0;
 
   if (cachedMessageIds.length) {
     messages = [
       ...currentMessages,
     ];
+    cachedMessageLength = cachedMessageIds.length;
     const firstCachedMessageId = cachedMessageIds[0];
 
     messages = messages.map((messageId) => {
@@ -297,7 +300,9 @@ export function* receiveNewMessage(action) {
         return messageId;
       }
     });
-  } else {
+  }
+
+  if (cachedMessageLength === cachedMessageIds.length) {
     messages = [
       ...currentMessages,
       message,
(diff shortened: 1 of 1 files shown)


src/store/messages/saga.ts
```diff
@@ -277,16 +277,19 @@ export function* stopSyncChannels(action) {
 
 export function* receiveNewMessage(action) {
   const { channelId, message } = action.payload;
+  const channelIdWithoutPrefix = channelId.replace('sendbird_group_channel_', '');
```
AdamDotCom, line 280: Can we move the replacement logic to `src/lib/chat/index.ts`? In the coming weeks we have ambitions to replace the message notification event system. It would be nice to isolate our Sendbird specific logic to a single file.

AdamDotCom (changes requested):
We recently moved from channel.url to channel.id which was the source of this bug. Could you look into other real-time Sendbird events (Delete, Receive)?

src/store/messages/saga.ts
```diff
@@ -277,16 +277,19 @@ export function* stopSyncChannels(action) {
...
 
-  const cachedMessageIds = [...(yield select(getCachedMessageIds(channelId)))];
-  const currentMessages = yield select(rawMessagesSelector(channelId));
+  const cachedMessageIds = [...(yield select(getCachedMessageIds(channelIdWithoutPrefix)))];
+  const currentMessages = yield select(rawMessagesSelector(channelIdWithoutPrefix));
 
   let messages = [];
+  let cachedMessageLength = 0;
```
dalefukami, line 286: I might be missing something but why do we need this variable? It seems to me we changed an if/else statement into an "if...and then also do"...why not just move the "also" part into the first if statement? (and was the `else` part of the statement the bug?)

dalefukami (commented):
Just a general question here, I would have normally expected some tests to have to change for something like this. Do write tests for these sagas?

src/store/messages/saga.ts
```diff
@@ -277,16 +277,19 @@ export function* stopSyncChannels(action) {
...
 
   if (cachedMessageIds.length) {
     messages = [
       ...currentMessages,
     ];
+    cachedMessageLength = cachedMessageIds.length;
```
dalefukami, line 292: I realize this isn't really part of your change but why do line 298 and line 300 return different types of things?
```rb
if (messageId === firstCachedMessageId) {
        cachedMessageIds.shift();
        return message;
      } else {
        return messageId;
      }
```
One is a message object and one is a messageId. That seems like it could be a source of bugs too.

hkhanouchi:
> Just a general question here, I would have normally expected some tests to have to change for something like this. Do we write tests for these sagas?

Yes, normally we write tests, but here I made a rollback on the saga file and made the changes on another file

src/store/messages/saga.ts
hkhanouchi: We can create a task for a refactor.

dalefukami:
BTW, please consider my comments non-blocking for the next couple of weeks while I get a more clear understanding of the team, etc. I'll use the request changes feature if I see something critical.

AdamDotCom (approved):

PR #317, opened 2023-01-19
Add concurrency protection to github actions

dalefukami:
### What does this do?

Adds concurrency protection to github actions to ensure only one run is happening at a time.

When merging PRs to developemnt, for example, we don't want to have competing deploys as they may collide and the timing would be indeterminate.

### DIFF
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -3,6 +3,9 @@ on:
   push:
     branches:
       - main
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
 jobs:
   validate:
     runs-on: ubuntu-latest
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -3,6 +3,9 @@ on:
   release:
     types:
       - published
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
 jobs:
   deploy:
     runs-on: ubuntu-latest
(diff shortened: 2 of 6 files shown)


.github/workflows/development.yml
```diff
@@ -3,6 +3,9 @@ on:
   push:
     branches:
       - main
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
```
dalefukami, line 8: For deploying we should wait for the currently executing action to finish rather than cancel mid-deploy and leave things in a strange state.

.github/workflows/run-build.yml
```diff
@@ -2,6 +2,10 @@ name: run-build
 on:
   pull_request: {}
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref }}
+  cancel-in-progress: true
```
dalefukami, line 7: For PR related runs, we might as well cancel the previously running action so we can get our most recent results as quickly as possible (and save on run minutes)

.github/workflows/production.yml
```diff
@@ -3,6 +3,9 @@ on:
   release:
     types:
       - published
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
```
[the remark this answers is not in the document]
dalefukami: Yep, there are various ways to reduce duplication but not for these types of configurations. You can move jobs to separate files but for defining the actual workflow (eg: "on release published") you have to have these files separate with their specific configuration.

PR #334, opened 2023-01-24
Direct message chat

mrval5:
### What does this do?
Direct messages
Task: https://wilderworld.atlassian.net/browse/ZOS-69
Demo: https://www.loom.com/share/9ba37998cc324e78a34b20d805b4582d

#### Things to consider:
  1. Does this change any APIs?


src/platform-apps/channels/direct-message-chat/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+      directMessage: list.find((directMessage) => directMessage.id === activeDirectMessageId),
+    };
+  }
+
+  static mapActions(): Partial<Properties> {
+    return { setActiveDirectMessageId };
+  }
+
+  handleClose = (): void => {
+    this.props.setActiveDirectMessageId(null);
```
hkhanouchi, line 44: Why are we sending null in `setActiveDirectMessageId` ?  I think the specified type is **string**.                                       
 `setActiveDirectMessageId: (activeDirectMessageId: string) => void;`

src/components/sidekick/index.tsx
```diff
@@ -76,6 +87,45 @@ export class Container extends React.Component<Properties, State> {
     );
   }
 
+  renderTabs(): JSX.Element {
+    return (
+      <div className='sidekick__tabs'>
+        <IconButton
+          className='sidekick__tabs-network'
+          icon={Icons.Network}
+          onClick={this.clickTab.bind(this, Tabs.NETWORK)}
```
AdamDotCom, line 96: If we changed our clickTab method above can we get rid of the `bind`?
```
 clickTab = (tab: Tabs) =>
```
```
onClick={this.clickTab(Tabs.NETWORK)}
```

src/platform-apps/channels/direct-message-members/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+      )
+      .join(', ')
+      .trim();
+  }
+
+  renderMember = (directMessage: DirectMessage): JSX.Element => {
+    return (
+      <div
+        className='direct-message-members__user'
+        onClick={this.handleMemberClick.bind(this, directMessage.id)}
```
AdamDotCom, line 59: Another `bind`

src/store/direct-messages/saga.test.ts
```diff
@@ -0,0 +1,52 @@
...
+      .put(setDirectMessages(DIRECT_MESSAGES_TEST))
+      .withReducer(reducer)
+      .run();
+
+    expect(storeState).toMatchObject({
+      list: DIRECT_MESSAGES_TEST,
+    });
+  });
+
+  // it('should store direct messages', async () => {
```
AdamDotCom, line 34: Missing test?

src/store/direct-messages/api.ts
```diff
@@ -0,0 +1,8 @@
+import { get } from '../../lib/api/rest';
+import { Channel } from '../channels';
+import { DirectMessage } from './types';
+
+export async function fetchDirectMessages(): Promise<DirectMessage[]> {
+  const directMessages = await get<Channel[]>('/directMessages/mine');
+  return await directMessages.body;
```
AdamDotCom, line 7: Do we need to `return await` or can we `return directMessages.body`?

AdamDotCom (approved):
Looks great. How hard would it be to add autofocus to the DM message input when the DM window is opened?

src/components/sidekick/index.tsx
mrval5: if I remove the `.bind` and keep `onClick={this.clickTab(Tabs.NETWORK)}` it will call the method on each render and to solve it I have to do `onClick={() => this.clickTab(Tabs.NETWORK)}` which it will create an anonymous function on each render which do not want.

hkhanouchi (approved):

src/components/sidekick/index.tsx
dalefukami: > ...it will create an anonymous function on each render which do not want.

Just as an FYI, `bind` is creating an entirely new on every render too. So, using `bind` isn't really a solution to the problem you've mentioned.

mrval5: you are right, I was wrong about it, it's also written in react documentation.

@dalefukami for learning purposes..., do you have an idea how we can make it better in performance ?

dalefukami: There are a couple of strategies that I tend towards:
1. If the list is fairly short and static (like this case) I'll just write 3 separate functions on the class. `this.clickNetworkTab: () => this.clickTab(Tabs.NETWORK)`. They can all call the general case and they're super short so it's pretty obvious that they're just shortcut methods for this purpose.
2. If the list is dynamic then the main strategy is to cache the functions as they're created the first time. I saw this [used in zero-web](https://github.com/m3m3n70/zero-web/blob/development/src/components/gossip-feed/index.tsx#L88) and it's pretty similar to what I've done previously.

PR #335, opened 2023-01-25
Support cookie renaming via client headers

ratik21:
ZERO-API PR: [https://github.com/m3m3n70/zero-api/pull/874](https://github.com/m3m3n70/zero-api/pull/874)

### What does this do?

Pass an X-header (`X-APP-PLATFORM`) while making any request to zero-api.

### How do I test this?

Screenshot on zero-api PR is added

### DIFF
--- a/src/lib/api/attachment.ts
+++ b/src/lib/api/attachment.ts
@@ -7,7 +7,7 @@ interface AttachmentResponse {
 
 export async function getAttachmentUrl(attachment: { key: string }): Promise<string> {
   const filter: any = { key: attachment.key };
-  const attachmentResponse = await get<AttachmentResponse>(`/api/feedItems/getAttachmentDownloadInfo`, filter).catch(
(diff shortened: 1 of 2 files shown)


src/lib/api/attachment.ts
```diff
@@ -13,6 +13,7 @@ export async function getAttachmentUrl(attachment: { key: string }): Promise<str
     .query({
       key: attachment.key,
     })
+    .set('X-APP-PLATFORM', 'zos')
```
AdamDotCom, line 16: Could we change this method (`getAttachmentUrl`) and this file to use the api/rest.ts methods instead of manually setting the header? Then we don't have to manually set the header.

AdamDotCom (changes requested):
Could we change the `getAttachmentUrl` method to use the api/rest.ts `get` request

src/lib/api/attachment.ts
ratik21: Good idea.

hkhanouchi (approved):

AdamDotCom (approved):

PR #337, opened 2023-01-26
show unread count

mrval5:
### What does this do?
render unread messages :
- in front of each chat
- total on all unread messages in tab head

Screenshot 2023-01-26 at 17 07 01

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/sidekick/index.tsx
+++ b/src/components/sidekick/index.tsx
@@ -23,6 +23,7 @@ interface PublicProperties {
 export interface Properties extends PublicProperties {
   user: AuthenticationState['user'];
   updateLayout: (layout: Partial<AppLayout>) => void;
+  allUnreadMessages: number;
 }
 
 export interface State {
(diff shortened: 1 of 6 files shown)


src/platform-apps/channels/direct-message-members/index.test.tsx
```diff
@@ -49,4 +49,31 @@ describe('direct-message-members', () => {
 
     expect(setActiveDirectMessage).toHaveBeenCalledWith('292444273_bd035e84edfbaf11251ffef196de2ab47496439c');
   });
+
+  it('should not render unread messages', function () {
```
ratik21, line 53:
```suggestion
  it('should not render read messages', function () {
```

src/components/sidekick/index.tsx
```diff
@@ -36,10 +37,12 @@ export class Container extends React.Component<Properties, State> {
   static mapState(state: RootState): Partial<Properties> {
     const {
       authentication: { user },
+      directMessages: { list },
     } = state;
 
     return {
       user,
+      allUnreadMessages: list.reduce((count, directMessage) => count + directMessage.unreadCount, 0),
```
ratik21, line 45: to confirm - is this real time, or would the user have to reload to see unread count icons after receiving new message?

mrval5: we do not have realtime yet for dm's and also channels

hkhanouchi (approved):

PR #338, opened 2023-01-31
Use zero-api to search for mentionable users in a channel

dalefukami:
### What does this do?

This searches for mentionable users via the api endpoint rather than using the user list cached on the front end.

[New endpoint for mentions search](https://wilderworld.atlassian.net/browse/ZOS-74)

An endpoint has been created specifically for searching for mentionable users in a channel. Use it to be consistent with zero-web and other projects.

### How do I test this?

In a channel, try to @ mention someone. Ensure the suggestion box contains users you would expect to see.

#### Things to consider:
  1. How will this affect security? N/A
  1. How will this affect performance? Will be slower than the stubbed version we have now.
  1. Does this change any APIs? No


src/store/channels/api.ts
```diff
@@ -21,3 +21,14 @@ export async function markAllMessagesAsReadInChannel(channelId: string, userId:
...
   return response.status;
 }
+
+export type MentionableUser = {
+  id: string;
+  name: string;
+};
+
+export async function searchMentionableUsers(channelId: string, search: string): Promise<MentionableUser[]> {
+  return await get<any>(`/chatChannels/${channelId}/mentionable-users`, search)
```
mrval5, line 31: I think
```suggestion
  return await get<MentionableUser[]>(`/chatChannels/${channelId}/mentionable-users`, search)
```

src/platform-apps/channels/util/api.ts
```diff
@@ -0,0 +1,15 @@
+import { searchMentionableUsers } from '../../../store/channels/api';
+import { User } from '../../../store/channels';
+
+export async function searchMentionableUsersForChannel(
+  channelId: string,
+  search: string,
+  // When the api is updated to only return channel users we can remove this argument
+  validUsers: Array<User>,
+  apiSearch = searchMentionableUsers
+) {
```
mrval5, line 10: let's add return type

src/platform-apps/channels/message.tsx
```diff
@@ -206,6 +208,10 @@ export class Message extends React.Component<Properties, State> {
     }
   }
 
+  searchMentionableUsers = async (search: string) => {
+    return await searchMentionableUsersForChannel(this.props.channelId, search, this.props.users);
```
mrval5, line 212: I see that we are calling the api directly without using redux, any reason for that ?

all the api calls are in sagas

src/platform-apps/channels/message.tsx
```diff
@@ -206,6 +208,10 @@ export class Message extends React.Component<Properties, State> {
     }
   }
 
+  searchMentionableUsers = async (search: string) => {
```
mrval5, line 211: return type

src/platform-apps/channels/channel-view.tsx
```diff
@@ -172,6 +175,10 @@ export class ChannelView extends React.Component<Properties, State> {
     );
   }
 
+  searchMentionableUsers = async (search: string) => {
+    return await searchMentionableUsersForChannel(this.props.id, search, this.props.users);
```
mrval5, line 179: can we add unit test

src/platform-apps/channels/channel-view.tsx
```diff
@@ -172,6 +175,10 @@ export class ChannelView extends React.Component<Properties, State> {
     );
   }
 
+  searchMentionableUsers = async (search: string) => {
```
mrval5, line 178: this method is in 2 places, what do you think if we move it to `message-input`, and let last be the responsible of it?

I can see only one reason why is here which because we need 2 data `id` and `users` and it's not changing nothing on this component - a thought

mrval5 (changes requested):

src/store/channels/api.ts
dalefukami: Hmm.. I can't actually figure out where this impacts anything but, I agree, it's probably correct.

src/platform-apps/channels/util/api.ts
dalefukami: I prefer to let typescript infer the return type generally. Is there a particular reason to strictly define it?

src/platform-apps/channels/message.tsx
dalefukami: This data seems like it's completely isolated to the individual component and user action. I didn't think there was any reason to bring the complexity of global state into the picture.

src/platform-apps/channels/channel-view.tsx
dalefukami: `message-input` is a general input component but `channel-view` and `message` are context specific. If anyone else uses the `message-input` they'd want different ways to fetch the actual list of users they can mention in that context.

I can reduce the duplication by sending the function directly to the `message` component from the top level `channel-view` though. As far as I know, there's no reason for the `message` to fetch possible mentions differently.

PR #364, opened 2023-02-09
ZOS-98 use channels saga instead of directMessages

mrval5:
### What does this do?
- move `directMessages` store to `normalizer` `channels`.
- move `activeDirectMessageId` to `chat`
- remove `unreadCount: 0` when we fetch messages because it's not correct, we reset the count when we mark it as read

#### Things to consider:
  1. Does this change any APIs?


src/store/channels-list/saga.ts
```diff
@@ -1,59 +1,68 @@
-import { Channel } from './../channels/index';
+import { ChannelType } from './types';
+// import { Channel } from './../channels/index';
```
dalefukami, line 2: Nit: missed comment

hkhanouchi (approved):

PR #367, opened 2023-02-13
Feature/mark as read move dm

mrval5:
### What does this do?
- move directMessages store to normalizer channels.
- move activeMessengerId to chat
- remove unreadCount: 0 when we fetch messages because it's not correct, we reset the count when we mark it as read
- rename direct-message to messenger
- move messenger outside channel-app
- move common component between channel-app and messenger to root component

#### Things to consider:
  1. Does this change any APIs?


src/store/channels-list/saga.ts
```diff
@@ -1,59 +1,68 @@
-import { Channel } from './../channels/index';
+import { ChannelType } from './types';
+// import { Channel } from './../channels/index';
```
AdamDotCom, line 2: Commented out import

AdamDotCom (approved):

PR #375, opened 2023-02-23
make sidekick opening interactive with storage

mrval5:
### What does this do?
- store the state of sidekick if it's open or closed.
- in case no value in the storage, we open the sidekick and we store the value.

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/sidekick/index.test.tsx
+++ b/src/components/sidekick/index.test.tsx
@@ -12,11 +12,19 @@ describe('Sidekick', () => {
       className: '',
       updateSidekick: jest.fn(),
       setActiveSidekickTab: jest.fn(),
+      syncSidekickState: jest.fn(),
       ...props,
     };
 
     return shallow(<Container {...allProps} />);
   };
+
+  it('sync state', () => {
+    const syncSidekickState = jest.fn();
+    subject({ syncSidekickState });
+
+    expect(syncSidekickState).toHaveBeenCalled();
+  });
 
   it('adds className', () => {
     const wrapper = subject({ className: 'todo' });
--- a/src/components/sidekick/index.test.tsx
+++ b/src/components/sidekick/index.test.tsx
@@ -6,10 +6,21 @@ import { IfAuthenticated } from '../authentication/if-authenticated';
 import { MessengerList } from '../messenger/list';
 
 describe('Sidekick', () => {
+  beforeAll(() => {
+    global.localStorage = {
+      setItem: jest.fn(),
+      getItem: () => null,
+      removeItem: () => {},
+      length: 0,
+      clear: () => {},
+      key: (_) => '',
+    };
+  });
+
   const subject = (props: any = {}) => {
     const allProps = {
       className: '',
-      updateLayout: () => undefined,
+      updateSidekick: jest.fn(),
       ...props,
     };
 
@@ -24,6 +35,29 @@ describe('Sidekick', () => {
     expect(ifAuthenticated.find('.todo').exists()).toBe(true);
   });
 
+  it('open the sidekick in case not data found in storage', () => {
+    const updateSidekick = jest.fn();
+    subject({ updateSidekick });
+
+    expect(updateSidekick).toHaveBeenCalledWith({ isOpen: true });
+  });
+
+  it('renders sidekick with class animation in', () => {
+    const wrapper = subject({ isOpen: true });
+
+    const sidekick = wrapper.find('.sidekick');
+
+    expect(sidekick.hasClass('sidekick--slide-in')).toBe(true);
+  });
+
+  it('it should not render out class animation', () => {
+    const wrapper = subject({ isOpen: false });
+
+    const sidekick = wrapper.find('.sidekick');
+
+    expect(sidekick.hasClass('sidekick--slide-out')).toBe(false);
+  });
+
   it('renders sidekick panel', () => {
     const wrapper = subject();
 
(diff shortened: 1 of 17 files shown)


src/store/layout/index.ts
```diff
@@ -1,19 +1,16 @@
...
-export interface LayoutState {
-  value: AppLayout;
-}
+export const updateSidekick = createAction<UpdateSidekickPayload>(SagaActionTypes.updateSidekick);
 
 const initialState: LayoutState = {
   value: {
     isContextPanelOpen: false,
-    isSidekickOpen: true,
+    isSidekickOpen: false,
```
dalefukami, line 13: We could maybe read from localStorage here to set the correct initial state/default?

dalefukami (commented):
I may be interpreting the story incorrectly but should we also be making sure the correct tab is selected? Maybe you already have a follow up PR coming.

src/components/sidekick/index.test.tsx
```diff
@@ -33,13 +67,16 @@ describe('Sidekick', () => {
   });
 
   it('renders sidekick when panel tab is clicked', () => {
-    const updateLayout = jest.fn();
-    const wrapper = subject(updateLayout);
+    const updateSidekick = jest.fn();
+    const wrapper = subject({ updateSidekick });
 
     const ifAuthenticated = wrapper.find(IfAuthenticated).find({ showChildren: true });
```
AdamDotCom, line 73: We should be able to simplify the wrapper lookups by removing `ifAuthenticated`.

src/components/sidekick/index.test.tsx
```diff
@@ -33,13 +67,16 @@ describe('Sidekick', () => {
...
+
     ifAuthenticated.find('.app-sidekick-panel__target').simulate('click');
```
AdamDotCom, line 75:
```suggestion
    wrapper.find('.app-sidekick-panel__target').simulate('click');
```

src/components/sidekick/index.test.tsx
```diff
@@ -33,13 +67,16 @@ describe('Sidekick', () => {
...
 
-    expect(ifAuthenticated.find('.sidekick__slide-out').exists()).toBe(false);
+    expect(ifAuthenticated.find('.sidekick--slide-out').exists()).toBe(false);
```
AdamDotCom, line 77:
```suggestion
    expect(wrapper.find('.sidekick--slide-out').exists()).toBe(false);
```

src/components/sidekick/index.tsx
```diff
@@ -23,60 +24,74 @@ interface PublicProperties {
...
 
-  slideAnimationEnded = (): void => {
-    if (!this.state.isOpen) {
-      this.setState({ isOpen: false });
+  get isOpen() {
+    return this.props.isOpen;
+  }
+
+  componentDidMount() {
+    const sidekickOpenValue = localStorage.getItem(SIDEKICK_OPEN_STORAGE);
```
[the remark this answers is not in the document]
AdamDotCom: Agreed, we should be able to move the local storage access into the store. I suspect we'll have similar parts of the layout making use of this approach in the future.

src/store/layout/index.ts
AdamDotCom: Something like?

```
function resolveFromLocalStorage(key) {
  const value = localStorage.getItem(key);

  if (value === null || value === 'true') return true;

  return false;
}

const initialState: LayoutState = {
  value: {
    isContextPanelOpen: false,
    isSidekickOpen: resolveFromLocalStorage(SIDEKICK_OPEN_STORAGE),
    hasContextPanel: false,
  },
};

```

src/components/sidekick/index.test.tsx
```diff
@@ -6,10 +6,21 @@ import { IfAuthenticated } from '../authentication/if-authenticated';
...
 describe('Sidekick', () => {
+  beforeAll(() => {
+    global.localStorage = {
+      setItem: jest.fn(),
+      getItem: () => null,
+      removeItem: () => {},
+      length: 0,
+      clear: () => {},
+      key: (_) => '',
+    };
```
AdamDotCom, line 17: I suspect you'll have to mock the local store for all tests if you move the lookup into the store. I think you should be able to move this into the `src/setupTests.ts` file.

AdamDotCom (changes requested):
I noticed a sidekick animation flash. 

To reproduce:
- Connect, minimize the sidekick
- Disconnect
- Connect, observe that the sidekick opens and closes

[sidekick-flash.webm](https://user-images.githubusercontent.com/31045/221084859-7a63233c-a57b-4608-9c36-12f4c886ffb6.webm)


I wonder if you'll have to reset the `canStartAnimation` state when a user disconnects?

mrval5:
> I may be interpreting the story incorrectly but should we also be making sure the correct tab is selected? Maybe you already have a follow up PR coming.

the story has nothing about the tab. I can create another task - PR for the tab.

src/components/sidekick/index.tsx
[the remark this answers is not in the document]
mrval5: good remark, I will move it to the store.

src/lib/storage/index.ts
```diff
@@ -0,0 +1,7 @@
+export function resolveFromLocalStorageAsBoolean(key) {
+  const value = localStorage.getItem(key);
+
+  if (value === null || value === 'true') return true;
+
+  return false;
```
dalefukami, line 6:
```suggestion
  return (value === null || value === 'true');
```

dalefukami:
Sorry, I should have been more clear. That was my interpretation. Do it whichever way you want. ;)

mrval5:
> > the story has nothing about the tab
> 
> "When I open messenger, it should open in the same view as i last had it"
> 
> I would consider the tab I had selected to be part of the "same view"

not sure if you remember that we cloned the story (id 2, 214) to split the sidekick and messenger state to 2 stories
 @AdamDotCom explained during the daily that this sprint we will work on the sidekick.

I'm a bit confused here. maybe better to work on both stories in the same time.

AdamDotCom (approved):

PR #379, opened 2023-02-24
Start a conversation refinements

hkhanouchi:
Screen Shot 2023-02-24 at 20 04 06

https://user-images.githubusercontent.com/27851097/221269576-3b5adae1-fbd4-4ece-866c-efe0f0a42298.mov


AdamDotCom (changes requested):
This is so great. 

A couple minor requests. Could we use the same font style for the 'Conversations' and 'New message' headings? It looks like they are slightly different.
Screen Shot 2023-02-24 at 11 08 30 PM
Screen Shot 2023-02-24 at 11 08 40 PM

Also the 'Continue' button should be wider.
Screen Shot 2023-02-24 at 11 15 00 PM

Can we remove the `x` in the search?
Screen-Shot-2023-02-24-at-11-17-31-PM

PR #386, opened 2023-03-01
Add recording voice message

hkhanouchi:
https://user-images.githubusercontent.com/27851097/222240238-a30e6595-2ff2-427a-bba2-7eceefd6ca0a.mov


AdamDotCom:
@hkhanouchi, it looks like the `uploadFileMessage` function should also work for uploading audio files: https://github.com/zer0-os/zOS/blob/main/src/store/messages/api.ts#L54

hkhanouchi:
> @hkhanouchi, it looks like the `uploadFileMessage` function should also work for uploading audio files: https://github.com/zer0-os/zOS/blob/main/src/store/messages/api.ts#L54 What happens when you send the audio file to that endpoint?

@AdamDotCom yes, that's what I thought too, but it gives me this error. 

Screen Shot 2023-03-01 at 14 39 37

hkhanouchi:
i have a Conflicting peer dependency `peer react@"16.x" from react-mic@12.4.6`, I tried overriding the dependency but it still doesn't work. if you have any idea i am listening.

src/components/message-input/index.test.tsx
```diff
@@ -96,6 +100,18 @@ describe('MessageInput', () => {
     expect(dropzone.find(ReplyCard).exists()).toBe(true);
   });
 
+  it('renders MessageAudioRecorder', function () {
+    const wrapper = subject({});
+    const dropzone = wrapper.find(Dropzone).shallow();
+
+    expect(dropzone.find(MessageAudioRecorder).exists()).toBe(false);
+
+    (wrapper.instance() as any).startMic();
```
AdamDotCom, line 109:
```suggestion
    wrapper.find('.record-voice__icon').simulate('click')
```
In most of our other tests we're avoiding calling instance. Can we simulate the user event for click?

src/components/message-input/index.test.tsx
```diff
@@ -96,6 +100,18 @@ describe('MessageInput', () => {
...
+    const dropzone2 = wrapper.find(Dropzone).shallow();
```
AdamDotCom, line 110: Are we able to reuse the `dropzone` reference from above? I think `dropzone.update()` or `dropzone.setProps({})` should trigger a re-render.

AdamDotCom (changes requested):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "zOS",
-  "version": "0.10.0",
+  "version": "0.10.1",
```
AdamDotCom, line 3:
```suggestion
  "version": "0.10.0",
```
We shouldn't need to increase the version number unless we want to trigger a Production release.

AdamDotCom:
@hkhanouchi, how hard would it be to use a more modern audio recording component? If we moved to https://github.com/samhirtarif/react-audio-recorder or something more modern?

PR #388, opened 2023-03-03
Add emoji

mrval5:
### What does this do?
- add emojis to channel and messenger.

Screenshot 2023-03-03 at 10 14 01

#### Things to consider:
  1. Does this change any APIs?


src/components/message-input/emoji-picker.tsx
```diff
@@ -0,0 +1,76 @@
...
+  };
+
+  render() {
+    if (!this.props.isOpen) {
+      return null;
+    }
+
+    return (
+      <Picker
+        theme={this.props.viewMode}
```
dalefukami, line 69: Maybe we could do a mapping here from our mode to the Picker mode. I know they currently match exactly but we shouldn't assume that will always be the case.

src/components/message/text-to-emojis.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { Emoji } from 'emoji-mart';
+import nodeEmoji from 'node-emoji';
+
+// eslint-disable-next-line no-useless-escape
+const emojiRegex = new RegExp('(:[a-zA-Z0-9-_+]+:(:skin-tone-[2-6]:)?)', 'g');
+
+const fallback = (text) => () => text;
+
+export const textToEmojis = (message: string) => {
```
dalefukami, line 10: I'm not entirely sure we need this any more since we actually save the utf character and not the colon string.

src/components/message-input/index.tsx
```diff
@@ -121,6 +129,17 @@ export class MessageInput extends React.Component<Properties, State> {
         markup={userMentionsConfig.markup}
         displayTransform={userMentionsConfig.displayTransform}
       />,
+      <Mention
```
dalefukami, line 132: Does this need to be here? When we switched from saving the colon based token to just the utf character I thought we didn't need this any more.

src/components/message-input/emoji-picker.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+    if (emojiMart && event && event.target) {
+      if (!emojiMart.contains(event.target as Node)) {
+        this.props.onClose();
+      }
+    }
+  };
+
+  insertEmoji = (emoji: any) => {
+    const emojiToInsert = emoji.colons + ' ';
```
dalefukami, line 43: We should use the native character here and not the token.

dalefukami (changes requested):

src/components/message-input/index.tsx
dalefukami: https://github.com/m3m3n70/zero-web/blob/development/src/components/mentions-text-area/index.tsx#L84-L88

src/components/message-input/emoji-picker.tsx
dalefukami: https://github.com/m3m3n70/zero-web/blob/development/src/components/message-input/index.tsx#L255

dalefukami:
For reference, here's the PR where we made the switch to utf characters: https://github.com/m3m3n70/zero-web/pull/624

src/components/message-input/index.tsx
mrval5: when I remove it and write start typing `:+1:` it doesn't convert to emoji

src/components/message/text-to-emojis.tsx
mrval5: when I remove it we render emoji as plain text

src/components/message-input/emoji-picker.tsx
mrval5: good one.

src/components/message/text-to-emojis.tsx
dalefukami: Maybe on legacy messages where we saved the colon token but if we've saved the utf character then plain text is exactly what it is.

mrval5: hm, do you have an idea why we still have `textToEmojis` in zero-web?

dalefukami: I believe we left it in to smooth the transition between messages that had the colon token and the new utf messages. But, it's been 2 months now so we shouldn't need it any more.

Having said that, if you did try it and it still doesn't render the utf characters correctly, then just leave it in.

dalefukami (approved):

src/components/message/text-to-emojis.tsx
mrval5: you were right about this, I've deleted it.

PR #390, opened 2023-03-08
Insert gif from giphy

hkhanouchi:
#### Things to consider:
  1. Does this change any APIs?


AdamDotCom:
A Giphy image is a url link and a non native file so we don't need to upload it. We should be sending this message to the chatChannels/:id/message endpoint. Having said that the chatChannels/:id/message needs to be adjusted to accept the giphy params.

The params for sending a file url in the `chatChannels/:id/message` endpoint would be:
```
file: {
    url: string,
    file_name?: string,
    file_size?: string,
    file_type?: string,
}
```

hkhanouchi:
> A Giphy image is a url link and a non native file so we don't need to upload it. We should be sending this message to the chatChannels/:id/message endpoint. Having said that the chatChannels/:id/message needed to be adjusted to accept the giphy params. Related PR: [m3m3n70/zero-api#937](https://github.com/m3m3n70/zero-api/pull/937)
> 
> The call from the saga to send the image url might look like:
> 
> ```
> const file = { url: giphyFile.url, name: giphyFile.name, type: giphyFile.type };
> const messagesResponse = yield call(sendMessagesByChannelId, channelId, undefined, undefined, undefined, file);
> ```
> 
> Outstanding questions:
> 
> * ZERO Web support mentioning users in a Giphy message, do we have this functionality in zOS?  **_YES_**
> * ZERO Web supports sending text alongside a Giphy message, do we have this functionality in zOS?  **_YES_**
> * ZERO Web supports deleting a Giphy message, do we have this functionality in zOS?   **_YES_**
> * ZERO Web hides the 'reply' and 'edit' menu items for a Giphy message, do we have this functionality in zOS?   **_YES_**

AdamDotCom (changes requested):
Can you add a test in the `saga.test.ts` file for: 
1. sending a Giphy as it's own message
2. sending a Giphy with text  

The Giphy message is showing up twice in my local environment (screen capture included).

https://user-images.githubusercontent.com/31045/224128076-fbd8aa15-5b3f-4129-8885-1317261649a7.mov

PR #392, opened 2023-03-14
zOS: Link Preview is disappearing 

ratik21:
### What does this do?

Fixes the link disappearing issue.

Link preview was disappearing "immediately" after sending a message with a link preview. 

### How do I test this?

Try sending a message with link preview in zOS. It should not disappear now.


hkhanouchi (approved):

AdamDotCom (changes requested):
Nice! I can't believe it was that straight forward. Could we add a saga test too to make sure we don't remove this preview logic in the future? We should have some examples in src/store/messages/saga.test.ts. There's also some guidance at https://redux-saga.js.org/docs/advanced/Testing/#redux-saga-test-plan

ratik21:
@AdamDotCom added

AdamDotCom (approved):

PR #393, opened 2023-03-14
Add search list of conversations

hkhanouchi:
https://user-images.githubusercontent.com/27851097/225014337-fd27de80-8602-47dc-9199-21653f2b6c94.mov


src/components/messenger/search-conversations/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+  placeholder?: string;
+  directMessagesList?: Channel[];
+
+  onChange: (directMessages: Channel[]) => void;
+}
+
+interface State {}
+
+export class SearchConversations extends React.Component<Properties, State> {
+  searchConverstions = (search) => {
```
AdamDotCom, line 19: Typo, missing an `a`
```suggestion
  searchConversations = (search) => {
```

src/components/messenger/search-conversations/utils.ts
AdamDotCom, line 1: Can we remove this file and make use of the `otherMembersToString` method in src/platform-apps/channels/util/index.ts? The methods look to be the same.

src/components/messenger/list/index.tsx
```diff
@@ -101,6 +115,10 @@ export class Container extends React.Component<Properties, State> {
     return users.map((user) => ({ ...user, image: user.profileImage }));
   };
 
+  converstionInMyNetworks = (directMessagesList: Channel[]) => {
```
AdamDotCom, line 118: Typo missing an `a`
```suggestion
  conversationInMyNetworks = (directMessagesList: Channel[]) => {
```

AdamDotCom (changes requested):
This search feature is such a nice addition. Some very minor typos.

src/components/messenger/search-conversations/utils.ts
hkhanouchi: I prefer to keep this function in the component so that this component does not depend on another component.
Otherwise I added this function as a props to our component as an alternative solution

PR #405, opened 2023-03-16
Ability to remove link preview from message

ratik21:
image

After removing:

image

### DIFF
--- a/src/components/message/index.tsx
+++ b/src/components/message/index.tsx
@@ -16,12 +16,14 @@ import EditMessageActions from '../../platform-apps/channels/messages-menu/edit-
 import MessageMenu from '../../platform-apps/channels/messages-menu';
 import AttachmentCards from '../../platform-apps/channels/attachment-cards';
 import { textToPlainEmojis } from './text-to-emojis';
+import { IconXClose } from '@zero-tech/zui/icons';
+import { IconButton } from '../icon-button';
 
 interface Properties extends MessageModel {
   className: string;
   onImageClick: (media: any) => void;
   onDelete: (messageId: number) => void;
-  onEdit: (messageId: number, message: string, mentionedUserIds: User['id'][]) => void;
+  onEdit: (messageId: number, message: string, mentionedUserIds: User['id'][], data?: object) => void;
   onReply: (reply: ParentMessage) => void;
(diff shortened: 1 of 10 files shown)


src/components/message/index.tsx
```diff
@@ -248,11 +254,20 @@ export class Message extends React.Component<Properties, State> {
                     </div>
                   )}
                   {message && this.renderMessageWithLinks()}
-                  {preview && (
-                    <LinkPreview
-                      url={preview.url}
-                      {...preview}
-                    />
+                  {preview && hidePreview === false && (
```
hkhanouchi, line 257: `{preview && !hidePreview && (`

src/store/messages/api.ts
```diff
@@ -46,10 +46,11 @@ export async function editMessageApi(
   channelId: string,
   messageId: number,
   message: string,
-  mentionedUserIds: string[]
+  mentionedUserIds: string[],
+  data?: object
```
mrval5, line 50: what is data ? what is the structure of it that we are using in zOS ?

src/store/messages/saga.ts
```diff
@@ -200,13 +201,14 @@ export function* deleteMessage(action) {
...
-  const { channelId, messageId, message, mentionedUserIds } = action.payload;
+  const { channelId, messageId, message, mentionedUserIds, data } = action.payload;
+
   const selectedMessage = yield select(messageSelector(messageId));
   const existingMessages = yield select(rawMessagesSelector(channelId));
 
   const messages = existingMessages.map((id) => {
     if (messageId === id) {
-      return { ...selectedMessage, updatedAt: Date.now(), message };
+      return { ...selectedMessage, updatedAt: Date.now(), message, hidePreview: data?.hidePreview ?? false };
```
mrval5, line 211: Maybe:
```suggestion
      return { ...selectedMessage, updatedAt: Date.now(), message, hidePreview: Boolean(data?.hidePreview)};
```

src/store/messages/api.ts
mrval5: I understood it's like optional `options` which it contains `hidePreview`, le's add the type please

mrval5 (changes requested):

mrval5 (approved):

PR #421, opened 2023-03-20
Resolves the Sendbird Session Expiry issue

AdamDotCom:
### What does this do?

Resolves the Sendbird Session Expiry issue. This is a multi step phase, we also need to update the Sendbird library.

1. Sendbird Sessions were expiring without being renewed
2. Our Sendbird client was getting instantiated many times within our app
3. We never explicitly disconnected users which left connections open even after a user had disconnected

### How do I test this?

Run the tests, Connect as a user, open the browser console, wait 15minutes, observe that there are no Sendbird errors in the console, have someone send you a message, verify the realtime updates occur.


src/components/chat-connect/chat-connect.tsx
```diff
@@ -24,24 +24,34 @@ export interface Properties {
...
   isReconnecting: boolean;
+  context: {
+    isAuthenticated: boolean;
+  };
 }
 
 export class Container extends React.Component<Properties> {
-  static mapState(state: RootState): Partial<Properties> {
+  static mapState(state: RootState, props: Properties): Partial<Properties> {
+    if (!props.context.isAuthenticated) return {};
```
dalefukami, line 37: Does this mean we won't get live updates for public channels?

dalefukami (approved):

PR #429, opened 2023-03-21
Feature/turn browser notifications

hkhanouchi:
https://user-images.githubusercontent.com/27851097/226597206-116eabe2-aa95-4aca-a634-471df1f46d3f.mov

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2023,6 +2023,7 @@
       "version": "11.10.5",
       "resolved": "https://registry.npmjs.org/@emotion/react/-/react-11.10.5.tgz",
       "integrity": "sha512-TZs6235tCJ/7iF6/rvTaOH4oxQg2gMAcdHemjwLKIjKz4rRuYe1HJ2TQJKnAcRAfOUDdU8XoDadCe1rl72iv8A==",
+      "dev": true,
       "dependencies": {
         "@babel/runtime": "^7.18.3",
         "@emotion/babel-plugin": "^11.10.5",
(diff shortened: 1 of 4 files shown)


dalefukami:
Quick question, is the part that actually sends the notifications coming in a follow up PR? I think it would make sense for that part to come first. Otherwise we'll just have this dialog that does nothing.

Also, when you click Ok. Shouldn't it popup the browser option to enable the notifications if you haven't blocked it previously?

hkhanouchi:
> notifications to show up



> > This story has been planned since the last sprint
> 
> I'm not sure what you mean by this. Turning on notifications should cause notifications to show up. That's part of this story. That's why I'm asking.
> 

Yes, I know, but to explain the context to you, it's that the story to display the notifications was blocked when I started working on the story of turn on notifications. (Sprint 4)

Screen Shot 2023-03-21 at 12 58 08

dalefukami:
Sorry, I'm confused, that story is related to the notification list...not the browser notifications. They are different things. The browser notifications show different notifications. It's unrelated to the list we show in that notifications popup.

hkhanouchi:
> > This story has been planned since the last sprint
> 
> I'm not sure what you mean by this. Turning on notifications should cause notifications to show up. That's part of this story. That's why I'm asking.
> 
> > the 'OK' button comes from the zUI component of the Modal, so to deactivate or hide it we have to modify it in this component .
> 
> 2 things:
> 
> 1. We can modify the zUI component to allow removing the ok button.
> 2. But in this case, we want the button. If you click ok it should request access from the browser to send notifications shouldn't it? That's the common way these things work.

About this, we don't have the Ok button in the Design.
We'll just remove the ok button in zUI for now.

Screen Shot 2023-03-21 at 13 03 46

dalefukami:
Did the design folks realize that we can actually do the first step for the user? Did you reach out to them and ask?

hkhanouchi:
> For some more context, I just found this https://www.figma.com/file/aETpyuG2AKjNcQtCjldcX0/zOS-Messenger-%26-Notifications?node-id=6897-300637&t=QikDoaHv8mazpoe5-4
> 
> So the intent does appear that we'd at least try to enable them the first time. Not sure how this fits into the dialog flow though. You'd have to ask the design folks probably.

Awesome, thanks.

PR #438, opened 2023-03-22
Resolve: 'Warning: Setting defaultProps...' warning

AdamDotCom:
Removes console warning: 'Warning: Setting defaultProps as an instance property on Container is not supported and will be ignored. Instead, define defaultProps as a static property on Container.'

Screen Shot 2023-03-22 at 2 00 24 PM

### DIFF
--- a/src/app-sandbox/container.test.tsx
+++ b/src/app-sandbox/container.test.tsx
@@ -1,3 +1,7 @@
+/**
+ * @jest-environment jsdom
+ */
+
 import React from 'react';
 import { shallow } from 'enzyme';
 import { Container, Properties } from './container';
--- a/src/app-sandbox/container.test.tsx
+++ b/src/app-sandbox/container.test.tsx
@@ -1,7 +1,3 @@
-/**
- * @jest-environment jsdom
- */
-
 import React from 'react';
 import { shallow } from 'enzyme';
 import { Container, Properties } from './container';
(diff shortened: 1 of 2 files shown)


src/app-sandbox/container.test.tsx
```diff
@@ -1,3 +1,7 @@
+/**
+ * @jest-environment jsdom
```
dalefukami, line 2: Is this part necessary? jsdom is a lot slower, I believe.

dalefukami (approved):

src/app-sandbox/container.test.tsx
AdamDotCom: The test suite fails if I don't define it. 

Test output:
```
 FAIL  src/app-sandbox/container.test.tsx
  ● Test suite failed to run

    The error below may be caused by using the wrong test environment, see https://jestjs.io/docs/en/configuration#testenvironment-string.
    Consider using the "jsdom" test environment.
    
    ReferenceError: document is not defined

      43 | export class Container extends React.Component<Properties, State> {
      44 |   static defaultProps = {
    > 45 |     document,
         |     ^
      46 |   };
      47 |
      48 |   static mapState(state: RootState): Partial<Properties> {
```

dalefukami: Maybe we just shouldn't have a default?

hkhanouchi (approved):

PR #444, opened 2023-03-23
Fix change sessions if account metamask is changed

hkhanouchi:
https://user-images.githubusercontent.com/27851097/227242461-43eb4019-375b-44b8-a9a7-2025661caf91.mov

### DIFF
--- a/src/components/authentication/index.tsx
+++ b/src/components/authentication/index.tsx
@@ -1,12 +1,10 @@
 import React from 'react';
-import Web3Utils from 'web3-utils';
 
 import { RootState } from '../../store';
 import { connectContainer } from '../../store/redux-container';
 import { inject as injectWeb3 } from '../../lib/web3/web3-react';
 import { inject as injectProviderService } from '../../lib/web3/provider-service';
-import { ConnectionStatus, Connectors } from '../../lib/web3';
-import { config } from '../../config';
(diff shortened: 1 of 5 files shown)


src/components/authentication/index.tsx
```diff
@@ -71,39 +73,22 @@ export class Container extends React.Component<Properties, State> {
...
-        this.props.nonceOrAuthorize({ signedWeb3Token: res.result });
-      }
-    );
+  async authorize() {
+    const { personalSignToken, currentAddress, nonceOrAuthorize, updateConnector, clearSession, providerService } =
+      this.props;
+
+    await personalSignToken(providerService.get(), currentAddress)
+      .then((signedWeb3Token) => {
+        clearSession();
```
AdamDotCom, line 86: @hkhanouchi, I added a call to `clearSession` to reset the user state and chat access token before we authorize a new user. You can view the `clearSession` logic [here](https://github.com/zer0-os/zOS/blob/main/src/store/authentication/saga.ts#L36).

AdamDotCom (approved):
@hkhanouchi, I moved the `sendAsync` logic out of the component and moved some tests around, then added a call to `clearSession` before we authorize a new address. Could you verify that your initial fix still works as intended?

PR #466, opened 2023-03-29
Refactor: Split message list into separate components

dalefukami:
### What does this do?

Refactors the messenger list into separate panels so we can separate the flow logic from each step of the flow.

While adding the group conversation functionality this seemed like a good step to make adding the next screens simpler.

### How do I test this?

Verify the messenger list still works as expected with the exception of the existing bugs.

### DIFF
--- a/src/components/messenger/list/conversation-list-panel.tsx
+++ b/src/components/messenger/list/conversation-list-panel.tsx
@@ -132,18 +132,16 @@ export class ConversationListPanel extends React.Component<ConversationListPanel
         <div className='messages-list__direct-messages'>{this.renderNewMessageModal()}</div>
         {this.props.directMessagesList && (
           <div className='messages-list__items'>
-            <div className='messages-list__items-conversations'>
-              <div className='messages-list__items-conversations-input'>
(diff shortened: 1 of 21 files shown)


src/components/messenger/list/conversation-list-panel.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+import Tooltip from '../../tooltip';
+import { lastSeenText } from './utils';
+import { otherMembersToString } from '../../../platform-apps/channels/util';
+import { SearchConversations } from '../search-conversations';
+import { Channel } from '../../../store/channels';
+import { IconMessagePlusSquare, IconMessageQuestionSquare } from '@zero-tech/zui/icons';
+import { IconButton } from '../../icon-button';
+
+interface ConversationListPanelProperties {
```
dalefukami, line 12: There's a lot in this panel that I want to refactor but this felt like an ok step to get merged in for now. The property names were just replicated from the source component, moving the tests down a level to test this component independently, etc.

hkhanouchi (approved):

PR #478, opened 2023-04-04
[Bug] Parent Message disappears when you're replying to a message

ratik21:
https://user-images.githubusercontent.com/33264364/229768828-2c1a020d-b1d8-4239-b5ae-e6e75db0f7ce.mov

### DIFF
--- a/src/lib/chat/chat-message.ts
+++ b/src/lib/chat/chat-message.ts
@@ -102,12 +102,12 @@ function extractMessageData(jsonData, isMediaMessage) {
 }
 
 export function map(sendbirdMessage) {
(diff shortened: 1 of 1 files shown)


dalefukami (approved):
Is there a test for this that we could add to?

ratik21:
> Is there a test for this that we could add to?

Not sure, because the bug was caused by an event which we're receiving from sendbird (`receiveNewMessage` in `initChannelHandlers`). The data received was not mapped correctly.

dalefukami:
> > Is there a test for this that we could add to?
> 
> Not sure, because the bug was caused by an event which we're receiving from sendbird (`receiveNewMessage` in `initChannelHandlers`). The data received was not mapped correctly.

But, the `map` function is ours, right? We can test that we map it correctly now?

PR #479, opened 2023-04-04
Saga Error, Channels app continues to poll when a user is Disconnected which breaks the app

AdamDotCom:
### What does this do?

Fixes the Channel and Conversation Polling issue where if I 'Connect' / Authorize, then 'Disconnect' (I am now Anonymous) the poll continues to attempt to load my Channels and Conversations. The request for Conversations fail since we no longer have permissions to access it.

### How do I test this?

- 'Connect' a User, open the development console, wait two minutes, verify that calls to chatChannels and mine are being made
- 'Disconnect' the User, open the development console, wait two minutes, verify that no calls to chatChannels or mine are being made


src/store/authentication/saga.ts
```diff
@@ -82,6 +83,11 @@ function* setUserAndChatAccessToken(params: {
         isLoading,
       })
     ),
+    put({
+      type: user
+        ? ChannelsListSagaActionTypes.StartChannelsAndConversationsAutoRefresh
+        : ChannelsListSagaActionTypes.StopChannelsAndConversationsAutoRefresh,
```
dalefukami, line 89: Should these be separate things rather than the same one? I'm thinking...even if your not authorized you'd still want the channels to be updating because you can see the public ones and you'd want any new ones to show up.

Secondary question, does this mean we'll poll for channels even if the channel app isn't open? Do we care?

dalefukami (approved):
Love the use of a the cancellable saga stuff!

Also love the renames but, in the future, a lot of these could have been a separate initial PR. It would make the true changes simpler to review and...reverts are more isolated. Eg: if we had to revert this one..then we also lose all the good renaming.

src/store/authentication/saga.ts
dalefukami: Do the unread counts update?

PR #484, opened 2023-04-05
Add Group Details panel to Conversation pane

dalefukami:
### What does this do?

Adds the Group Details panel into the group conversation flow.

Note: For now, this just interjects the second panel in the flow always. Once the endpoint is built to search for pre-existing conversations with the same users we can skip the second step if possible.

A step towards allowing users to set the name and image for a group conversation.

### How do I test this?

Create group conversations

https://user-images.githubusercontent.com/43770/230182486-c19cebb8-1bbb-49c6-950f-bbe3e69db189.mp4

### DIFF
--- a/src/components/messenger/list/group-details-panel.tsx
+++ b/src/components/messenger/list/group-details-panel.tsx
@@ -1,0 +1,39 @@
+import * as React from 'react';
+
+import { Button } from '@zero-tech/zui/components';
+
+import { Option } from '../autocomplete-members';
+import { PanelHeader } from './panel-header';
+import { SelectedUserTag } from './selected-user-tag';
+
+import { bem } from '../../../lib/bem';
+const c = bem('group-details-panel');
+
+export interface Properties {
+  users: Option[];
+
+  onBack: () => void;
+  onCreate: (data: { users: Option[] }) => void;
+}
+
+export class GroupDetailsPanel extends React.Component<Properties> {
+  createGroup = () => {
+    this.props.onCreate({ users: this.props.users });
+  };
+
+  render() {
+    return (
+      <>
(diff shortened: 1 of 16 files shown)


src/store/channels-list/api.ts
```diff
@@ -16,3 +16,11 @@ export async function createConversation(userIds: string[]): Promise<DirectMessa
   const directMessages = await post<Channel[]>('/directMessages').send({ userIds });
   return directMessages.body;
 }
+
+export async function fetchConversationsWithUsers(_userIds: string[]): Promise<any[]> {
+  // Simulate request timing
```
AdamDotCom, line 21: This looks like test / debug code?

src/components/messenger/list/group-details-panel.tsx
```diff
@@ -0,0 +1,43 @@
+import * as React from 'react';
+
+import { Button } from '@zero-tech/zui/components';
+
+import { Option } from '../autocomplete-members';
+import { PanelHeader } from './panel-header';
+import { SelectedUserTag } from './selected-user-tag';
+
+import { bem } from '../../../lib/bem';
+const c = bem('group-details-panel');
```
AdamDotCom, line 10: 🪄

AdamDotCom (changes requested):
Looks great. I think there's some debug code in src/store/channels-list/api.ts.

src/store/channels-list/api.ts
dalefukami: Yep, this is intentional because the endpoint doesn't exist yet. For now, the system behaves exactly the same as it did previously where it doesn't check for existing conversations. Once the endpoint exists this can be enabled.

PR #492, opened 2023-04-07
Zos 263

AdamDotCom:
#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/store/authentication/saga.ts
+++ b/src/store/authentication/saga.ts
@@ -13,6 +13,7 @@ import { User } from './types';
 import { clearUserLayout, initializeUserLayout } from '../layout/saga';
 import { fetch as fetchNotifications } from '../notifications';
 import { clearChannelsAndConversations } from '../channels-list/saga';
+import { clearNotifications } from '../notifications/saga';
 
 export interface Payload {
   signedWeb3Token: string;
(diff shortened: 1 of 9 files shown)


src/store/authentication/saga.ts
```diff
@@ -108,7 +115,13 @@ export function* initializeUserState(user: User) {
...
 
 export function* clearUserState() {
-  yield clearUserLayout();
+  yield all([
+    call(clearChannelsAndConversations),
+    call(clearMessages),
+    call(clearUsers),
+    call(clearNotifications),
+    call(clearUserLayout),
+  ]);
```
dalefukami, line 124: I think this is a great place where using saga channels will be very useful. This is fine for now but let's keep that in mind when we're in here next time.

src/store/notifications/index.ts
```diff
@@ -23,17 +23,14 @@ const listSlice = createNormalizedListSlice({
...
   'chat_dm_mention',
 ];
-export function denormalizeNotifications(state) {
-  const result = listSlice
-    .denormalize(state.notificationsList.value, state)
-    .filter((n) => relevantNotificationTypes.includes(n.notificationType));
 
-  return result;
+export function denormalizeNotifications(state) {
+  return listSlice.denormalize(state.notificationsList.value, state);
```
dalefukami, line 35: Was this change necessary for clearing the state? It seems easier to filter at this level rather than have to remember to filter from all the various ways we might receive data (eg: realtime notifications, etc)

dalefukami (approved):

PR #493, opened 2023-04-10
Manage the create conversation stage in saga

dalefukami:
### What does this do?

Refactors the messenger list to rely on the saga to manage the stage of conversation creation.

This is a step towards having the saga control the entire user flow.

We believe controlling the user journey via a saga will lead to better control over behaviour and reactions to failures, etc.

### How do I test this?

Test the whole flow of creating a conversation.

### DIFF
--- a/src/store/create-conversation/saga.ts
+++ b/src/store/create-conversation/saga.ts
@@ -1,6 +1,47 @@
-import { takeLatest, put, call } from 'redux-saga/effects';
-import { SagaActionTypes, setActive, setGroupCreating } from '.';
+import { takeLatest, put, call, select } from 'redux-saga/effects';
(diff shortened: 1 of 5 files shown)


src/store/create-conversation/saga.ts
```diff
@@ -1,15 +1,61 @@
...
+export function* reset(_action) {
+  yield put(setActive(false));
+  yield put(setGroupCreating(false));
+  yield put(setStage(Stage.None));
+}
+
+export const getStage = (state) => state.createConversation.stage;
+
+export function* back(_action) {
+  const currentStage = yield select(getStage);
```
AdamDotCom, line 20: 🏆

PR #497, opened 2023-04-12
Move management of full create conversation flow into saga

dalefukami:
### What does this do?

Pushes the management of the entire user journey of creating a new conversation into a saga.

Attempting a new pattern for managing a user journey.

### How do I test this?

Test the entire flow of creating various types of conversations.

### DIFF
--- a/src/store/create-conversation/saga.ts
+++ b/src/store/create-conversation/saga.ts
@@ -54,8 +54,10 @@ export function* createConversation(action) {
 }
 
 export function* saga() {
-  yield take(SagaActionTypes.Start);
-  yield startConversation();
+  while (true) {
+    yield take(SagaActionTypes.Start);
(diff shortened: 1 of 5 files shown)


src/store/create-conversation/saga.ts
```diff
@@ -65,26 +35,91 @@ export function* performGroupMembersSelected(action) {
...
+}
+
+export function* startConversation() {
+  try {
+    yield call(reset);
+    yield put(setStage(Stage.CreateOneOnOne));
+
+    let currentStage = Stage.CreateOneOnOne;
+    while (currentStage !== Stage.None) {
+      const { back, cancel, handlerResult } = yield race({
```
AdamDotCom, line 70: 🎉

AdamDotCom (approved):

PR #509, opened 2023-04-19
Enable browser level notifications

AdamDotCom:
### What does this do?

Publishes Pusher based Messenger Conversation type notifications to the browser.

### How do I test this?

- Verify that Browser Notifications are enabled for the site
- Connect to zOS as two different users in two different browser sessions
- Start a Messenger Conversation between the two users
- Mention either one of the users in the conversation
- Verify that a Browser notification appears

### Screen capture

https://user-images.githubusercontent.com/31045/233151462-e120bc65-7ed4-4719-966b-5d3320a2d131.mp4


dalefukami (commented):
I'm a little confused here. This is tying the browser notifications to the system level notifications but, as we discussed, the browser notifications should be based on a completely different event. I.E., any time a message is received.

I would have expected to see something more like zero-web: https://github.com/m3m3n70/zero-web/blob/9fa019ce5ded67a8b40493a7190ced7a5f816aa1/src/store/sagas/chat-messages.ts#L259-L260

src/store/messages/saga.ts
```diff
@@ -367,6 +372,14 @@ export function* clearMessages() {
   yield put(removeAll({ schema: schema.key }));
 }
 
+export function* sendBrowserNotification(channelId, message) {
+  const channel = yield select(rawChannelSelector(channelId));
```
dalefukami, line 376: This is assuming we have ALL the conversations already loaded. Is that the case? What if the conversation is past the first page of 100? What happens?

PR #510, opened 2023-04-19
Pusher level event updates: supports chat_dm_message_replied events, maps Profile Summary

AdamDotCom:
### What does this do?

Adds support for the chat_dm_message_replied system event, maps the origin user's profile summary to the store so it displays correctly in the notification modal.

### Screen capture
Screen Shot 2023-04-19 at 2 01 13 PM


src/components/notification/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+  } else if (notification.notificationType === 'chat_dm_message_replied') {
+    let senderName = displayName(notification.originUser?.profileSummary);
+    return {
+      id: notification.id,
+      createdAt: notification.createdAt,
+      body: `${senderName} replied to you in a conversation`,
+      originatingName: senderName,
+      originatingImageUrl: notification.originUser?.profileSummary?.profileImage,
+      isUnread: notification.isUnread,
+    };
```
ratik21, line 45: this is repeated (only `body` is different), maybe wrap in a function? 

```js
const send = (notification, body) => {
  return {
      id: notification.id,
      createdAt: notification.createdAt,
      body,
      originatingName: senderName,
      originatingImageUrl: notification.originUser?.profileSummary?.profileImage,
      isUnread: notification.isUnread,
    }
```    

Here you can use 
```js
return send(notification, `${senderName} replied to you in a conversation`);
```

src/components/notification/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+function displayName(profileSummary) {
+  let displayName = [
+    profileSummary?.firstName,
+    profileSummary?.lastName,
+  ]
+    .filter((e) => e)
+    .join(' ');
+  displayName = displayName || 'Someone';
```
ratik21, line 58: What's the difference b/w using `||` & `??` ? 🤔 I tend to always use `??`

dalefukami: The difference, I believe, is `??` checks for `null` or `undefined` and then falls back to the second one whereas `||` checks for the javascript "falsy":

Eg:
`0 ?? 16` would evaluate to 0
`0 || 16` would evaluate to 16

So, it's a matter of whether you want to consider falsy values as a reason to use the default or if you only want to know if there's a value at all.

ratik21: ah, okay. So i am assuming `'' ?? 'some text'` would also return `''`? Thanks for sharing!

dalefukami: image

Correct!

ratik21 (approved):

PR #523, opened 2023-04-26
Disassociate having a web3 address with being authenticated

AdamDotCom:


dalefukami (approved):
Nice, that was simple.

PR #527, opened 2023-04-26
Hide cloudinary provider creation behind a function

dalefukami:
### What does this do?

This hides the singleton cloudinary provider behind a getter function.

Previously, this object was instantiated as soon as the file was loaded. This is a side-effect we'd like to avoid.

### How do I test this?

Run the tests
Smoke test the app where we have images displayed

### DIFF
--- a/src/lib/cloudinary/provider.tsx
+++ b/src/lib/cloudinary/provider.tsx
@@ -1,7 +1,17 @@
 import { CloudinaryProvider } from '@zer0-os/zos-component-library';
 import { config } from '../../config';
 
-export const provider = new CloudinaryProvider({
-  cloud_name: config.cloudinary.cloud_name,
-  max_file_size: config.cloudinary.max_file_size,
-});
+let provider;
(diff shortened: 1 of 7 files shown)


src/lib/cloudinary/provider.tsx
```diff
@@ -1,7 +1,17 @@
...
-export const provider = new CloudinaryProvider({
-  cloud_name: config.cloudinary.cloud_name,
-  max_file_size: config.cloudinary.max_file_size,
-});
+let provider;
+export function getProvider() {
+  if (!provider) {
+    provider = createProvider();
+  }
+  return provider;
```
ratik21, line 9: maybe 
```js
return provider ?? createProvider();
```
?

ratik21 (approved):

src/lib/cloudinary/provider.tsx
dalefukami: Come on.. that's too clean. :laughing:

PR #535, opened 2023-04-28
Add a simple template script to generate components and store/saga files

dalefukami:
### What does this do?

Adds a script so you can run `node ./templates/generate.js component src/components/funny-button` and `node ./templates/generate.js saga src/store/interesting-saga`

I find I always create the same few files with a boilerplate starting point. This makes that slightly quicker.

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,6 +114,7 @@
     "husky": "^8.0.1",
     "jest-extended": "^1.2.0",
     "jest-when": "^3.5.2",
+    "mustache": "^4.2.0",
     "prettier": "^2.7.1",
     "prettier-plugin-multiline-arrays": "^1.0.0",
     "pretty-quick": "^3.1.3",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -114,6 +114,7 @@
     "husky": "^8.0.1",
     "jest-extended": "^1.2.0",
     "jest-when": "^3.5.2",
+    "mustache": "^4.2.0",
```
ratik21, line 117: we have a library named `mustache` 😂

templates/saga/index.ts.mustache
```diff
@@ -0,0 +1,31 @@
+import { createAction, createSlice, PayloadAction } from '@reduxjs/toolkit';
+
+// TODO: Add reducer to ../reducer.ts
+// TODO: Add saga to ../saga.ts
+
+export enum SagaActionTypes {
+  Start = '{{sagaName}}/start',
+}
+
+export type {{sagaName}}State = {
```
ratik21, line 10: just wondering, is it possible to pass the `sagaName` & `component name` from cli arguments (optional args)?

ratik21 (approved):
🚀

templates/saga/index.ts.mustache
dalefukami: Those get calculated based on the folder name.

`node ./templates/generate.js component src/components/funny-button` will create a component named `FunnyButton`, for example.

PR #561, opened 2023-05-10
Fixed reply overflow

AdamDotCom:
### What does this do?

Fixes the over flow when replying to a long message.

### Screen capture

Screen Shot 2023-05-10 at 11 29 13 AM
Screen Shot 2023-05-10 at 11 29 05 AM


dalefukami (approved):
Nice, glad it was an easy one!

PR #579, opened 2023-05-15
Zos 338: When not logged in all urls should direct to the login page

ratik21:


src/components/authentication/index.tsx
```diff
@@ -97,8 +108,14 @@ export class Container extends React.Component<Properties, State> {
       });
   }
 
+  redirectIfNotLoggedIn = () => {
+    if (!this.state.isLoggedIn) {
```
dalefukami, line 112: I think this is where we want to check a feature flag to decide if we should redirect to login or just stay here and show the public facing zOS.

ratik21: the user has to be logged in i think, because even if we have `allowPublicZOS = true`, user won't see any data because we don't have the "current" user in the context yet, which we need. 

I think `allowPublicZOS` should check against public view of zOS **after** the user has logged in.

dalefukami: No, we don't need the user in the context. If you hit zOS without being logged in you get to see all the other apps. You get to see the public wilder channels, etc.

dalefukami (approved):

PR #585, opened 2023-05-17
Open the initial conversation when a user first logs in after registering

dalefukami:
### What does this do?

Opens the initial conversation for the user when they've first completed the registration process.

### How do I test this?

Register as a new user and verify that once they've completed the registration process (i.e., set their Name) the app opens and the very first conversation is open and ready.

### DIFF
--- a/src/store/channels-list/channels.ts
+++ b/src/store/channels-list/channels.ts
@@ -1,0 +1,10 @@
+import { multicastChannel } from 'redux-saga';
+import { call } from 'redux-saga/effects';
+
+let theConversationsChannel;
+export function* conversationsChannel() {
+  if (!theConversationsChannel) {
+    theConversationsChannel = yield call(multicastChannel);
+  }
+  return theConversationsChannel;
+}
--- a/src/store/registration/saga.ts
+++ b/src/store/registration/saga.ts
@@ -1,4 +1,4 @@
-import { call, put, select, take } from 'redux-saga/effects';
+import { call, put, select, spawn, take } from 'redux-saga/effects';
 import {
   AccountCreationErrors,
   InviteCodeStatus,
@@ -21,6 +21,9 @@ import {
 import { fetchCurrentUser } from '../authentication/api';
 import { nonce as nonceApi } from '../authentication/api';
 import { passwordStrength } from '../../lib/password';
+import { conversationsChannel } from '../channels-list/channels';
+import { rawConversationsList } from '../channels-list/saga';
+import { setActiveMessengerId } from '../chat';
 
 export function* validateInvite(action) {
   const { code } = action.payload;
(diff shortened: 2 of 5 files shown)


src/store/channels-list/channels.ts
```diff
@@ -0,0 +1,10 @@
+import { multicastChannel } from 'redux-saga';
+import { call } from 'redux-saga/effects';
+
+let theConversationsChannel;
+export function* conversationsChannel() {
+  if (!theConversationsChannel) {
+    theConversationsChannel = yield call(multicastChannel);
```
ratik21, line 7: what does calling `multicastChannel` do?

src/store/registration/saga.ts
```diff
@@ -140,4 +143,21 @@ export function* saga() {
...
+
+export function* channelsLoaded() {
+  const existingConversationsList = yield select(rawConversationsList());
+  if (existingConversationsList.length > 0) {
+    yield put(setActiveMessengerId(existingConversationsList[0]));
+  }
+}
+
+function* conversationsWatcher() {
+  const channel = yield call(conversationsChannel);
```
ratik21, line 158: i don't follow this part. This is calling a saga function `conversationsChannel` which is calling a "multicast" channel. Is it fetching/loading our channel(s) info?

src/store/registration/saga.ts
```diff
@@ -140,4 +143,21 @@ export function* saga() {
...
+  const payload = yield take(channel, '*');
+  if (payload.loaded) {
+    yield call(channelsLoaded);
+  }
 }
```
ratik21, line 163: to clarify, `yield take(channel, '*')` will "listen" for all events were the "channel" redux state is updated, and is checking if `{ loaded: true }` is set? (after which it sets the active messenger id)

ratik21 (approved):
had a few doubts

src/store/channels-list/channels.ts
dalefukami: `multicastChannel` is a redux-saga thing. It allows you to publish an event that multiple listeners can "take" from.

src/store/registration/saga.ts
dalefukami: In this line we're just getting access to the channel. The next line "listens" for events on that channel with the `take` call in the same way all the other `take` calls work. When an event comes it'll move on.

src/store/registration/saga.ts
dalefukami: Not really the "redux state". It's events that come across that specific channel. In this case, it's the conversationsChannel that we're listening to.

The `{loaded: true}` is just the value I chose to use when publishing the event. I decided this conversation event was a "conversations loaded" event and this seemed an ok way to indication that. Another option would be to publish it more like an action payload: `{type: "CONVERSATIONS_LOADED"}`. Maybe that would make more sense?

PR #590, opened 2023-05-19
Update error states in `authentication/create-email-account`

ratik21:


src/store/registration/index.ts
```diff
@@ -41,6 +41,8 @@ export enum AccountCreationErrors {
   PASSWORD_REQUIRED = 'PASSWORD_REQUIRED',
   PASSWORD_TOO_WEAK = 'PASSWORD_TOO_WEAK',
   PASSWORD_INVALID = 'PASSWORD_INVALID',
+  PROFILE_PRIMARY_EMAIL_REQUIRED = 'PROFILE_PRIMARY_EMAIL_REQUIRED',
+  PROFILE_PRIMARY_EMAIL_ALREADY_EXISTS = 'PROFILE_PRIMARY_EMAIL_ALREADY_EXISTS',
```
ratik21, line 45: note: excluded `handle required or exists` error from the api since we're passing the email as the handle.

dalefukami (approved):
Thanks for adding these!

PR #597, opened 2023-05-23
Zos 301: Add tooltip popup to display rewards

ratik21:
+ Moved saga logic (`fetchRewards`) from rewards popup component to the messenger-list level (since we need to show rewards in the tooltip)
+ added a tooltip-popup component in zOS (TODO: move to zUI)
+ added unit tests

image

image

### DIFF
--- a/src/components/messenger/list/index.tsx
+++ b/src/components/messenger/list/index.tsx
@@ -32,8 +32,11 @@ import { RewardsPopupContainer } from '../../rewards-popup/container';
 import { bem } from '../../../lib/bem';
 import classnames from 'classnames';
 import { enterFullScreenMessenger } from '../../../store/layout';
-import { Avatar } from '@zero-tech/zui/components';
+import { Avatar, Status } from '@zero-tech/zui/components';
 import { RewardsFAQModal } from '../../rewards-faq-modal';
+import { TooltipPopup } from '../../tooltip-popup/tooltip-popup';
+import { fetch as fetchRewards } from '../../../store/rewards';
+
 const c = bem('messenger-list');
 
 export interface PublicProperties {
@@ -52,6 +55,8 @@ export interface Properties extends PublicProperties {
   allowExpand: boolean;
   includeRewardsAvatar: boolean;
   userAvatarUrl: string;
+  zero: string;
+  isRewardsLoading: boolean;
 
   startCreateConversation: () => void;
   startGroup: () => void;
@@ -61,11 +66,14 @@ export interface Properties extends PublicProperties {
   membersSelected: (payload: MembersSelectedPayload) => void;
   createConversation: (payload: CreateMessengerConversation) => void;
   enterFullScreenMessenger: () => void;
+  fetchRewards: (_obj: any) => void;
 }
 
 interface State {
   isRewardsPopupOpen: boolean;
   isRewardsFAQModalOpen: boolean;
+  isRewardsTooltipOpen: boolean;
+  isRewardsPopupOpenedOnce: boolean; // to track if the user has viewed the rewards "once"
 }
 
 export class Container extends React.Component<Properties, State> {
@@ -75,6 +83,7 @@ export class Container extends React.Component<Properties, State> {
       registration,
       authentication: { user },
       layout,
+      rewards,
     } = state;
     const conversations = denormalizeConversations(state)
       .sort((messengerA, messengerB) =>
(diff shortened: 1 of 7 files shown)


src/components/messenger/list/index.tsx
```diff
@@ -171,34 +195,66 @@ export class Container extends React.Component<Properties, State> {
...
-        >
-          {this.props.includeRewardsAvatar && (
-            <Avatar size={'small'} type={'circle'} imageURL={this.props.userAvatarUrl} />
-          )}
-          <button
-            onClick={this.openRewards}
-            className={classnames(c('rewards-button'), {
-              [c('rewards-button', 'open')]: this.state.isRewardsPopupOpen,
-            })}
+        {this.props.includeRewardsAvatar ? ( // only show the rewards tooltip popup if in full screen mode
```
dalefukami, line 238: I think it'd be a good idea to use a separate flag here. You're assuming that the reason we want to show the rewards Avatar is the same reason we would want to show the popup. But, that may not always be the case. It's usually better to have the component render use very specific flags.

src/components/messenger/list/index.tsx
```diff
@@ -171,34 +195,66 @@ export class Container extends React.Component<Properties, State> {
...
 
+  stringifyZero() {
+    const stringValue = this.props.zero.padStart(19, '0');
+    const whole = stringValue.slice(0, -18);
+    const decimal = stringValue.slice(-18).slice(0, 4).replace(/0+$/, '');
+    const decimalString = decimal.length > 0 ? `.${decimal}` : '';
+    return `${whole}${decimalString}`;
+  }
+
+  renderRewardsBar() {
```
dalefukami, line 206: Wonder if it's worth pulling this into a sub-component?

src/components/messenger/list/index.tsx
```diff
@@ -171,34 +195,66 @@ export class Container extends React.Component<Properties, State> {
...
-            onClick={this.openRewards}
-            className={classnames(c('rewards-button'), {
-              [c('rewards-button', 'open')]: this.state.isRewardsPopupOpen,
-            })}
+        {this.props.includeRewardsAvatar ? ( // only show the rewards tooltip popup if in full screen mode
+          <TooltipPopup
+            open={this.props.isRewardsLoading === false && this.state.isRewardsTooltipOpen}
+            align='center'
+            side='left'
+            content={`You’ve earned ${this.stringifyZero()} ZERO today`}
```
dalefukami, line 243: Is this supposed to be the total or the latest daily amount?

dalefukami (approved):
A couple of refactoring possibilities but generally this is ok.

One main question I have is whether this is supposed to be a dynamic notification. If you happen to keep your browser open over night, should you see the new rewards notification when the new ones are calculated?

src/components/messenger/list/index.tsx
```diff
@@ -171,34 +195,66 @@ export class Container extends React.Component<Properties, State> {
...
-            <Avatar size={'small'} type={'circle'} imageURL={this.props.userAvatarUrl} />
-          )}
-          <button
-            onClick={this.openRewards}
-            className={classnames(c('rewards-button'), {
-              [c('rewards-button', 'open')]: this.state.isRewardsPopupOpen,
-            })}
+        {this.props.includeRewardsAvatar ? ( // only show the rewards tooltip popup if in full screen mode
+          <TooltipPopup
+            open={this.props.isRewardsLoading === false && this.state.isRewardsTooltipOpen}
```
dalefukami, line 240:
```suggestion
            open={!this.props.isRewardsLoading && this.state.isRewardsTooltipOpen}
```

src/components/messenger/list/index.tsx
```diff
@@ -111,11 +122,17 @@ export class Container extends React.Component<Properties, State> {
...
+      fetchRewards,
       enterFullScreenMessenger: () => enterFullScreenMessenger(),
     };
   }
 
-  state = { isRewardsPopupOpen: false, isRewardsFAQModalOpen: false };
+  state = {
+    isRewardsPopupOpen: false,
+    isRewardsFAQModalOpen: false,
+    isRewardsTooltipOpen: true, // initally open, will close after user clicks on 'x' button
```
dalefukami, line 133: This is going to be open every time you log in? Shouldn't that only happen if you have new rewards? Also, the very first time you register we're supposed to have the whole rewards popup open. Does this cause the tooltip and the modal to be open at the same time?

src/components/messenger/list/index.tsx
ratik21: done

src/components/messenger/list/index.tsx
ratik21: not sure if it's worth right now. I will make a note though, if any more changes, we can move this to a sub-component

src/components/messenger/list/index.tsx
ratik21: updated to latest daily amount 😅

src/components/messenger/list/index.tsx
ratik21: done, thanks

PR #599, opened 2023-05-24
zOS 290: add invite toast notification (onboarding flow)

domw30:
### What does this do?
Display an invite toast notification during first time login (after rewards pop up is closed) which shows how to invite friends.

So that a user is aware of how to invite friends/others so they can create a valuable network on the app.

### How do I test this?
Sign up as a new user and login for the first time. The rewards pop up will open automatically during first time login, close this pop up by clicking 'x' or outside of the pop up container, wait 10 seconds and confirm the invite toast notification appears. Click 'Invite' button to open the modal with invite dialog.

- added ToastNotification to zUI.
- added logic to display toast when rewards pop up is closed during first time login.
- import ToastNotification to zOS onboarding.
- added/updated first time login tests.

### DIFF
--- a/src/components/messenger/list/index.tsx
+++ b/src/components/messenger/list/index.tsx
@@ -32,7 +32,8 @@ import { RewardsPopupContainer } from '../../rewards-popup/container';
 import { bem } from '../../../lib/bem';
 import classnames from 'classnames';
 import { enterFullScreenMessenger } from '../../../store/layout';
-import { Avatar } from '@zero-tech/zui/components';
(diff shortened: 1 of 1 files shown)


domw30:
> Nice. I think there's room for some unit tests here and I was kind of thinking maybe we could pull the timing/business flow logic up to the saga level as it'd make it easier to change. Ratik and I are going to be in this component this afternoon so I'm going to merge this before hand. Then, perhaps we can pair up early next week and see if my ideas would pan out?

Sounds good to me @dalefukami, thanks. Let me know if you want me to merge 👍

PR #604, opened 2023-05-24
Add account creation method selection page

dalefukami:
### What does this do?

Registration Flow: Adds the page to select which type of account a user wants to create.

Note: Behind a feature flag for now until the wallet page is completed.

### DIFF
--- a/src/authentication/select-method/index.tsx
+++ b/src/authentication/select-method/index.tsx
@@ -1,0 +1,32 @@
+import * as React from 'react';
+import { Link } from 'react-router-dom';
+
+import { Button } from '@zero-tech/zui/components';
+
+import './styles.scss';
+import { bem } from '../../lib/bem';
+const c = bem('select-method');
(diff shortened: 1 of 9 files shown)


src/authentication/select-method/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+export interface Properties {
+  onEmailSelected: () => void;
+  onWalletSelected: () => void;
+}
+
+export class SelectMethod extends React.PureComponent<Properties> {
+  render() {
+    return (
+      <div {...cn()}>
+        <h3 {...cn('heading')}>How do you want to create an account?</h3>
```
dalefukami, line 20: Experimenting here with this classname shortcut. It's quite a bit shorter but, obviously, is non-standard so it takes a moment to understand what's going on if you haven't seen it before. Trying to find good ways to reduce the overhead of scanning a component's structure versus being able to "grep" the code for class names, etc.

```
<h3 {...cn('heading')}>
<h3 className={c('heading')}>
```

PR #608, opened 2023-05-26
feat: image upload improvements (colour, icon, text)

domw30:
### What does this do?
- fixes background colour of image upload component (all uses of component)
- removes hard coded svg icon and replaces with `icon` prop to make component more generic
- removes hard coded upload text and replaces with `uploadText` prop to make component more generic
- fixes text colour for group details use of image upload

To be aligned with designs - previously using outdated colours.
Also making the component more generic so it can be used elsewhere in the application and be customised. 

### How do I test this?
Run through the invite flow to Uploading an image step, or, create group flow to uploading an image step.

Screenshot 2023-05-26 at 10 48 26

Screenshot 2023-05-26 at 10 53 44

**As Per Design Request - Oval shape to Circular**
Before
Screenshot 2023-05-26 at 14 30 14

After
Screenshot 2023-05-26 at 14 29 39


src/components/image-upload/styles.scss
```diff
@@ -1,14 +1,14 @@
-@use '../../shared-components/theme-engine/theme' as theme;
+@use '~@zero-tech/zui/styles/theme' as theme;
```
dalefukami, line 1: :tada:

src/authentication/create-account-details/index.tsx
```diff
@@ -52,14 +54,20 @@ export class CreateAccountDetails extends React.Component<Properties, State> {
     return this.props.errors.image;
   }
 
+  renderImageUploadIcon = (): JSX.Element => <IconUpload2 isFilled size={24} color='#EDEDEF' />;
```
dalefukami, line 57: Do we need to use the `color` prop here or can we just style the upload component with the right color? They're all the same.

dalefukami (approved):

src/authentication/create-account-details/index.tsx
domw30: Yes you're right @dalefukami. It looks to me that the global default colour is the same as what is being set here (`#EDEDEF`) , therefore I have removed the `color` prop. Also, the view box for Icons is 24x24 on each icon in zUI, as a result, I have removed the `size` prop as well as `24` is essentially the default height and width. Let me know if that doesn't seem right to you. Thanks!

dalefukami (approved):
Awesome, looks good.

PR #609, opened 2023-05-26
feat: add icon to create button and improve styling (create conversation flow)

domw30:
### What does this do?
- fixes width of `create` button
- adds icon to `create` button
- import icon from `zui`
- improves styling as per design

To fix the width of the button and improve styling as per design.

Before: 

Screenshot 2023-05-26 at 12 02 32

After:

https://github.com/zer0-os/zOS/assets/39112648/c1716180-8c46-4979-9d0a-efe683c8f2f3

### How do I test this?
Click to create a new conversation. Click `Start group chat`. Enter a user to start a group chat with. Click `continue`. See `create` button.


src/components/messenger/list/styles.scss
```diff
@@ -324,9 +324,29 @@ $side-padding: 16px;
...
+      &__button-icon {
+        color: theme.$color-secondary-11;
+      }
+    }
+
+    &:hover {
+      &__button-icon {
+        color: theme.$color-black;
+      }
+    }
```
dalefukami, line 349: Is this one needed? Seems like the hover color should already come from the Button component.

dalefukami (approved):

src/components/messenger/list/styles.scss
domw30: @dalefukami, correct, with the hot-fix you made to zUI this is no longer necessary. I have removed the `button-icon` class and the styles here. As you said, the colour and hover colour comes from the Button component and works as expected now with the latest zUI version. Thank you!

PR #610, opened 2023-05-26
feat: group detail panel positioning and refactor

domw30:
### What does this do?
This PR improves the positioning of the details panel elements and refactors some of the styling that was in place, i.e. use `gap` instead of margins/padding.

So UI is in line with figma designs and improved style approach.

### How do I test this?
Go through the create conversation flow, select to create a new group, select a user to start a group with and continue to the group details panel.

**Before**
Screenshot 2023-05-26 at 13 22 08

**After**
Screenshot 2023-05-26 at 13 17 08

Screenshot 2023-05-26 at 13 17 10

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "zOS",
-  "version": "0.32.0",
+  "version": "0.33.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "zOS",
-      "version": "0.32.0",
+      "version": "0.33.0",
       "dependencies": {
         "@craco/craco": "^6.4.3",
         "@giphy/js-fetch-api": "^4.7.1",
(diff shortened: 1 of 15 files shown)


dalefukami (approved):

dalefukami:
One thing to keep in mind is when there are large amounts of data. I'm not sure if anyone tested this panel with a huge group of people so it may already be broken here but wanted to point out for these types of style changes it's usually good to test those scenarios.

dalefukami:
@domw30 you might need to double check the button size now that I merged in the zUI fix.

domw30:
> 

Hey @dalefukami ! Thanks again for the review. You're right here, I had a look and I can see that the Selected Contacts container/component doesn't really handle a large number of contacts/members selected. I checked the figma designs to see if there was any information on how this should behave but couldn't see any details. As a result, I reached out to Ethan in `zOS` channel so he is aware of this. I think it would be best to get some input from design on how to handle this scenario and create a new task for any suggested changes/updates, and then add some tests for coverage on this component. Does that sound good to you? let me know :)

dalefukami:
Yep, new task. My main concern was breaking things that already worked.

PR #613, opened 2023-05-26
Zos 363 fix connect wallet issue

ratik21:
### What does this do?

Fixes the Connect Wallet bug which neo keeps having. 

If it happens, then localStorage key is cleared, page is reloaded & we show the popup to connect wallet.

### DIFF
--- a/src/components/web3-connect/index.tsx
+++ b/src/components/web3-connect/index.tsx
@@ -156,6 +156,16 @@ export class Container extends React.Component<Properties, State> {
     if (web3.account !== previouslyAccount) {
       this.props.setAddress(web3.account);
     }
+
+    if (
+      connectionStatus === ConnectionStatus.Connecting &&
+      previousConnectionStatus === ConnectionStatus.Connected &&
+      web3.account
+    ) {
+      localStorage.removeItem('previousConnector');
+      window.location.reload();
+      this.reconnectPreviousConnector();
+    }
   }
 
   get shouldRender() {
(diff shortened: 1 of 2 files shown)


src/components/web3-connect/index.tsx
```diff
@@ -156,6 +156,16 @@ export class Container extends React.Component<Properties, State> {
     if (web3.account !== previouslyAccount) {
       this.props.setAddress(web3.account);
     }
```
ratik21, line 158: we should really refactor this code, it's pretty hard to understand. Plus i've added one more condition 😂

src/components/web3-connect/index.tsx
```diff
@@ -156,6 +156,16 @@ export class Container extends React.Component<Properties, State> {
...
+
+    if (
+      connectionStatus === ConnectionStatus.Connecting &&
+      previousConnectionStatus === ConnectionStatus.Connected &&
+      web3.account
+    ) {
+      localStorage.removeItem('previousConnector');
+      window.location.reload();
+      this.reconnectPreviousConnector();
```
dalefukami, line 167: Pretty sure if we reload we'll never get to this line?

dalefukami (approved):
Yeah, I'm kinda right in here right now so I'm hoping to make clear it up a bit.

src/components/web3-connect/index.tsx
ratik21: yes i am pretty sure. Spend considerable time on testing. Though it could happen there is a case (or comes a case in future) where we fallback to this line, but i am good with merging this.

PR #617, opened 2023-05-30
Refactor: Move invite toast management to user registration saga

dalefukami:
### What does this do?

Pushes the timing logic of when to render the invite toast up to the existing user registration saga.

Simplification of state management in the component and concentrates the user flow logic in a single place

### How do I test this?

Register a new user and verify that the invite toast is shown 10 seconds after the rewards popup is closed for the first time.

### DIFF
--- a/src/component-mocks.ts
+++ b/src/component-mocks.ts
@@ -53,8 +53,10 @@ jest.mock('@zero-tech/zui/components', () => ({
   Tabs: () => null,
   TabsNav: () => null,
   TextStack: () => null,
+  ToastNotification: () => null,
   ToggleGroup: () => null,
   BASE_CLASSNAME: () => null,
   Tooltip: () => null,
+  Video: () => null,
 }));
 export {};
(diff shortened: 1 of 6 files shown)


src/component-mocks.ts
```diff
@@ -53,8 +53,10 @@ jest.mock('@zero-tech/zui/components', () => ({
   Tabs: () => null,
   TabsNav: () => null,
   TextStack: () => null,
+  ToastNotification: () => null,
```
dalefukami, line 56: I regenerated the mock files for the zUI components and icons as I noticed we hadn't done it in a while and there were a couple of new ones. I can split this out to a separate PR if preferred.

src/store/registration/saga.test.ts
```diff
@@ -457,6 +466,28 @@ describe('channelsLoaded', () => {
   });
 });
 
+describe('openInviteToastWhenRewardsPopupClosed', () => {
+  it('sets the invite open flag after the rewards propup is closed', async () => {
```
domw30, line 470: Nice! I'll need to dive into some more of the tests here to fully understand testing the sagas but from a high level I can follow whats going on and makes sense

src/store/registration/saga.ts
```diff
@@ -230,3 +232,9 @@ function* openFirstConversation() {
     yield call(channelsLoaded);
   }
 }
+
+export function* openInviteToastWhenRewardsPopupClosed() {
+  yield take(SagaActionTypes.RewardsPopupClosed);
```
domw30, line 237: So much cleaner 🙌

PR #621, opened 2023-06-01
feat: Implement image replacement functionality

domw30:
### What does this do?
After an image has been uploaded, there is now an option to choose another. This functionality is triggered by an edit icon. The drag and drop functionality also works for replacing an image.

Adds error state to `ImageUpload`.

This PR also fixes the aspect ratio issue on uploaded images.

So a user is able to change the image they have selected. This may be for a number of reasons, i.e. selected the wrong image in the media picker.

### How do I test this?
Navigate to one of the flows that contains the ImageUpload component. Upload an image and select the IconButton to open the media picker and replace the image, or drag and drop an image file over the ImageUpload component.

https://github.com/zer0-os/zOS/assets/39112648/9399e660-7e09-4443-9585-5e06a2cc200d

Error State:
Screenshot 2023-06-01 at 13 45 33

Screenshot 2023-06-01 at 14 32 01

### DIFF
--- a/src/authentication/create-account-details/index.tsx
+++ b/src/authentication/create-account-details/index.tsx
@@ -67,6 +67,7 @@ export class CreateAccountDetails extends React.Component<Properties, State> {
               onChange={this.trackImage}
               icon={this.renderImageUploadIcon()}
               uploadText='Select or drag and drop'
+              onError={Boolean(this.props.errors.image)}
             />
           </div>
           {this.imageError && <Alert variant='error'>{this.imageError}</Alert>}
--- a/src/authentication/create-account-details/index.tsx
+++ b/src/authentication/create-account-details/index.tsx
@@ -67,7 +67,7 @@ export class CreateAccountDetails extends React.Component<Properties, State> {
               onChange={this.trackImage}
               icon={this.renderImageUploadIcon()}
               uploadText='Select or drag and drop'
-              onError={Boolean(this.props.errors.image)}
+              isError={Boolean(this.props.errors.image)}
             />
           </div>
           {this.imageError && <Alert variant='error'>{this.imageError}</Alert>}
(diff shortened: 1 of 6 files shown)


src/components/image-upload/index.tsx
```diff
@@ -11,32 +15,60 @@ export interface Properties {
...
     if (typeof this.props.onChange === 'function') {
       this.props.onChange(files[0]);
     }
   };
 
-  renderImage = () => {
+  onSimulateClick = () => {
+    if (this.fileInputRef.current) {
+      this.fileInputRef.current.click(); // simulating a click on the file input when the button is clicked
+    }
```
domw30, line 36: The `onClick` on the `inputProps` is meant to simulate a click on the file input, which opens the file dialog for selecting a file. As a result, we can add the same behaviour by using a reference to the input element and programmatically clicking it.

src/components/messenger/list/group-details-panel.tsx
```diff
@@ -56,7 +56,7 @@ export class GroupDetailsPanel extends React.Component<Properties, State> {
             <Input value={this.state.name} onChange={this.nameChanged} />
           </div>
 
-          <div>
+          <div className={c('field-info-container')}>
```
domw30, line 59: Changes required here as the `ImageUpload` component design with error message resulted in changes to the positioning of the upload component in this group details panel.

dalefukami (dismissed):

src/authentication/create-account-details/index.tsx
```diff
@@ -67,6 +67,7 @@ export class CreateAccountDetails extends React.Component<Properties, State> {
               onChange={this.trackImage}
               icon={this.renderImageUploadIcon()}
               uploadText='Select or drag and drop'
+              isError={Boolean(this.props.errors.image)}
```
dalefukami, line 70: Need the error message here?
`errorMessage={this.props.errors.image}`

src/components/account/create.tsx
```diff
@@ -248,8 +250,6 @@ export class Container extends React.Component<Properties, State> {
             {Boolean(this.state.displayNameError) && this.renderError(this.state.displayNameError)}
           </div>
 
-          {Boolean(this.state.error) && this.renderError(this.state.error)}
```
dalefukami, line 251: I don't think this state value is necessarily an image specific error so we might have to leave this component alone for now:

`this.setState({ error: 'Creating user failed' });`

dalefukami (approved):
Just a couple of quick notes.

PR #625, opened 2023-06-02
Zos 289: Show remaining invites in `invite-dialog`

ratik21:
image

image

### DIFF
--- a/src/components/invite-dialog/container.test.tsx
+++ b/src/components/invite-dialog/container.test.tsx
@@ -13,7 +13,7 @@ describe('Container', () => {
       inviteUrl: '',
       assetPath: '',
       invitesUsed: 0,
-      maxInvitesPerUser: 0,
+      maxUses: 0,
       fetchInvite: () => null,
       ...props,
     };
--- a/src/components/invite-dialog/container.test.tsx
+++ b/src/components/invite-dialog/container.test.tsx
@@ -12,6 +12,8 @@ describe('Container', () => {
       inviteCode: '',
       inviteUrl: '',
       assetPath: '',
+      invitesUsed: 0,
+      maxInvitesPerUser: 0,
       fetchInvite: () => null,
       ...props,
     };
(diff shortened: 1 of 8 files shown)


src/components/invite-dialog/container.test.tsx
```diff
@@ -12,6 +12,8 @@ describe('Container', () => {
       inviteCode: '',
       inviteUrl: '',
       assetPath: '',
+      invitesUsed: 0,
+      maxInvitesPerUser: 0,
```
dalefukami, line 16: I might rename this to `maxUses` or something slightly more generic. You never know when the intent may change.

src/components/invite-dialog/index.tsx
```diff
@@ -88,6 +95,20 @@ export class InviteDialog extends React.Component<Properties, State> {
               {this.state.copyText}
             </button>
           </div>
+          <div className={c('remaining-invite')}>
+            <IconGift1
+              className={classNames({
```
dalefukami, line 100: Rather than use this same classname twice you can put it on the parent element. Also, to be a touch more consistent with bem class naming you'd use the modifier.
```
<div className={classNames(c('remaining-invite'), {[c('remaining-invite', 'none')]: this.getInvitesRemaining() === 0})}>
```

dalefukami (approved):
Sweet. The comments are optional.

src/components/invite-dialog/index.tsx
ratik21: This worked in parent:
```js
className={
  classNames(c('remaining-invite'), {[c('no-invite-left')]: this.getInvitesRemaining() === 0})
}

But i don't fully understand how. I'll read up on `bem` & conditionally joining classnames more.
```

src/components/invite-dialog/container.test.tsx
ratik21: makes sense. done

src/components/invite-dialog/index.tsx
dalefukami: So, your version would end up with a div that has classes:

`invite-dialog__remaining-invite invite-dialog__no-invite-left`

What you want is something like:

`invite-dialog__remaining-invite invite-dialog__remaining-invite--none`

That is, it's always the element `remaining-invite` and, in this case, it's also in the state `none`.

PR #626, opened 2023-06-02
Add an action type to the user login/logout events

dalefukami:
### What does this do?

Adds a `type` attribute to the multicast user events.

Allows listeners to filter on the event type that they care about.

### How do I test this?

Login and logout in various configurations. Verify that the create conversation state is cleared, notifications are re-registered, inivites are fetched for the new user, etc.


domw30 (approved):
Looks good to me 👍

PR #628, opened 2023-06-05
Zos 297: As a user of zero messenger web app who is in a network i want to invite someone to zero messenger only so i do not add them to my current network

ratik21:
[ZOS 297](https://wilderworld.atlassian.net/browse/ZOS-297?atlOrigin=eyJpIjoiMzg4MjU0MzFhY2NiNGQyZDgyNmNmNzU3NWE4NWYyMjQiLCJwIjoiaiJ9)

IF you're in full screen mode, and a part of networks, then we show an additional alert in the `invite-dialogue`

image

### DIFF
--- a/src/components/invite-dialog/index.tsx
+++ b/src/components/invite-dialog/index.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 
-import { Image, Skeleton } from '@zero-tech/zui/components';
+import { Image, Skeleton, Alert } from '@zero-tech/zui/components';
 import { IconXClose, IconGift1 } from '@zero-tech/zui/icons';
 
 import { clipboard } from '../../lib/clipboard';
@@ -22,6 +22,7 @@ export interface Properties {
   maxUses: number;
   inviteUrl: string;
   assetsPath: string;
+  isUserInFullScreenModeAndInWorlds: boolean;
   clipboard?: Clipboard;
 
   onClose?: () => void;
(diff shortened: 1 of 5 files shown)


src/components/invite-dialog/index.tsx
```diff
@@ -77,6 +78,14 @@ export class InviteDialog extends React.Component<Properties, State> {
             alt='Hands reaching out to connect'
             className={c('image')}
           />
+
+          {this.props.isUserInFullScreenModeAndInWorlds && (
+            <Alert variant='info' className={c('network-alert')}>
+              This invite will add someone to your direct messages, <b>not</b> your current network. If you would like
```
dalefukami, line 84: Take out this second sentence. There's no way to invite someone to a network. (see the first note in the Jira story)

src/components/invite-dialog/index.tsx
```diff
@@ -22,6 +22,7 @@ export interface Properties {
   maxUses: number;
   inviteUrl: string;
   assetsPath: string;
+  isUserInFullScreenModeAndInWorlds: boolean;
```
dalefukami, line 25: I don't think checking for full screen is necessary. The invite dialog doesn't care if you're in full screen mode or not. The message only matters if you're in worlds or not.

src/components/invite-dialog/index.tsx
ratik21: done

src/components/invite-dialog/index.tsx
ratik21: okay, updating

dalefukami (approved):

PR #629, opened 2023-06-05
Zos 353: Realtime messages arent consistent

ratik21:
### What does this do?

Forces a channels + conversations refresh after reconnection with the sendbird client. 

Realtime message doesn't seem as realtime as they should be.

### DIFF
--- a/src/components/chat-connect/chat-connect.tsx
+++ b/src/components/chat-connect/chat-connect.tsx
@@ -9,6 +9,7 @@ import {
   receiveDeleteMessage as receiveDeleteMessageAction,
 } from '../../store/messages';
 import { receiveIsReconnecting } from '../../store/chat';
+import { startChannelsAndConversationsAutoRefresh } from '../../store/channels-list';
 import { RootState } from '../../store/reducer';
 import { unreadCountUpdated } from '../../store/channels';
 import { updateConnector } from '../../store/web3';
@@ -19,8 +20,8 @@ export interface Properties {
   isLoading: boolean;
   chatAccessToken: string;
   chat?: Chat;
-  reconnectStart: () => void;
-  reconnectStop: () => void;
+  receiveIsReconnecting: (payload: boolean) => void;
+  startChannelsAndConversationsAutoRefresh: () => void;
   receiveNewMessage: (channelId: string, message: Message) => void;
   receiveDeleteMessage: (channelId: string, messageId: number) => void;
   receiveUnreadCount: (channelId: string, unreadCount: number) => void;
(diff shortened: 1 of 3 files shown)


ratik21:
Before (please skip through the waiting/watch in 2x):

https://drive.google.com/file/d/1vJR8lRrDtmvTe-1-mQfEkO79Rj7q1VEg/view?usp=sharing

After (forcing channel refresh just after reconnecting)

https://drive.google.com/file/d/16q-m-kuDqnDZUrTvFdMA80wEU_lfhLUg/view?usp=sharing

src/components/chat-connect/chat-connect.tsx
```diff
@@ -75,6 +76,17 @@ export class Container extends React.Component<Properties> {
     this.chat = props.chat || chat;
   }
 
+  reconnectStart = () => {
```
dalefukami, line 79: I think we should move these all up into sagas. I'm not sure why we have a component managing these external connections. But, that's a problem for another day. :)

src/lib/chat/index.ts
```diff
@@ -27,7 +27,8 @@ export class Chat {
       modules: [new GroupChannelModule()],
     }) as SendbirdGroupChat;
 
-    this.sendbird.options.sessionTokenRefreshTimeout = 1800; // Max
+    this.sendbird.options.sessionTokenRefreshTimeout = 1800; // 30min Max
+    this.sendbird.options.websocketResponseTimeout = 1800; // 30min Max
```
dalefukami, line 31: I'm guessing we probably don't want this.

src/lib/chat/index.ts
```diff
@@ -45,6 +46,14 @@ export class Chat {
     this.initSessionHandler(events);
     this.initConnectionHandlers(events);
     this.initChannelHandlers(events);
+
+    // every 5s check if the connection state is CLOSED, if it is, set the app to foreground,
+    // to prevent sendbird sdk from disconnecting (when the app is in the background)
+    setInterval(() => {
```
dalefukami, line 52: Fine for now but we should move this into the sagas. It should be rare for a component to have to schedule things to happen like this. We could also maybe only do it when the tab is in the background?

dalefukami (approved):

src/lib/chat/index.ts
ratik21: removed

src/lib/chat/index.ts
```diff
@@ -27,7 +27,8 @@ export class Chat {
       modules: [new GroupChannelModule()],
     }) as SendbirdGroupChat;
 
-    this.sendbird.options.sessionTokenRefreshTimeout = 1800; // Max
+    this.sendbird.options.sessionTokenRefreshTimeout = 1800; // 30min Max
```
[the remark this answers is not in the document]
ratik21: increased to 2 hours

PR #630, opened 2023-06-05
feat: start a direct message conversation with existing zero users by searching their name

domw30:
### What does this do?
When typing into the search conversations input, a new section has been added that will display users that match the string that is entered. 

The idea is that a user doesn't have to send an invite link to another user to search, find and start a direct conversation with.

### How do I test this?
Type characters into the search conversation input and click a result that appears. This should open the direct message window that enables you to send a message to that user.

https://github.com/zer0-os/zOS/assets/39112648/f052582a-599c-4192-b9ad-99b55c076adb

#### Things to consider:
  1. Does this change any APIs?


src/store/create-conversation/saga.ts
```diff
@@ -102,6 +103,13 @@ const PREVIOUS_STAGES = {
   [Stage.GroupDetails]: Stage.StartGroupChat,
 };
 
+function* handleCreateConversation() {
```
domw30, line 106: Added this `handleCreateConversation` function to prevent changing to much around the setting of the `Stages`. Previously, when `this.props.stage === SagaStage.None` the action `createConversation/create` would not dispatch and therefore the direct message chat would not pop up in the UI. Not sure if this is the best approach, but appears to work well with little change to support code.

src/components/messenger/list/conversation-list-panel.test.tsx
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+import { act } from '@testing-library/react';
```
dalefukami, line 2: Don't think we need this to simulate the event you're simulating.

src/components/messenger/list/conversation-list-panel.tsx
```diff
@@ -9,24 +9,40 @@ import { IconButton } from '../../icon-button';
...
+  searchChanged = async (search: string) => {
     this.setState({ filter: search });
+
+    if (!search) {
+      return this.setState({ userSearchResults: null });
+    }
+
+    const items: Item[] = await this.props.search(search);
+    const conversationMemberIds = this.props.conversations.flatMap((c) => c.otherMembers.map((m) => m.profileId));
+    const filteredItems = items?.filter((item) => !conversationMemberIds.includes(item.id));
```
dalefukami, line 43: :thinking: I think there might be a usability question here. If I'm in a group conversation with you and someone else.... and then I try to start a one on one conversation with only you....you won't show up in the results, right? I think we want to show existing convos with a match but only filter out the one on one based ones from the api search results?

src/components/messenger/list/conversation-list-panel.tsx
```diff
@@ -9,24 +9,40 @@ import { IconButton } from '../../icon-button';
...
+  userSearchResults: Option[];
 }
 
 export class ConversationListPanel extends React.Component<Properties, State> {
-  state = { filter: '', inviteDialogOpen: false };
+  state = { filter: '', inviteDialogOpen: false, userSearchResults: [] };
 
-  searchChanged = (search: string) => {
+  searchChanged = async (search: string) => {
     this.setState({ filter: search });
```
dalefukami, line 35: Curious, does this cause a rerender before we hit the `await` for the api search? would be nice to quickly filter the existing convos so that you're not stuck waiting for a slow api request.

src/store/create-conversation/saga.ts
dalefukami: I think we need to tweak this a little. My concern is that we'll get into a state where 2 things are listening for this event at the same time. I _think_ one will eat it and the other will be left in an inconsistent state.

Might be worth pairing for a bit here but, at a quick glance, what you probably want is that in the "normal" case we're listening for either a createconversation event OR a `SagaActionTypes.Start` event. It'd be a `race` call because you want whichever comes first. If you get the createconvo event then do the create convo...if you get a start event then drop into the startConversation flow.

src/components/messenger/list/conversation-list-panel.tsx
domw30: When I create a group and then search one of those users that I do not have a current conversation with they show up in the search results, I think thats what you mean ? 😬 

https://github.com/zer0-os/zOS/assets/39112648/21509fe9-0001-40e0-98e6-1d505d749352

src/store/create-conversation/saga.ts
domw30: I looked into this a little and followed your suggestion to use a `race` call. I updated the `saga` function to listen for either a `SagaActionTypes.Start` or `SagaActionTypes.CreateConversation` event, so whichever event comes first, race should cancel the other. If `SagaActionTypes.Start` comes first, the saga will call `startConversation`; if `SagaActionTypes.CreateConversation` comes first, it will call `createConversation` with the event payload, as suggested! Let me know what you think 💪  Thanks for the guidance above!

PR #632, opened 2023-06-05
Remove web3 registration feature flag

dalefukami:
### What does this do?

Removes the web3 registration feature flag now that the feature is complete.


colbr (approved):
Looks good

PR #633, opened 2023-06-06
feat: web3 login routing

colbr:
### What does this do?

- Created a `LoginPage` component for the login page
- Made `Web3Login` and `EmailLogin` components which "plug into" the main `Login` page component
- Switch between `Web3Login` and `EmailLogin` based on selected option in `login` saga
- Some refactoring to styles as I moved

To finish off #627 

### How do I test this?

1. Open the login page
2. Click `Web3 Wallet`
3. Check that wallet selector renders
4. Click `Email`
5. Check that email inputs render

#### Things to consider:
  1. Does this change any APIs?


colbr:
I spent a while debugging a weird error, where if you swap to `Web3 Wallet`, back to `Email`, then back to `Web3 Wallet`, the wallet selector doesn't render. I'm still not really sure what's going on there, but I'll investigate.

src/authentication/email-login/styles.scss
```diff
@@ -1,26 +1,5 @@
...
-    @include form();
-  }
-
-  &__other-options {
-    text-align: center;
-    margin-top: 32px;
-    padding-top: 32px;
-    border-top: 1px solid theme.$color-greyscale-transparency-7;
-  }
+  @include form();
```
colbr, line 4: This should be refactored out later - didn't think it was worth doing right now

src/pages/login/login-container.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import { LoginComponent } from './login-component';
+
+export interface LoginContainerProperties {
+  isLoggingIn: boolean;
+  stage: LoginStage;
+  switchLoginStage: (stage: LoginStage) => void;
+}
+
+export class LoginContainer extends React.Component<LoginContainerProperties> {
```
colbr, line 15: Wasn't sure if I handled this correctly - let me know if there's something wrong in here. Something didn't feel quite right about it, but it may be because I'm just not used to the pattern.

src/store/login/index.ts
```diff
@@ -37,6 +39,7 @@ export const initialState: LoginState = {
 
 export const loginByEmail = createAction<{ email: string; password: string }>(SagaActionTypes.EmailLogin);
 export const loginByWeb3 = createAction<Connectors>(SagaActionTypes.Web3Login);
+export const switchLoginStage = createAction<LoginStage>(SagaActionTypes.SwitchLoginStage);
```
colbr, line 42: Wasn't sure if we'd want to limit the params to `Web3 | Email`, so it's not possible to switch to `Done`, but unsure.

PR #636, opened 2023-06-06
Mentions don't render properly in the conversation list

ratik21:
Before

Conversations

After

image


dalefukami (approved):
Killer. Thanks for doing that.

If you're ambitious... we have a similar problem when there's an image as the last message. :)

PR #637, opened 2023-06-07
feat: highlight matching characters when searching users and conversations

domw30:
### What does this do?
When using the search function in the conversations panel, matching characters are highlighted in purple.

As per designs.

### How do I test this?
Type into the search conversation/user input. If any characters/words match the text in the results should be highlighted purple.

#### Things to consider:
  1. Does this change any APIs?


src/store/create-conversation/saga.ts
```diff
@@ -56,6 +56,7 @@ export function* createConversation(action) {
 
 export function* saga() {
   yield fork(authWatcher);
+  yield fork(handleCreateConversation);
```
dalefukami, line 59: I'm guessing a rebase or merge from main will get rid of this but just wanted to point it out to double check.

dalefukami (approved):
:boom:

src/store/create-conversation/saga.ts
domw30: Yeah there were a few conflicts to fix up. All sorted :) Shall merge this tomorrow.

PR #645, opened 2023-06-09
Remove Network bar from zOS + messenger

ratik21:
[JIRA](https://wilderworld.atlassian.net/browse/ZOS-388?atlOrigin=eyJpIjoiMWYzNTBlMDA1NDA4NDI4Zjk0NWYyYWYxMjk4ZTgyNTEiLCJwIjoiaiJ9)

zOS
image

Messenger
image

### DIFF
--- a/src/Main.test.tsx
+++ b/src/Main.test.tsx
@@ -29,14 +29,6 @@ describe('Main', () => {
     expect(wrapper).toHaveElement(WalletManager);
   });
 
-  /*
-  it('renders view mode toggle', () => {
-    const wrapper = subject();
-
-    expect(wrapper).toHaveElement(ViewModeToggle);
-  });
-  */
-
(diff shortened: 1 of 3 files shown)


domw30 (approved):
Looks good to me. Maybe lets check with Dale if we're happy to just comment the code out rather than remove 👍

src/Main.test.tsx
```diff
@@ -30,11 +29,13 @@ describe('Main', () => {
     expect(wrapper).toHaveElement(WalletManager);
   });
 
+  /*
```
dalefukami, line 32: Yeah, just delete this. We always have git history.

dalefukami (approved):
Yep, delete is generally better than comment. It could be a few months before we resurrect this again and we can either dig through git history or just rewrite it from scratch.

PR #650, opened 2023-06-12
Don't stop polling if error occured

ratik21:
For now, if an error occures during polling, i am just returning an empty state. This is how it'll look

image

But now it'll continue to poll.

### DIFF
--- a/src/store/channels-list/api.ts
+++ b/src/store/channels-list/api.ts
@@ -4,13 +4,23 @@ import { get, post } from '../../lib/api/rest';
 import { DirectMessage } from './types';
 
 export async function fetchChannels(id: string) {
-  const channels = await get<any>(`/api/networks/${id}/chatChannels`);
-  return await channels.body;
+  try {
+    const channels = await get<any>(`/api/networks/${id}/chatChannels`);
+    return await channels.body;
+  } catch (error: any) {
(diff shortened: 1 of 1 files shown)


src/store/channels-list/api.ts
```diff
@@ -4,13 +4,23 @@ import { get, post } from '../../lib/api/rest';
...
 
 export async function fetchChannels(id: string) {
-  const channels = await get<any>(`/api/networks/${id}/chatChannels`);
-  return await channels.body;
+  try {
+    const channels = await get<any>(`/api/networks/${id}/chatChannels`);
+    return await channels.body;
+  } catch (error: any) {
+    console.log('Error occured while fetching chatChannels ', error?.response?.body?.error); // eg. error.code = ENOTFOUND
+    return [];
```
ratik21, line 12: Probably need some design states? Eg. "Authorization required", "Internet not found" ..etc. In order to properly display those errors to the user

dalefukami: I think in most cases the user doesn't even know we're polling so it's probably not a huge deal for now. Maybe eventually we need to distinguish between a first time fetch and the background poll.

dalefukami (approved):
Thanks!

src/store/channels-list/api.ts
ratik21: ok got it. i'll record the foreground/background polling in Jira

PR #651, opened 2023-06-12
Display most recent conversation in main container as default conversation on login

ratik21:
[JIRA](https://wilderworld.atlassian.net/browse/ZOS-387?atlOrigin=eyJpIjoiMjNjZWJiMjAzMTM5NDRkM2E2ZDUwNjMyNWVlZTA0MjIiLCJwIjoiaiJ9)

https://github.com/zer0-os/zOS/assets/33264364/c8666168-edd9-4729-a3f4-cdb33bc00983

### DIFF
--- a/src/store/login/saga.ts
+++ b/src/store/login/saga.ts
@@ -1,4 +1,5 @@
-import { call, put, race, spawn, take, takeEvery, takeLatest } from 'redux-saga/effects';
+import getDeepProperty from 'lodash.get';
+import { call, put, race, select, spawn, take, takeEvery, takeLatest } from 'redux-saga/effects';
 
 import { emailLogin as apiEmailLogin } from './api';
 import { EmailLoginErrors, LoginStage, SagaActionTypes, Web3LoginErrors, setErrors, setLoading, setStage } from '.';
@@ -7,7 +8,9 @@ import { logout, nonceOrAuthorize, terminate } from '../authentication/saga';
 import { setWalletModalOpen } from '../web3';
 import { Events as AuthEvents, getAuthChannel } from '../authentication/channels';
 import { Web3Events, getWeb3Channel } from '../web3/channels';
-import { openFirstConversationAfterChannelsLoaded } from '../registration/saga';
+import { conversationsChannel } from '../channels-list/channels';
+import { rawConversationsList } from '../channels-list/saga';
+import { setactiveConversationId } from '../chat';
 
 export function* emailLogin(action) {
   const { email, password } = action.payload;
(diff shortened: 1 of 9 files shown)


src/components/messenger/list/index.tsx
```diff
@@ -138,6 +141,12 @@ export class Container extends React.Component<Properties, State> {
     this.props.fetchRewards({});
   }
 
+  componentDidUpdate(): void {
+    if (this.props.isMessengerFullScreen && !this.props.activeMessengerId) {
+      this.props.openConversation(this.props.conversations[0]?.id);
+    }
+  }
+
```
dalefukami, line 149: We should manage this in a saga somewhere. Pretty sure there's already a saga for opening fullscreen mode?

dalefukami (changes requested):

dalefukami:
BTW, I think this resolves this story: https://wilderworld.atlassian.net/browse/ZOS-391 instead of the one you noted...though they may have the same solution.

ratik21:
> BTW, I think this resolves this story: https://wilderworld.atlassian.net/browse/ZOS-391 instead of the one you noted...though they may have the same solution.

Should i pull this in as well, and add a note in the comment that this story is handled by this PR?

src/components/messenger/list/index.tsx
ratik21: Handled both flows (login + expand) in saga.

dalefukami:
> Should i pull this in as well, and add a note in the comment that this story is handled by this PR?

Yep, that makes sense

src/store/login/saga.ts
```diff
@@ -148,4 +149,7 @@ export function* saga() {
     const action = yield take(SagaActionTypes.EmailLogin);
     success = yield call(emailLogin, action);
   } while (!success);
+
+  // After successful login
+  yield spawn(openFirstConversationAfterChannelsLoaded);
```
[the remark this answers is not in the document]
ratik21: updated to `spawn` after the `UserLogin` event. 

> Then, I'm guessing, we can probably remove the one from the registration flow since the event fires in both cases.

Little confused about this one. I see that we're firing the event in both cases (registration, login). But aren't these separate saga flows? i.e you're either registering (going through account setup) & then firing the event, or you're logging in, and then firing the event. Isn't that separate "events"? Or is there an overlap here?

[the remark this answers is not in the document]
dalefukami: You're correct. We fire them via both flows. From an front end perspective what is a registration? It's effectively creating an account and then immediately logging in. The entire UI doesn't care if you got into the app through the login pages or the registration pages. All it cares about is that a new user is authenticated and we know who they are now. That's the UserLogin event. It can also actually happen from public zOS when you click the Connect button.

src/store/registration/saga.ts
```diff
@@ -222,22 +223,24 @@ export function* saga() {
   yield updateProfilePage();
 
   // After successful registration
-  yield spawn(openFirstConversation);
+  yield spawn(openFirstConversationAfterChannelsLoaded);
```
dalefukami, line 226: Should be able to remove this now since it's not specific to the registration flow any more, right?

dalefukami (approved):

src/store/login/saga.ts
```diff
@@ -132,6 +133,9 @@ function* listenForLoginEvents() {
   while (true) {
     yield take(authChannel, AuthEvents.UserLogin);
 
+    // After successful login
+    yield spawn(openFirstConversationAfterChannelsLoaded);
```
dalefukami, line 137: Don't have to move this now but I'd consider which piece of the app this impacts. Maybe a listener should exist in the area where the activeConversationId is maintained. That would isolate this logic from the login logic more cleanly. If that area of the app ever gets refactored/removed then the login code would never need to be touched.

Again, don't have to change this now. Just trying to point out alternative design choices.

src/store/registration/saga.ts
ratik21: done, i also moved the function implementation from registration to login, since it is where it's being used now.

dalefukami: Yeah, makes sense!

PR #657, opened 2023-06-13
Rename `activeMessengerId` to `activeConversationId`

ratik21:


dalefukami (approved):
It's a little concerning how frequently we access this variable. lol

PR #659, opened 2023-06-13
Move the flow of logging in fully into sagas

dalefukami:
### What does this do?

Moves the login, page load, change account flows fully into sagas.

Takes away a lot of the logic that was happening based on component property changes. Now things are happening based on actual user flows or in reaction to user/authentication events.

### How do I test this?

As an email user and web3 user:
* Register for a new account
* Login from scratch
* Reload the page when logged in
* Reload the page when not logged in

In all cases verify that chats are loaded, the page is in the right opening state, etc.


src/store/authentication/api.ts
```diff
@@ -53,3 +53,21 @@ export async function fetchChatAccessToken(): Promise<{ chatAccessToken: string
 
   return body;
 }
+
+export async function emailLogin({ email, password }: { email: string; password: string }) {
```
dalefukami, line 57: This just moved from the `login` api file to here.

src/store/authentication/saga.ts
```diff
@@ -28,90 +30,75 @@ export const currentUserSelector = () => (state) => {
...
   }
 
   return { nonce };
 }
 
-export function* terminate() {
-  yield processUserAccount({ user: null, nonce: null, chatAccessToken: null, isLoading: false });
+export function* completeUserLogin(user = null) {
+  if (!user) {
+    user = yield call(fetchCurrentUser);
```
ratik21, line 52: what happens if the `fetchCurrentUser` fails? Do we fail here, or return `null` & continue?

src/store/authentication/saga.ts
```diff
@@ -28,90 +30,75 @@ export const currentUserSelector = () => (state) => {
...
+  }
+  yield put(setUser({ data: user }));
+  yield call(initializeUserState, user);
+  yield call(publishUserLogin, user);
+}
+
+export function* terminate(isChangeAccount = false) {
```
ratik21, line 59: nit: i would prefer `isAccountChanged`

src/store/channels-list/saga.ts
```diff
@@ -147,7 +148,26 @@ export function* channelsReceived(action) {
...
+  const userChannel = yield call(getAuthChannel);
+  while (true) {
+    yield take(userChannel, Events.UserLogout);
+    yield put({ type: SagaActionTypes.StopChannelsAndConversationsAutoRefresh });
+  }
+}
+
 export function* saga() {
+  yield spawn(listenForUserLogin);
+  yield spawn(listenForUserLogout);
```
ratik21, line 170: one doubt here, so this will get "spwaned" & keep listening "forever" in the background? I.e if i login, 169 get's triggered, but if i logout, and then again login, 169 gets fired again, right? (Given that we spawn it a single time from looking at the code)

src/store/login/saga.ts
```diff
@@ -162,14 +170,18 @@ export function* openFirstConversationAfterChannelsLoaded() {
   }
 }
 
+function* listenForLogoutEvents() {
```
ratik21, line 173: Can we rename this to `listenForUserLogout` (same as used previously).

Also one question, i see that we've implemented this listener in 2 places. Just to confirm this relates to the "single responsibility principle" that you were talking about, right? Otherwise we could have just added a single listener, and if user logout is received, just do a) `stopChannelRefresh(..)`, b) `reset()` together ..?

src/store/page-load/saga.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  if (featureFlags.allowPublicZOS) {
+    yield initializePublicLayout();
+    return;
+  }
+
+  history.replace({
+    pathname: '/login',
+  });
+}
```
ratik21, line 30: I actually don't understand this file. We're not listening for a saga event here. Where does this code gets triggered/used? 🤔

src/platform-apps/channels/container.tsx
```diff
@@ -57,13 +57,19 @@ export class Container extends React.Component<Properties> {
   componentDidMount() {
     const { domainId } = this.props;
 
-    this.props.fetchChannels(domainId);
+    if (domainId) {
+      this.props.fetchChannels(domainId);
+    }
```
ratik21, line 62: what happens if `domainId` is `undefined`? Do we see an empty state?

ratik21 (approved):
mostly questions

src/store/channels-list/saga.ts
dalefukami: That's correct.

src/store/login/saga.ts
dalefukami: That's right. This one is here because it's managing the state for the Login flow. The `reset` here is based on the context of this set of sagas and store data.

src/store/page-load/saga.ts
dalefukami: This code gets called immediately when the global `store/saga.ts` is loaded. It's the first file in the list of sagas that get run.

src/store/authentication/saga.ts
ratik21: yeah, agree. A global error state for handling an error while authentication makes sense to me

src/store/login/saga.ts
ratik21: got it, thanks

PR #660, opened 2023-06-14
feat: implement scrollbar containers and styles

domw30:
### What does this do?
- adds `scrollbar()` mixin (rather than use from zos-component-library)
- adds a new `ScrollbarContainer` that handles the style and behaviour of the scrollbar.
- adds a faded panel to to the `ScrollbarContainer` and its behaviour.
- adds the `ScrollbarContainer` to the conversation list panel message items.
- adds the `Scrollbar` container to the channel list content.
- adds `scrollbar()` mixin styles for the invert scroll on the main container.
- adds tests.

As per designs.

### How do I test this?
Open the app and start scrolling on the channel list, the main channel message container and the conversations panel 🙂 

#### Things to consider:
  1. Does this change any APIs?

https://github.com/zer0-os/zOS/assets/39112648/64fd399a-7862-4cc5-8215-38d792c5b6ac

### DIFF
--- a/src/_scrollbar.scss
+++ b/src/_scrollbar.scss
@@ -1,4 +1,4 @@
-@mixin scrollbar($width: 4px, $color: rgba(225, 225, 254, 0.15), $radius: 5px, $opacity: 0.19) {
+@mixin scrollbar($width: 4px, $height: 292px, $color: rgba(225, 225, 254, 0.15), $radius: 5px, $opacity: 0.19) {
   &::-webkit-scrollbar {
     width: $width;
   }
@@ -6,5 +6,6 @@
   &::-webkit-scrollbar-thumb {
     background: $color;
     border-radius: $radius;
+    height: $height;
   }
 }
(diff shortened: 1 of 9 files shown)


domw30:
Note: The `channel-view` (main central container) uses an inverted scroll container. As a result, I have added the `scrollbar()` mixin to global `index.scss`. We could possibly add inverted scroll behaviour to the `ScrollbarContainer`, but as this is slightly out of scope for this task I have pushed forward with the simplest solution for now.

src/_scrollbar.scss
```diff
@@ -0,0 +1,10 @@
+@mixin scrollbar($width: 4px, $color: rgba(225, 225, 254, 0.15), $radius: 5px, $opacity: 0.19) {
```
dalefukami, line 1: There must be an existing scrollbar style somewhere that we should remove?

dalefukami (approved):

src/index.scss
```diff
@@ -1,7 +1,7 @@
 @use 'modules/font' as fonts;
 @use 'modules/animation' as animation;
 @use 'shared-components/theme-engine/theme' as theme;
-@import '~@zer0-os/zos-component-library/dist/assets/scroll';
```
domw30, line 4: @dalefukami - there was a mixin being imported here with the previous styles

dalefukami: aha, thanks!

PR #668, opened 2023-06-15
Zos 485: Invite data reload when opening the dialog

ratik21:
### What does this do?

Reloads the invite data each time you click on "invite friends" button.

Earlier this only fetched a single time.

### DIFF
--- a/src/store/create-invitation/saga.ts
+++ b/src/store/create-invitation/saga.ts
@@ -1,4 +1,4 @@
-import { put, call, takeLatest } from 'redux-saga/effects';
+import { put, call, takeLeading } from 'redux-saga/effects';
 import { SagaActionTypes, reset, setInvite } from '.';
 import { getInvite } from './api';
 import { config } from '../../config';
@@ -26,5 +26,5 @@ export function* fetchInvite() {
 }
 
 export function* saga() {
-  yield takeLatest(SagaActionTypes.GetCode, fetchInvite);
+  yield takeLeading(SagaActionTypes.GetCode, fetchInvite);
 }
--- a/src/store/create-invitation/saga.ts
+++ b/src/store/create-invitation/saga.ts
@@ -1,56 +1,31 @@
-import { take, put, call, race, spawn, takeEvery } from 'redux-saga/effects';
+import { put, call, takeLatest } from 'redux-saga/effects';
 import { SagaActionTypes, reset, setInvite } from '.';
 import { getInvite } from './api';
 import { config } from '../../config';
-import { Events, getAuthChannel } from '../authentication/channels';
 
 export function* fetchInvite() {
-  while (true) {
-    const { cancel } = yield race({
-      cancel: take(SagaActionTypes.Cancel),
-      get: take(SagaActionTypes.GetCode),
-    });
+  // reset before fetching
+  yield put(reset());
 
-    if (cancel) {
-      return yield put(reset());
-    }
+  try {
+    const invitation = yield call(getInvite);
+    console.log('invitation', invitation);
 
-    try {
-      const invitation = yield call(getInvite);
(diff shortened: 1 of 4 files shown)


src/store/create-invitation/saga.ts
```diff
@@ -33,16 +33,24 @@ export function* fetchInvite() {
...
+  const authChannel = yield call(getAuthChannel);
+  while (true) {
+    yield take(authChannel, Events.UserLogout);
+    yield put({ type: SagaActionTypes.Cancel });
   }
 }
+
+export function* saga() {
+  yield spawn(listenForUserLogin);
+  yield spawn(listenForUserLogout);
```
ratik21, line 54: Thought this would be cleaner. 

But I had 2 doubts

a) Are there any events besides `UserLogin` & `UserLogout` in the `authChannel`. Seems like previously we were just looking for any event "except" the login before cancel 🤔 

b) Do we even need to "listen" now? Since we're fetching everytime we're opening the invite dialogue.

src/store/create-invitation/saga.ts
```diff
@@ -33,16 +33,24 @@ export function* fetchInvite() {
...
+  yield takeEvery(SagaActionTypes.GetCode, fetchInvite);
```
ratik21, line 55: @dalefukami  I wanted to use `takeLatest` here (i.e take the Latest one, and ignore an already running saga action). BUT that doesn't work. I don't seem to understand fully why. The `fetchInvite` gets called yes, but it gets stuck on this line 

https://github.com/zer0-os/zOS/blob/aa448e2603e6683ea9ebedfb01ebffdfa9a91ed6/src/store/create-invitation/saga.ts#L19

i.e I don't receive any value of `invitation` (BUT the api gets called :/). 

And everything works as normal while using `takeEvery`..

src/store/create-invitation/saga.ts
dalefukami: Yeah, at this point, you can just eliminate the auth event listeners. There were only 2 and this was written before we discovered you could filter based on the `type` so that's why it was listening for "*".

What we _could_ do is get rid of the custom Cancel event we have here and swap in the UserLogout event. Effectively line 10 was saying, "try to get the invite but a user logs out in the middle bail out." (based on the listener for user logout). But, I'm not sure it matters any more if we're going to fetch the data every time. If an invite fetch is still running when the user logs out... it doesn't really matter because the request has already been started. Maybe we just need to `reset` the data every time just before we start sending the api request.

src/store/create-invitation/saga.ts
dalefukami: So, this is probably because there's a conflict here with `takeLatest` and the `while(true)` (line 8) in  the previous implementation. You're now listening for the `GetCode` even twice. Once in the `takeEvery` and then again at the start of the while loop. Since we can switch to the simpler takeEvery we no longer need the while loop.

src/store/create-invitation/saga.ts
```diff
@@ -1,48 +1,30 @@
...
-  while (true) {
-    const { userId = undefined } = yield take(channel, '*');
-
-    if (userId) {
-      yield spawn(fetchInvite);
-    } else {
-      yield put({ type: SagaActionTypes.Cancel });
-    }
-  }
+  yield takeLatest(SagaActionTypes.GetCode, fetchInvite);
```
dalefukami, line 29: Only thing we might change is that this could be `takeLeading`? The chances that we have 2 invite requests happen quickly enough and they're for different users is extremely low(?) and there's no point cancelling an in flight request.

dalefukami (approved):

src/store/create-invitation/saga.ts
ratik21: done.

PR #674, opened 2023-06-19
fix: rewards modal design and copy issues

domw30:
### What does this do?
Right now there are a number of stylistic issues with the Rewards modal, both in terms of elements and copy.

1. Alters the modal to utilize the “x” button icon specified in the designs.
2. Alter the accordion design component to match design specifications (done by updating in zUI and pulling in new version to zOS).
3. Update the position modal, one accordion open per time.
4. Updates the FAQ copy

As per design.

### How do I test this?
When viewing the rewards pop up, select `Learn more about zero rewards`. The reward faq modal will appear and compare this against the figma design [here](https://www.figma.com/file/aETpyuG2AKjNcQtCjldcX0/ZERO-Messenger-%2F-Web?type=design&node-id=7563-326198&t=tFqXY6i77gQzc6G1-4)

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "@zero-tech/zapp-daos": "latest",
     "@zero-tech/zapp-nfts": "latest",
     "@zero-tech/zapp-staking": "latest",
-    "@zero-tech/zui": "^0.23.0",
+    "@zero-tech/zui": "^0.24.1",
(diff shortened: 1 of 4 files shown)


dalefukami (approved):
A note for the future, next time it'd be good to split these up in to separate PRs. FAQ changes, updating zui, etc.

PR #679, opened 2023-06-19
Move chat connection and event handling to sagas

dalefukami:
### What does this do?

Moves the Sendbird chat connection and event handling fully to sagas rather than in components.

Makes event handling more clear and doesn't rely on compnonent re-renders to keep global state up to date.

### How do I test this?

Test the various real-time pieces of chatting:
1. Receiving a new message in an open conversation updates the conversation in real time
2. Deleting a message in an open conversation updates the conversation in real time
3. Receiving a new message in a not open conversation updates the unread count in the conversation list

### DIFF
--- a/src/store/chat/bus.ts
+++ b/src/store/chat/bus.ts
@@ -8,6 +8,7 @@ export enum Events {
   UnreadCountChanged = 'chat/message/unreadCountChanged',
   ReconnectStart = 'chat/recconectStart',
   ReconnectStop = 'chat/recconectStop',
+  InvalidToken = 'chat/invalidToken',
 }
 
 let theBus;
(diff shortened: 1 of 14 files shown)


src/store/chat/bus.ts
```diff
@@ -0,0 +1,49 @@
...
+  MessageReceived = 'chat/message/received',
+  MessageDeleted = 'chat/message/deleted',
+  UnreadCountChanged = 'chat/message/unreadCountChanged',
+  ReconnectStart = 'chat/recconectStart',
+  ReconnectStop = 'chat/recconectStop',
+  InvalidToken = 'chat/invalidToken',
+}
+
+let theBus;
+export function* getChatBus() {
```
dalefukami, line 15: After some discussion with Ratik we thought we'd migrate the names for redux-saga channels to "bus" to avoid confusion with our domain "channels". Tried it here and it seems to feel ok.

PR #680, opened 2023-06-19
Clear the interval when disconnecting sendbird

dalefukami:
### What does this do?

Clears our 'keep-alive' interval in the sendbird class and adds a little null safety.

So things don't explode when you logout

### How do I test this?

Login...then logout and confirm the page doesn't explode with an error.


ratik21 (approved):
Thanks!

PR #684, opened 2023-06-19
Dont show rewards tooltip if rewards are '0'

ratik21:
Prevents this state

image

### DIFF
--- a/src/components/rewards-bar/index.tsx
+++ b/src/components/rewards-bar/index.tsx
@@ -78,6 +78,15 @@ export class RewardsBar extends React.Component<Properties, State> {
   closeRewardsFAQModal = () => this.setState({ isRewardsFAQModalOpen: false });
   closeRewardsTooltip = () => this.setState({ isRewardsTooltipOpen: false });
 
+  showNewRewardsToUser = () => {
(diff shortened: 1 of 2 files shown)


src/components/rewards-bar/index.tsx
```diff
@@ -78,6 +78,15 @@ export class RewardsBar extends React.Component<Properties, State> {
   closeRewardsFAQModal = () => this.setState({ isRewardsFAQModalOpen: false });
   closeRewardsTooltip = () => this.setState({ isRewardsTooltipOpen: false });
 
+  showNewRewardsToUser = () => {
+    return (
+      this.props.isMessengerFullScreen &&
+      !this.props.isFirstTimeLogin &&
+      this.isNewRewardsLoaded() &&
+      this.stringifyZero(this.props.zeroPreviousDay) !== '0'
```
ratik21, line 86: I am thinking about moving this to saga

dalefukami: Yep, good idea!

dalefukami (approved):
If this works then we can merge it now and you can do the move to sagas as a follow up.

PR #688, opened 2023-06-20
Move some of the rewards logic to sagas

ratik21:


src/store/rewards/saga.ts
```diff
@@ -29,6 +31,24 @@ export function* syncFetchRewards() {
...
+}
+
+export function* rewardsPopupClosed() {
+  // set last viewed rewards to the current rewards when the popup is closed
+  const { zeroPreviousDay, showNewRewards } = yield select((state) => state.rewards);
+  if (showNewRewards) {
+    localStorage.setItem(lastViewedRewardsKey, zeroPreviousDay);
+    yield put(setShowNewRewards(false));
+  }
+}
```
ratik21, line 49: will add tests for these if the logic looks good

src/store/login/saga.ts
```diff
@@ -17,8 +17,9 @@ import { setWalletModalOpen } from '../web3';
 import { Events as AuthEvents, getAuthChannel } from '../authentication/channels';
 import { Web3Events, getWeb3Channel } from '../web3/channels';
 import { conversationsChannel } from '../channels-list/channels';
-import { rawConversationsList } from '../channels-list/saga';
+import { delay, rawConversationsList } from '../channels-list/saga';
```
dalefukami, line 20: I'm not sure why we have our own `delay`. redux-saga has one already I think.

src/store/login/saga.ts
```diff
@@ -167,6 +168,9 @@ export function* openFirstConversationAfterChannelsLoaded() {
   const isMessengerFullScreen = yield select((state) => getDeepProperty(state, 'layout.value.isMessengerFullScreen'));
   if (payload.loaded && isMessengerFullScreen) {
     yield call(openFirstConversation);
+
+    yield delay(2000);
+    yield call(checkNewRewardsLoaded);
```
dalefukami, line 173: Hmm.. is there a reason to delay this? Don't we want to just render this right away if they have new rewards when they log in? As in, is there even a reason to wait for the conversations to load?

src/store/rewards/saga.ts
```diff
@@ -29,6 +31,24 @@ export function* syncFetchRewards() {
...
+export function* checkNewRewardsLoaded() {
+  const zeroPreviousDay = yield select((state) => state.rewards.zeroPreviousDay);
+  const isFirstTimeLogin = yield select((state) => state.registration.isFirstTimeLogin);
+  if (!isFirstTimeLogin && zeroPreviousDay !== '0' && localStorage.getItem(lastViewedRewardsKey) !== zeroPreviousDay) {
+    yield put(setShowNewRewards(true));
+  }
+}
+
+export function* rewardsPopupClosed() {
+  // set last viewed rewards to the current rewards when the popup is closed
```
dalefukami, line 43: What if they close the tooltip but not the rewards popup? We don't always show the rewards popup, that's only on registration, I think?

I think we should maybe set the last viewed at the same time we set the `showNewRewards`?

src/store/rewards/saga.ts
```diff
@@ -29,6 +31,24 @@ export function* syncFetchRewards() {
   }
 }
 
+export function* checkNewRewardsLoaded() {
```
dalefukami, line 34: Does this still happen when we poll for new rewards? Seems like this is wired up only for when you finish logging in but I wasn't quite sure. Maybe I'm crazy. Didn't we set it up to poll every hour?

src/store/login/saga.ts
ratik21: no reason, i just thought it would look a little nicer on the UI. i.e the first conversation loads up (takes a second or two since it's a sendbird call), and then popup the rewards tooltip. 

https://github.com/zer0-os/zOS/assets/33264364/27c1d80f-9b02-4b17-9de7-8295e9531b3e

src/store/rewards/saga.ts
ratik21: We record that the user has viewed the rewards (by using `localStorage.last_viewed_rewards`) when it "views" the reward popup, and not the reward tooltip 😅 

Because you always have this status (shown below) that the user hasn't viewed the latest rewards
image

After the user opens the popup & closes it, then we record that the user has viewed these (this is how it was originally implemented, it is just moved to the sagas here)

src/store/rewards/saga.ts
ratik21: oh you're right. We should add this check in the after we're fetching the rewards each time i think. I have removed the delay, and added the `yield call(checkNewRewardsLoaded)` in the `fetch` (rewards).

ratik21: thanks for spotting.

src/store/rewards/saga.ts
dalefukami: ahhh, right... interesting... those seem like different things to me maybe we'll eventually split them. Eg: if I don't want to have to keep closing the tooltip and it's not obvious that opening the rewards would do that for me.

 For now, you're right, this is correct!

dalefukami (approved):

PR #690, opened 2023-06-21
feat: add settings menu

domw30:
### What does this do?
Triggers a drop down menu per the below-provided specs when a user clicks on their pfp in the upper left corner of the conversation list sidebar.

Populate a drop down menu with two items:
- a header with the user’s pfp, display name, and wallet / zID (as applicable, see design)
- logout option

* when clicking to logout the user should be redirected to login page.

Visibility for user to see which account they are logged into and ability to logout/disconnect.

### How do I test this?
Open messenger. Click profile icon (avatar in top left corner in the sidekick. Dropdown menu should appear with the two items detailed above. Check logout functionality by clicking logout. Should redirect you to the login page.

#### Things to consider:
  1. Does this change any APIs?

https://github.com/zer0-os/zOS/assets/39112648/b5cd84f2-a4ce-408d-909c-36af46f51997

### DIFF
--- a/src/components/messenger/list/index.tsx
+++ b/src/components/messenger/list/index.tsx
@@ -15,8 +15,8 @@ import {
   membersSelected,
   startCreateConversation,
 } from '../../../store/create-conversation';
+import { logout } from '../../../store/authentication';
 import { CreateMessengerConversation } from '../../../store/channels-list/types';
-
 import { IconExpand1, IconXClose } from '@zero-tech/zui/icons';
 
 import './styles.scss';
@@ -68,6 +68,7 @@ export interface Properties extends PublicProperties {
   enterFullScreenMessenger: () => void;
   fetchRewards: (_obj: any) => void;
   rewardsPopupClosed: () => void;
+  logout: () => void;
 }
 
 interface State {
@@ -129,6 +130,7 @@ export class Container extends React.Component<Properties, State> {
       fetchRewards,
       enterFullScreenMessenger: () => enterFullScreenMessenger(),
       rewardsPopupClosed,
+      logout,
     };
   }
 
@@ -224,6 +226,7 @@ export class Container extends React.Component<Properties, State> {
           includeRewardsAvatar={this.props.includeRewardsAvatar}
           userAvatarUrl={this.props.userAvatarUrl}
           onRewardsPopupClose={this.props.rewardsPopupClosed}
+          onLogout={this.props.logout}
         />
 
         <div className='direct-message-members'>
--- a/src/components/messenger/list/index.tsx
+++ b/src/components/messenger/list/index.tsx
@@ -49,6 +49,8 @@ export interface Properties extends PublicProperties {
   allowClose: boolean;
   allowExpand: boolean;
   includeRewardsAvatar: boolean;
+  userName: string;
+  userHandle: string;
   userAvatarUrl: string;
   zero: string;
   zeroPreviousDay: string;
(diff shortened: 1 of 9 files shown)


src/index.scss
```diff
@@ -12,6 +12,7 @@ html {
 }
 
 body {
+  height: 100vh;
```
domw30, line 15: @dalefukami - this fixes the hidden dropdown. Brett and I had a session and looked into it some more and appears Radix is forcing an `overflow: hidden !important` to the body when the dropdown menu is open (likely to prevent scrolling when open). To resolve, adding `100vh` to the body here ensures the dropdown menu is visible once opened. I wasn't able to see any unwanted consequences of this, but let me know if there is causes problems elsewhere.

dalefukami: oh man, that sucks. Don't like it when a 3rd party is forcing styles on elements it doesn't own. :grimacing:

src/components/messenger/list/index.tsx
```diff
@@ -222,8 +228,11 @@ export class Container extends React.Component<Properties, State> {
           isMessengerFullScreen={this.props.isMessengerFullScreen}
           isFirstTimeLogin={this.props.isFirstTimeLogin}
           includeRewardsAvatar={this.props.includeRewardsAvatar}
+          userName={this.props.userName}
+          userHandle={this.props.userHandle}
```
dalefukami, line 232: I'm not entirely sure this is the right field to use. It's mostly coincidence that these align with email and web3 address at the moment (particularly for new users) but a lot of existing users would have different things here and we might actually use this in the future. I think it's ok to use this right now but we should fetch the real data before calling this done.... unless the intent is to actually use the user's handle which may change to be user managed.

dalefukami (approved):

src/components/rewards-bar/index.test.tsx
```diff
@@ -94,11 +97,11 @@ describe('rewards-bar', () => {
...
 
-  it('renders the rewards avatar as necessary', function () {
+  it('renders the SettingsMenu as necessary', function () {
     let wrapper = subject({ includeRewardsAvatar: true });
-    expect(wrapper).toHaveElement('Avatar');
+    expect(wrapper).toHaveElement('SettingsMenu');
 
     wrapper.setProps({ includeRewardsAvatar: false });
-    expect(wrapper).not.toHaveElement('Avatar');
+    expect(wrapper).not.toHaveElement('SettingsMenu');
```
ratik21, line 105: maybe also add a test for when you click on the settings menu, you should see the avatar with their display name & the logout button?

src/components/settings-menu/index.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  it('calls onLogout prop when logout item is selected', function () {
+    const mockOnLogout = jest.fn();
+    const wrapper = subject({ onLogout: mockOnLogout });
+    const dropdownMenu = wrapper.find(DropdownMenu);
+
+    dropdownMenu.prop('items')[2].onSelect();
+
+    expect(mockOnLogout).toHaveBeenCalled();
+  });
```
ratik21, line 33: ah okay you got those tests here. nvm :)

src/components/settings-menu/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+export interface Properties {
+  userName: string;
+  userHandle: string;
+  userAvatarUrl: string;
+
+  onLogout: () => void;
+}
+
+export class SettingsMenu extends React.Component<Properties> {
+  containsAtSymbol = this.props.userHandle.includes('@');
```
ratik21, line 19: we don't need to like declare these (eg. `const containsAtSymbol`? ) 😮

ratik21 (approved):
lgtm

src/components/settings-menu/index.tsx
domw30: I think this is just how class components work. I was surprised to lol

dalefukami: This is probably a subtle bug, actually. We should turn this into a function. I believe (haven't tested) that if the properties were to change that we'd now have only the original value upon construction.

domw30: Yep had a quick look into it and that's correct. When initialising class fields like above, they get values once at the time of class instantiation and won't get updated when props change. 

I'll create a PR to transform containsAtSymbol and userHandle into methods.

PR #691, opened 2023-06-22
Zos 513 UI allow users to send text pdf files

ratik21:
[JIRA](https://wilderworld.atlassian.net/browse/ZOS-464?atlOrigin=eyJpIjoiZGNmNzQzM2IxZTRlNDUxYzhiMjJjODgwMjdkYTRhNmEiLCJwIjoiaiJ9)

Allows the user to attach pdf & text files. Includes some small refactorings as well.

https://github.com/zer0-os/zOS/assets/33264364/fc6fdd62-5830-4d37-912b-ce5d822ce9cc

### DIFF
--- a/src/lib/api/attachment.ts
+++ b/src/lib/api/attachment.ts
@@ -1,6 +1,6 @@
 import { get } from './rest';
 
-interface AttachmentResponse {
+export interface AttachmentResponse {
   signedUrl: string;
   key: string;
 }
@@ -22,13 +22,7 @@ export async function getAttachmentUrl(attachment: { key: string }): Promise<str
 
 export function download(key: string): void {
   (async () => {
-    // This works, it's using the `https://res.cloudinary.url/..` to display the file
-    // (eg. this is how it looks like: https://res.cloudinary.com/fact0ry-dev/image/upload/v1687385590/ax9wzrns7msfezo0ziqu.pdf)
-    // but zero-web is uploading files to S3 first, and this api (getAttachmentUrl) was
-    // created to get the file from S3, should we do that here as well?
-    const url = key;
-
-    //const url = await getAttachmentUrl({ key });
+    const url = await getAttachmentUrl({ key });
     if (url) {
       const win = window.open('about:blank', '_blank');
       win.location.assign(url);
--- a/src/lib/api/attachment.ts
+++ b/src/lib/api/attachment.ts
@@ -7,9 +7,11 @@ interface AttachmentResponse {
 
 export async function getAttachmentUrl(attachment: { key: string }): Promise<string> {
   const filter: any = { key: attachment.key };
-  const attachmentResponse = await get<AttachmentResponse>('/api/feedItems/getAttachmentDownloadInfo', filter).catch(
-    (err) => console.error(err)
-  );
+  const attachmentResponse = await get<AttachmentResponse>(
+    '/api/feedItems/getAttachmentDownloadInfo',
+    undefined,
+    filter
+  ).catch((err) => console.error(err));
 
   if (attachmentResponse && attachmentResponse.ok) {
     return attachmentResponse.body.signedUrl;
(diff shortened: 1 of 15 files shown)


src/lib/api/attachment.ts
```diff
@@ -20,8 +22,13 @@ export async function getAttachmentUrl(attachment: { key: string }): Promise<str
 
 export function download(key: string): void {
   (async () => {
-    const url = await getAttachmentUrl({ key });
+    // This works, it's using the `https://res.cloudinary.url/..` to display the file
+    // (eg. this is how it looks like: https://res.cloudinary.com/fact0ry-dev/image/upload/v1687385590/ax9wzrns7msfezo0ziqu.pdf)
+    // but zero-web is uploading files to S3 first, and this api (getAttachmentUrl) was
+    // created to get the file from S3, should we do that here as well?
```
dalefukami, line 28: hmm... Maybe Joel will have some context here. I think we probably should use S3 for non-image attachments? or maybe all attachments? Cloudinary is more for web served images.

src/platform-apps/channels/attachment-cards/index.tsx
```diff
@@ -31,6 +38,7 @@ export default class AttachmentCards extends React.Component<Properties, undefin
             <AttachmentCard
               key={attachment.url}
               attachment={attachment}
+              onRemove={this.props.onRemove ? this.itemRemoved.bind(this, attachment) : null}
```
dalefukami, line 41: We send the attachment into `AttachmentCard`, right? It should publish `onRemove` with that value so we don't have to bind things here.

src/store/messages/saga.ts
```diff
@@ -294,6 +294,7 @@ export function* stopSyncChannels(action) {
   );
 }
 
+// this also "recieves" the message which I had just "sent". is that ideal or should we handle it?
```
dalefukami, line 297: Is this causing problems? i.e., are we double rendering or something? As long as we're properly just updating the right one then this is probably fine.

dalefukami (commented):
Just the one main question. I think we need to use S3 for the different types of attachments.

src/store/messages/saga.ts
ratik21: i had just noticed it and put in a comment, currently it seems to behave properly.

src/lib/api/attachment.ts
ratik21: I checked a little more, and in zOS we're sending the "native file" directly to this route https://github.com/m3m3n70/zero-api/blob/7c96d9646ad1e0b779017821eccdf6a43ff3af27/src/routes/upload.ts#L85

This is uploading the file first to cloudinary, and then doing a `sendFileMessage` via sendbird. 

I think if we're going to refactor this, then i think we upload to s3 first, and then use the `signedUrl` directly & send to this route - https://github.com/m3m3n70/zero-api/blob/development/src/routes/chatChannels.ts#L66 i believe?

src/platform-apps/channels/attachment-cards/index.tsx
ratik21: I tried without bind (`onRemove={this.props.onRemove ? this.itemRemoved : null}`) but it doesn't work 🤔

src/lib/api/attachment.ts
jtulloch: i haven't looked at what we're doing in zos. but in zero, when you add a file to a message, we figure out what type of file it is. if it's an image or video, we upload to cloudinary. if it's any other type of file, it becomes an 'attachment', which gets uploaded to S3.

jtulloch: this is where the type is determined:
https://github.com/m3m3n70/zero-web/blob/development/src/components/message-input/index.tsx#L189

this is where we upload non-media "attachments":
https://github.com/m3m3n70/zero-web/blob/development/src/app/api/file/index.ts#L98

ratik21: updated to upload to s3 first, and then call the api.

src/platform-apps/channels/attachment-cards/index.tsx
dalefukami: Sorry, I wasn't clear in my first comment. What I mean is in the `AttachmentCard` component it should have an `onRemove` that calls the passed in handler with the actual thing:

`AttachmentCard`:
```
  onRemove = () => {
    this.props.onRemove(this.props.attachment);
  }
```

Now this spot no longer needs to bind because it'll just receive the attachment automatically from the AttachmentCard.

ratik21: done

dalefukami (approved):

PR #697, opened 2023-06-23
Increase size of group conversation participant avatars next to message bubble in conversation window

ratik21:
image

image

image


domw30 (approved):
Nice. We should refactor these to use the Avatar component from zUI, but I think that should be a task when we have bandwidth to do so.

colbr (approved):

ratik21:
@domw30 will keep in mind

PR #698, opened 2023-06-23
Revamp view control header display

ratik21:
+ removes top title for network users in full screen mode as well
+ add collapse icon in view control header

image

Note: `IconExpand1` will be replaced by `IconCollapse1` after adding it in zUI.
image


domw30 (approved):
Looking good !

PR #704, opened 2023-06-26
Show rewards tooltip only in Messenger (full screen) mode

ratik21:
Sorry 👀 

Rewards


domw30 (approved):
Do we want to add a test for this?

Also, assuming that because we are only showing this in messenger full screen mode, we don't need to worry about the UI issues here? In your screenshot I can see the 'X' icon and the orange status/notification are a little off in terms of position/spacing.

Other than that, looks good to me!

domw30:
@ratik21, could you check the UI for this while here please. Just want to ensure it's as per design. 

Screenshot 2023-06-26 at 10 11 51

ratik21:
@domw30 yes will add a test. I think in full screen mode it should render the UI properly

PR #708, opened 2023-06-27
feat(conversations): change header

colbr:
### What does this do?

- Gets rid of the `Conversations` header
- Moves the `New Conversation` button to the right of the search bar 

Screenshot 2023-06-28 at 4 02 47 PM

Design changed slightly

### How do I test this?

1. Open the app
2. Open messenger
3. Check that the conversation search UI matches Figma

#### Things to consider:
  1. Does this change any APIs?


src/components/messenger/list/styles.scss
```diff
@@ -30,12 +30,32 @@ $side-padding: 16px;
...
+      }
+
+      &-conversations-search {
+        flex: 1;
+
+        // @note: this is here because parent size calculations with flex 1
+        // are a little funky with zUI input right now.
+        input {
+          width: 0;
+        }
```
colbr, line 51: This was a weird one - couldn't exactly work out what was going on here. Will need to dig into `zUI` to work it out, which I'll do another day.

PR #709, opened 2023-06-27
fix: collapse icon ui fixes

domw30:
Before:

Screenshot 2023-06-27 at 09 54 14

After: 

Screenshot 2023-06-27 at 10 04 25

### DIFF
--- a/src/components/messenger/chat/styles.scss
+++ b/src/components/messenger/chat/styles.scss
@@ -157,6 +157,14 @@ $recent-indicator-size: 8px;
     color: theme.$color-greyscale-11;
   }
 
+  &__icon-collapse {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    width: 32px;
+    height: 32px;
+  }
+
   &--transition {
     transition: all animation.$animation-duration-half ease-out;
 
--- a/src/components/messenger/chat/index.tsx
+++ b/src/components/messenger/chat/index.tsx
@@ -179,7 +179,12 @@ export class Container extends React.Component<Properties, State> {
 
             {this.props.allowCollapse && (
               <div>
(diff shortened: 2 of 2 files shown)


src/components/messenger/chat/styles.scss
```diff
@@ -157,6 +157,14 @@ $recent-indicator-size: 8px;
     color: theme.$color-greyscale-11;
   }
 
+  &__icon-collapse {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    width: 32px;
+    height: 32px;
```
ratik21, line 165: i little confused here. We pass `size{..}` into the `IconButton` in zUI and we're also setting the height & width here? What's the difference b/w the two?

domw30:
```
  render() {
    return (
      <button className={classNames('icon-button', this.props.className)} onClick={this.handleClick}>
        <this.props.Icon label={this.props.label} size={this.props.size} isFilled={this.props.isFilled} />
      </button>
    );
  }
  ```
  
You can see in `IconButton` here there is the button container wrapped around the icon. We need to pass this a className and set the height/width relevant to the size passed to the Icon itself. So for 24px size passed to the icon, we want to set the container (button) at height and width of 32px so the background/on hover aligns. You can see this in the figma design here:

The icon size 24px:
Screenshot 2023-06-27 at 10 24 14

The button container 32px as stated bottom right of screenshot.
Screenshot 2023-06-27 at 10 24 25

ratik21: ahhhhh, so the container is 32px, but the "icon size" is actually 24px. I didn't knew that before, thought you just need to take the icon size

ratik21 (approved):

PR #714, opened 2023-06-27
Reduce size of classname specification in nodes

dalefukami:
### What does this do?

Updates a component to use a more succinct way of setting the classname on nodes.

This is a proposal PR. I'm experimenting with ways to enhance the scannability of components.

This is an attempt to make components simpler to scan. It may seem like a small change but I find the result reduces clutter and allows for quicker scanning for important information:

```
<div {...cn('title')}>Start a new conversation:</div>
```
vs
```
<div classname={c('title')}>Start a new conversation:</div>
```


domw30 (approved):
I would prefer the explicit `className` attribute, but happy to go either way.

PR #719, opened 2023-06-28
feat(rewards-popup): number formatting

colbr:
### What does this do?

Formats the reward amount in the rewards modal as described in [Jira](https://wilderworld.atlassian.net/browse/ZOS-430?atlOrigin=eyJpIjoiMTlkZTM0OTMxNWM5NDA3YzkwZmE1MTNkODY3Yzc3ZTkiLCJwIjoiaiJ9).

The rewards amount did not match Figma.

### How do I test this?

- Open the rewards modal
- Check that the number is formatting as per the description in Jira.

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,6 @@
     "@zero-tech/zui": "^0.25.1",
     "audio-react-recorder-fixed": "^1.0.3",
     "classnames": "^2.3.1",
-    "decimal.js": "^10.4.3",
     "emoji-mart": "^3.0.1",
     "es6-promise-debounce": "^1.0.1",
     "ethers": "^5.5.2",
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "lodash.get": "^4.4.2",
     "lodash.kebabcase": "^4.1.1",
     "lodash.uniqby": "^4.7.0",
+    "millify": "^6.1.0",
     "moment": "^2.29.4",
     "normalizr": "^3.6.2",
     "pusher-js": "^8.0.2",
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "@zero-tech/zui": "^0.25.1",
     "audio-react-recorder-fixed": "^1.0.3",
     "classnames": "^2.3.1",
+    "decimal.js": "^10.4.3",
     "emoji-mart": "^3.0.1",
     "es6-promise-debounce": "^1.0.1",
     "ethers": "^5.5.2",
--- a/src/components/rewards-popup/index.test.tsx
+++ b/src/components/rewards-popup/index.test.tsx
@@ -18,8 +18,9 @@ describe('RewardsPopup', () => {
     return shallow(<RewardsPopup {...allProps} />);
   };
 
-  // @note 29-06-2023
-  //
+  /* @note 29-06-2023
+   * Removed this test as component isn't rigged up to render USD price
+   */
   // it('renders your rewards count in USD', function () {
   //   const wrapper = subject({ usd: '$5.71' });
   //
--- a/src/components/rewards-popup/index.test.tsx
+++ b/src/components/rewards-popup/index.test.tsx
@@ -18,12 +18,14 @@ describe('RewardsPopup', () => {
     return shallow(<RewardsPopup {...allProps} />);
   };
 
-  it('renders your rewards count in USD', function () {
-    const wrapper = subject({ usd: '$5.71' });
-
-    const skeleton = wrapper.find('.rewards-popup__rewards-usd SkeletonText');
-    expect((skeleton.prop('asyncText') as any).text).toEqual('$5.71');
-  });
+  // @note 29-06-2023
+  //
+  // it('renders your rewards count in USD', function () {
+  //   const wrapper = subject({ usd: '$5.71' });
+  //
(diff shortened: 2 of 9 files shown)


src/lib/number.ts
```diff
@@ -0,0 +1,14 @@
+import { millify } from 'millify';
+
+export function formatNumber(num: number | string): string {
```
jtulloch, line 3: i have a few comments about this function. rather than comment on each line, i'll just throw them into one comment. haha.
- 'formatNumber' feels overly generic for something that's actually very specific. all we're actually using this for is formatting a token value, i think?
- i think we should be able to assume string input here.
- and. .. let's not cast to float. let's just keep things as a string.

src/lib/number.test.ts
```diff
@@ -0,0 +1,19 @@
...
+  it('should format numbers less than 1 million with at most two decimal places', () => {
+    expect(formatNumber(12345.694206942)).toEqual('12,345.69');
+    expect(formatNumber('12345.6942069420')).toEqual('12,345.69');
+    expect(formatNumber('12345.6')).toEqual('12,345.6');
+  });
+
+  it('should format numbers greater than or equal to 1 million with suffix', () => {
+    expect(formatNumber(12345678)).toEqual('12.35M');
+    expect(formatNumber('12345678')).toEqual('12.35M');
+  });
```
jtulloch, line 18: is this something we actually want? i didn't see the designs, but this feels inconsistent with how we're representing token amounts.

colbr: The following was in the Jira issue:

> Truncate long fractional numbers – likely to arise one users can redeem rewards – and full integer values by rounding up and incorporating English words, i.e.,  (show 2 decimals) 12345.6942069420 → 12,345.70 or 12,345.69, 12345678901112 → 12,345,678,901,112 or 12.35 Trillion (bonus if simple)

I interpreted this as if we're wanting to millify large numbers, but could be wrong here!

colbr:
Thanks for the review @jtulloch I'll fix it up!

src/lib/number.ts
colbr: Hmm damn my comment disappeared, but I agree with these and have fixed them ☺️

package.json
```diff
@@ -36,6 +36,7 @@
     "@zero-tech/zui": "^0.24.4",
     "audio-react-recorder-fixed": "^1.0.3",
     "classnames": "^2.3.1",
+    "decimal.js": "^10.4.3",
```
colbr, line 39: Wasn't sure if it was heavy handed to import a decimal library here, but I figured it's a tiny library and we should really use it wherever we have a wei amount (YAGNI I know)

src/lib/number.test.ts
colbr: Just found [this Figma frame](https://www.figma.com/file/aETpyuG2AKjNcQtCjldcX0/ZERO-Messenger-%2F-Web?type=design&node-id=9703%3A268151&mode=design&t=iBCZXrRXxzczRa8n-1), which says we should millify 6 figures and above in the rewards popup

jtulloch: 🤷 well there you go.

package.json
jtulloch: the only thing is, i just don't think we should ever need to use decimals. decimals should only ever be appearing in the final display string. everything else should just be integers represented as strings, or if we're doing calculations, BigNumbers. even for USD, or other tokens with minimal decimal places, we should always just be using integers.

domw30 (approved):
🚀

src/lib/number.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  it('should commify numbers less than 1 million', () => {
+    expect(formatWeiAmount(get('12345'))).toEqual('12,345');
+  });
+
+  it('should round numbers less than 1 million to 2 DP of precision', () => {
+    expect(formatWeiAmount(get('12345.694206942'))).toEqual('12,345.69');
+    expect(formatWeiAmount(get('12345.6'))).toEqual('12,345.6');
+  });
+
+  it('should format numbers greater than or equal to 100,000 with suffix M', () => {
```
ratik21, line 16:
```suggestion
  it('should format numbers greater than or equal to 100,000 with suffix K', () => {
```

ratik21 (approved):

src/components/rewards-popup/index.test.tsx
```diff
@@ -18,12 +18,15 @@ describe('RewardsPopup', () => {
     return shallow(<RewardsPopup {...allProps} />);
   };
 
-  it('renders your rewards count in USD', function () {
-    const wrapper = subject({ usd: '$5.71' });
-
-    const skeleton = wrapper.find('.rewards-popup__rewards-usd SkeletonText');
-    expect((skeleton.prop('asyncText') as any).text).toEqual('$5.71');
-  });
+  /* @note 29-06-2023
```
jtulloch, line 21: let's just actually delete this test rather than commenting out, and then 🚢

jtulloch (changes requested):

src/components/rewards-popup/index.test.tsx
ratik21: done

PR #723, opened 2023-06-28
Fade in the messages once fetched

dalefukami:
### What does this do?

Fades in the messages once they're fetched to ease the feel a bit.

Based on feedback from the design team we decided this smooths out the transition between chats and between the loading state to completed state.

### DIFF
--- a/src/platform-apps/channels/styles.scss
+++ b/src/platform-apps/channels/styles.scss
@@ -216,6 +216,8 @@ $scrollbar-width: 5px;
 }
 
 .messages__container {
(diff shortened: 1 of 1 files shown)


src/platform-apps/channels/styles.scss
```diff
@@ -216,6 +216,8 @@ $scrollbar-width: 5px;
 }
 
 .messages__container {
+  animation: fadein 200ms ease-out forwards;
```
ratik21, line 219: ok i know how this works now 😅

ratik21 (approved):

PR #724, opened 2023-06-28
fix: support text & pdf 

ratik21:


src/store/messages/api.ts
```diff
@@ -72,13 +73,22 @@ export async function uploadAttachment(file: File): Promise<AttachmentUploadResu
   });
   const uploadInfo: AttachmentResponse = response.body;
 
-  // note: for adding progress bar to the upload
-  // https://github.com/m3m3n70/zero-web/blob/development/src/app/api/file/index.ts#L105
+  console.log('uploadInfo', uploadInfo);
```
dalefukami, line 76: Debugging?

ratik21: hell 🤦‍♂️ i had removed from local, forgot to push

PR #725, opened 2023-06-28
Fixes issue when the 'more' menu doesn't show up on messages prior to a date stamp

dalefukami:
### What does this do?

Fixes an issue where messages just prior to the date stamp in a conversation wouldn't react to mouse events thus preventing the 'more' menu from showing up.

Before:
before

After:
after

### DIFF
--- a/src/platform-apps/channels/styles.scss
+++ b/src/platform-apps/channels/styles.scss
@@ -221,6 +221,7 @@ $scrollbar-width: 5px;
     overflow: inherit;
     transition: opacity 0.9s ease-out;
     opacity: 1;
+    clear: both;
 
     .message__header {
       display: flex;


src/platform-apps/channels/styles.scss
```diff
@@ -221,6 +221,7 @@ $scrollbar-width: 5px;
     overflow: inherit;
     transition: opacity 0.9s ease-out;
     opacity: 1;
+    clear: both;
```
ratik21, line 224: what does this do (from a css perspective) 👁️ ?

dalefukami: The issue is related to using "float: left" and "float: right" for the messages. When you "float" something it takes it out of the normal layout flow so the thing below it will actually render over top of the thing you floated. The `clear` means roughly, "take into account floated things when positioning me".  That's a really bad simplification but probably good enough for now. lol

ratik21: haha ok, i got the context. thanks!

PR #728, opened 2023-06-29
feat(rewards-popup): match close button to Figma

colbr:
### What does this do?

`Close` button
- Correct button size
- Correct button color

Match UI to Figma

### How do I test this?

- Open Rewards popup
- [Compare `Close` button to Figma](https://www.figma.com/file/aETpyuG2AKjNcQtCjldcX0/ZERO-Messenger-%2F-Web?type=design&node-id=7019%3A313888&mode=design&t=iBCZXrRXxzczRa8n-1)

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/rewards-popup/styles.scss
+++ b/src/components/rewards-popup/styles.scss
@@ -45,8 +45,14 @@
 
   &__close-button {
     position: absolute;
-    top: 16px;
-    right: 16px;
+    top: 18px;
+    right: 18px;
+
+    svg {
+      path {
+        color: theme.$color-greyscale-12;
+      }
+    }
   }
 
   &__heading {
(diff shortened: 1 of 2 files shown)


src/components/rewards-popup/styles.scss
```diff
@@ -45,8 +45,14 @@
 
   &__close-button {
     position: absolute;
-    top: 16px;
-    right: 16px;
+    top: 18px;
+    right: 18px;
```
colbr, line 49: `18px` might look funky from a code-standpoint, but the `IconButton` padding is slightly different in zUI and zDS, so I had to accomodate for this.

domw30: Yeah I think IconButton needs some updating. I often have to put a container around it to fix up the size correctly as per figma design. Normally in figma the svgs are 24px with a container of 32px, but setting `size={32}` on `IconButton` sets the svg to 32px not the container....if that makes sense.

domw30 (approved):
👍

PR #734, opened 2023-06-30
feat: update message menu dropdown

domw30:
### What does this do?
- replaces portal menu with existing zUI DropdownMenu component.
- replaces the existing tests with tests to work with DropdownMenu component.
- handles styles and positioning of the open menu.
- ensures that only one dropdown menu is open per time.
- fixes the position of the open dropdown menu when mouse is no longer hovering over the message row.
- moves edit message actions to relevant folder

- as per figma designs and required refactoring.

### How do I test this?
- open the messenger and hover over a message to view the three dot menu. Click the three dot menu.

#### Things to consider:
  1. Does this change any APIs?

https://github.com/zer0-os/zOS/assets/39112648/138e7cd7-7778-49bb-baaa-4fddb2957e48


src/platform-apps/channels/messages-menu/styles.scss
```diff
@@ -1,127 +1,47 @@
...
-      text-transform: uppercase;
-      cursor: pointer;
-      margin-left: 30px;
-      background: none;
-      border: none;
-      outline: none;
-      font-weight: 600;
-    }
-  }
+.dropdown-menu__underlay {
```
domw30, line 25: I'm using this underlay to handle the ability to have only one dropdown menu open per time. This also prevents any other elements that have hover styles from being triggered.

src/platform-apps/channels/messages-menu/index.test.tsx
```diff
@@ -1,96 +1,111 @@
...
+    });
   });
 
-  it('it should renders edit button when icon menu is clicked', () => {
-    const onEdit = jest.fn();
-    const wrapper = subject({ onEdit, canEdit: true });
+  describe('Edit Button', () => {
+    it('should render when canEdit is true and isMediaMessage is false', () => {
+      const onEdit = jest.fn();
+      const wrapper = subject({ canEdit: true, onEdit }) as ShallowWrapper;
```
dalefukami, line 63: I'm surprised you had to force this to a type. It should know that from the return value of the function. :thinking:

src/platform-apps/channels/messages-menu/styles.scss
dalefukami: Seems like something that should come with the dropdown menu automatically? Where else would we want multiple dropdown menus to be rendered at the same time?

Not worth changing right now but we should keep this in mind.

dalefukami (approved):

PR #736, opened 2023-07-02
(sentry) update `mark-as-read` api to be called from saga instead of `componentDidUpdate()`

ratik21:


domw30 (approved):
Looks good to me. Lets get @dalefukami's approval too

src/store/messages/saga.ts
```diff
@@ -111,6 +112,10 @@ export function* fetch(action) {
       hasLoadedMessages: true,
     })
   );
+
+  // Publish a system message across the channel
+  const channel = yield call(conversationsChannel);
+  yield put(channel, { channelId });
```
dalefukami, line 118: Let's add a type here. I know we still have a few of these events that don't specify a type but we should try to make the new ones use it. It allows us to filter properly in the listeners (rather than listening for '*')

src/store/channels/saga.ts
```diff
@@ -62,9 +63,20 @@ export function* clearChannels() {
   yield put(removeAll({ schema: schema.key }));
 }
 
+function* listenForChannelLoadedEvent() {
+  const channel = yield call(conversationsChannel);
+  while (true) {
+    const { channelId = undefined } = yield take(channel, '*');
```
dalefukami, line 69: I'm trying to decide if this is the right event. If we start pre-loading messages for some reason, for example, then we'd be marking the messages as read all incorrectly for some things. Maybe we just need to make this more strict and only react if the channel matches the open conversation / channel?

dalefukami (commented):
:heart: the move to sagas!!

src/store/messages/saga.ts
ratik21: makes sense, updated. Will keep in mind for the new ones.

src/store/channels/saga.ts
dalefukami: Also just had another thought, we also need to mark as read when new real time messages come in and the conversation is open.

ratik21: I know about the `activeConversationId`, but i am not sure if we have an "active channel id"..? Or can we just extract/know that from the url maybe? eg

image

dalefukami: Yeah, I'm not sure, we'd have to dig around.  But somehow we must know because we load the messages for the channel and receive real time messages, etc. Maybe it mostly currently works just based on the fact that it's open? We should probably have that in state somewhere.

ratik21: So this is getting the channelId from the current url i think, and passing it via props in the UI

https://github.com/zer0-os/zOS/blob/a51af1a8074d6efe3cdc68ad5d657066e5282a55/src/platform-apps/channels/index.tsx#L19

> But somehow we must know because we load the messages for the channel and receive real time messages, etc

correct.

ratik21: I will update to read the "current channel id" from the url. And also trigger `mark-as-read` on receiveNewMessage sendbird event

dalefukami: In this case I think maybe we need to have the component tell us when the channel changes. All the routing is done via React so it makes sense that it is the original source of that state. I.E., when the channel changes in the main component (wherever that is) it can call a "setCurrentChannel" or something that will just update the global state.

ratik21: I have updated the PR as per your comments. But the code got a little more complex that i would like 🤔

src/components/chat-view-container/chat-view-container.test.tsx
```diff
@@ -25,6 +25,7 @@ describe('ChannelViewContainer', () => {
       markAllMessagesAsReadInChannel: () => undefined,
```
dalefukami, line 25: can we remove this now?

src/store/messages/saga.ts
```diff
@@ -362,6 +367,7 @@ export function* receiveNewMessage(action) {
       })
     ),
     call(sendBrowserNotification, channelId, message),
+    call(markAllMessagesAsRead, channelId, message.sender?.userId),
```
dalefukami, line 370: I don't think this is right. We only want to mark the messages read if it's the conversation/channel that's currently open.

src/components/chat-view-container/chat-view-container.tsx
```diff
@@ -96,22 +97,27 @@ export class Container extends React.Component<Properties, State> {
   componentDidMount() {
     const { channelId } = this.props;
     if (channelId) {
+      // only set the active channel ID if it's not a direct message conversation
+      if (!this.props.activeConversationId) {
```
dalefukami, line 101: Wouldn't this be set if you had both channels and the chat window open? So, if you have the chat window open we wouldn't change the activeChannelId correctly?

Also, do we clear the activeChannelId somewhere when we navigate away from the channels route?

I think maybe managing the activeChannelId could be it's own PR so we can figure out the details there first.

src/store/channels/saga.ts
```diff
@@ -23,11 +24,10 @@ export function* joinChannel(action) {
...
 
-/**
- * Marks all messages as "read" in a channel for a specific user (queries zero-api & sendbird).
- */
-export function* markAllMessagesAsReadInChannel(action) {
-  const { channelId, userId } = action.payload;
+export function* markAllMessagesAsRead(channelId, userId) {
+  if (!channelId || !userId) {
+    return;
+  } // in case of an admin message userId can be undefined
```
dalefukami, line 30: Isn't the user id supposed to be the currently logged in user? We're saying, "this user has read these messages". So, admin messages shouldn't have any impact here.

We probably still need this in the case of public viewers because they won't have a userId but I'm worried we're passing the wrong user id here maybe?

src/store/channels/saga.ts
```diff
@@ -62,9 +90,19 @@ export function* clearChannels() {
   yield put(removeAll({ schema: schema.key }));
 }
 
+function* listenForChannelLoadedEvent() {
+  const channel = yield call(conversationsChannel);
+  while (true) {
+    const { channelId = undefined } = yield take(channel, ChannelEvents.MessagesLoadedForChannel);
+    if (channelId) {
+      yield call(markAllAsRead);
```
dalefukami, line 98: I think maybe we should have 2 events. One for Channels and one for Conversations. I know they got merged together but we should treat them as totally different things.

Also, we shouldn't mark everything read like this. We should only mark the conversations as read IF this event was for the already open channel. The bug here is, what if we decided to start preloading messages for channels that we're not viewing at the moment? We'd get a bunch of these events but we wouldn't want to just "mark the current conversation" unread.

src/store/messages/saga.ts
dalefukami: Ah, this is also where we're using the wrong userId. Maybe just don't send the userId and let the other saga fetch the current user.

src/components/chat-view-container/chat-view-container.tsx
dalefukami: Actually, maybe this should be one or two steps higher where we're specifically rendering the Channel. Eg: the `platform-app/channels/container.tsx` component.

src/store/channels/saga.ts
```diff
@@ -62,9 +90,19 @@ export function* clearChannels() {
   yield put(removeAll({ schema: schema.key }));
 }
 
+function* listenForChannelLoadedEvent() {
+  const channel = yield call(conversationsChannel);
+  while (true) {
+    const { channelId = undefined } = yield take(channel, ChannelEvents.MessagesLoadedForChannel);
```
dalefukami, line 96: Wonder if we can use the `takeEveryFromBus` helper that I created a couple weeks ago?

dalefukami (changes requested):

src/store/channels/saga.ts
ratik21: Works! Thankss

src/components/chat-view-container/chat-view-container.tsx
ratik21: created a separate PR for it

dalefukami (approved):
Looks good once the other PR is merged into `main`. Nicely done.

src/store/messages/saga.ts
```diff
@@ -361,6 +381,7 @@ export function* receiveNewMessage(action) {
         lastMessage: message,
       })
     ),
+    call(markAllAsReadAction),
```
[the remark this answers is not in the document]
ratik21: I thought this would be..automatically handled..? Since we're marking as read only for active channel/conversation.

src/store/channels/saga.ts
```diff
@@ -23,12 +25,7 @@ export function* joinChannel(action) {
   );
 }
 
-/**
- * Marks all messages as "read" in a channel for a specific user (queries zero-api & sendbird).
- */
-export function* markAllMessagesAsReadInChannel(action) {
-  const { channelId, userId } = action.payload;
-
+export function* markAllMessagesAsRead(channelId, userId) {
```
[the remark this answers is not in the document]
ratik21: will add a check. but this makes me realize i didn't test on the public zos mode. How do we mark those messages as read (currently) if the user-id isn't defined .. 🤔

src/store/messages/saga.ts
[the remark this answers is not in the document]
ratik21: also if we pass the channelId, wouldn't i mark the messages as read for a channel which isn't "active" (i.e receiveNewMessage on *channelX* but i am on *channelY*)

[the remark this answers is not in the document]
dalefukami: But those sagas both have that check already, don't the?

src/store/channels/saga.ts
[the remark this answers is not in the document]
dalefukami: We don't. There's no unread messages in public zos

src/store/channels/saga.ts
```diff
@@ -41,6 +42,30 @@ export function* markAllMessagesAsReadInChannel(action) {
...
+  const activeChannelInfo = activeChannelId ? yield select(rawChannelSelector(activeChannelId)) : null;
+
+  // just ensure first that you're not in full screen mode before marking all messages as read in a "channel"
+  if (!isMessengerFullScreen && activeChannelId && activeChannelInfo?.unreadCount > 0) {
+    yield call(markAllMessagesAsRead, activeChannelId, currentUser.id);
+  }
+}
+
+// mark all messages in read in current active conversation
+export function* markAllMessagesAsReadInCurrentConversation() {
```
dalefukami, line 59: Ah, I just realized these aren't as specific as I thought. This function should take the channel id and ONLY call `markAllMessagesAsRead` if the id passed in matches the active conversation. (and likewise for the channel one). Something like, `markUnreadIfActive(channelId)`

dalefukami: Otherwise, we'll be marking the active channel as read every time anything loads in the background.

src/store/messages/saga.ts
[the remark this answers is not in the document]
ratik21: no they don't 👀 

bdw i am not sure how to use `takeEveryFromBus` in case i wanted the event to return a data, and pass that data to the calling function (in this case the `channelId`)
image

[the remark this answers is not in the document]
dalefukami: Something like:
```
yield takeEveryFromBus(
  yield call(conversationsChannel),
  ChannelEvents.MessagesLoadedForConversation,
  (action) => markAllMessagesAsReadIfActive(action.channelId),
);
```

src/store/channels/saga.ts
ratik21: :+1:, updated

dalefukami (approved):

PR #750, opened 2023-07-06
Add padding to the "no results view"

ratik21:
### What does this do?

Based on Eth & Kelly's feedback, updated the top padding for the "no results" text to be consistent across both cases.

image
image

### DIFF
--- a/src/components/messenger/autocomplete-members/index.tsx
+++ b/src/components/messenger/autocomplete-members/index.tsx
@@ -7,6 +7,7 @@ import './styles.scss';
 import '../list/styles.scss';
 
 import { itemToOption } from '../lib/utils';
+import classNames from 'classnames';
 
 export interface Properties {
   search: (query: string) => Promise<Item[]>;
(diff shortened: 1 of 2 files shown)


src/components/messenger/autocomplete-members/index.tsx
```diff
@@ -55,7 +56,9 @@ export class AutocompleteMembers extends React.Component<Properties, State> {
         />
 
         {this.state.results?.length === 0 && this.state.results !== '' && (
-          <div className='messages-list__empty'>{`No results for '${this.state.searchString}' `}</div>
+          <div className={classNames('messages-list__empty', 'messages-list__empty-top-padding')}>
```
dalefukami, line 59: hmm... this feels like the wrong place just because we should be able to just render in "our" container appropriately. But, it works and probably isn't worth any more time :)

dalefukami (approved):

src/components/messenger/autocomplete-members/index.tsx
ratik21: we do in 2 places, 

here
https://github.com/zer0-os/zOS/blob/dab491d4ac6ede9497855845b202beec19e3459c/src/components/messenger/autocomplete-members/index.tsx#L58

and here
https://github.com/zer0-os/zOS/blob/dab491d4ac6ede9497855845b202beec19e3459c/src/components/messenger/list/conversation-list-panel.tsx#L115

I needed to add "top padding" in only one of these. So i added a new scss class and conditionally joined them in the place where i needed that.

ratik21: ok got your point, will keep in mind in case i stumble across this in the future again.

PR #751, opened 2023-07-06
Login and sign up pages should redirect users who are already logged in to the platform

ratik21:
IF you're already logged in, and you try to hit up login/signup page, it should redirect you to the index page.

https://github.com/zer0-os/zOS/assets/33264364/68700d32-37c7-4340-aef7-77d31ba85236


src/invite.tsx
```diff
@@ -15,13 +15,15 @@ import { CreateWalletAccountContainer } from './authentication/create-wallet-acc
...
+  shouldRender: boolean;
 }
 
 export class Container extends React.Component<Properties> {
   static mapState(state: RootState): Partial<Properties> {
-    const { registration } = state;
+    const { registration, pageload } = state;
     return {
       stage: registration.stage,
+      shouldRender: pageload.loadPage,
```
dalefukami, line 26: Just wanted to point out how awesome this subtle decision was. The state variable is named based on the context of the state and the component property is named based on it's impact. :rocket:

src/store/page-load/index.ts
```diff
@@ -0,0 +1,22 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+
+export type PageloadState = {
+  loadPage: boolean;
+};
+
+export const initialState = {
+  loadPage: false,
```
dalefukami, line 8: I might have gone with something more boolean-ish in name. `isLoaded`, or `isComplete`?  Then it'd later read as `pageLoad.isComplete`. That might be nice.

I'm not too worried about it since it's super easy to change later.

dalefukami (approved):

src/store/page-load/index.ts
ratik21: updated, thanks

PR #752, opened 2023-07-06
Add `setActiveChannelId` state

ratik21:
### What does this do?

Whenever you're on an active channel, it will update the global state and set it the channel. Having an `activeChannelId` & an `activeConversationId` as a global state is useful accross saga events.

### DIFF
--- a/src/components/chat-view-container/chat-view-container.tsx
+++ b/src/components/chat-view-container/chat-view-container.tsx
@@ -3,6 +3,7 @@ import classNames from 'classnames';
 import { RootState } from '../../store/reducer';
 
 import { connectContainer } from '../../store/redux-container';
+import { setActiveChannelId } from '../../store/chat';
 import {
   fetch as fetchMessages,
   send as sendMessage,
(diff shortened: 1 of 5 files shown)


src/components/chat-view-container/chat-view-container.tsx
```diff
@@ -106,12 +109,11 @@ export class Container extends React.Component<Properties, State> {
...
       this.props.stopSyncChannels(prevProps);
       this.props.fetchMessages({ channelId });
-      this.setState({
-        reply: null,
-      });
+      this.setState({ reply: null });
     }
 
     if (channelId && prevProps.user.data === null && this.props.user.data !== null) {
+      this.props.setActiveChannelId(channelId);
```
dalefukami, line 116: Do we need this any more? We already have the update events in the higher level component so this would just duplicate that, wouldn't it?

dalefukami (approved):
Just the one question about the extra event. I think we can remove that and ship it!

src/components/chat-view-container/chat-view-container.tsx
ratik21: I actually couldn't reproduce this one (condition), but left it just to make sure i don't break anything. I will remove it

PR #753, opened 2023-07-07
feat: send message input refactor and style update

domw30:
### What does this do?
- The changes required include adding the Send Icon when composing a message, in place of the audio record microphone icon. 
- Refactoring for readability and structure.
- Update UI/UX.

Scroll to be added in follow up PR.

- as per figma and feedback

[figma1](https://www.figma.com/file/aETpyuG2AKjNcQtCjldcX0/ZERO-Messenger-%2F-Web?node-id=9680%3A494050&mode=dev)
[figma2](https://www.figma.com/file/wTOVNZonl9MmEzORwBHDuq/zDS-Web-%26-Global?node-id=9909%3A59893&mode=dev)

### How do I test this?
- spin up the messenger chat and start typing to see updates.

#### Things to consider:
  1. Does this change any APIs?

https://github.com/zer0-os/zOS/assets/39112648/0985af7a-2bfa-47e2-b537-4d096e8317c9

With Scroll

https://github.com/zer0-os/zOS/assets/39112648/b57f3605-6e75-4602-9d39-858634e9627f


src/components/message-input/styles.scss
```diff
@@ -27,242 +27,182 @@
...
+      resize: none;
+    }
 
-    > div {
-      height: 32px !important;
-      width: 32px !important;
-      border-radius: 5px !important;
-      margin: -1px -1px 0 0 !important;
+    &__suggestions {
+      background: none !important;
```
ratik21, line 185: I am a little confused where we should use `!important` & where we should not use it. Could you please explain that..? 🤔

src/components/message-input/styles.scss
```diff
@@ -27,242 +27,182 @@
...
-    justify-content: center;
-    height: 32px;
-    width: 32px;
+    flex-direction: column;
+    flex: 1 0 0;
+    font-size: 14px;
+    line-height: 17px;
+
+    &__wrap {
+      &__suggestions {
```
ratik21, line 126: why not use `&__wrap-suggestions` here, if it's directly a child of the same class with no additional parent css?

src/components/message-input/index.tsx
```diff
@@ -290,78 +294,115 @@ export class MessageInput extends React.Component<Properties, State> {
   };
 
   renderInput() {
+    const hasInputValue = this.state.value?.length > 0;
+
+    console.log(this.props.isMessengerFullScreen);
```
ratik21: debugging log? I am not alone it seems :joy:

src/components/message-input/index.tsx
```diff
@@ -42,6 +44,7 @@ interface State {
   isEmojisActive: boolean;
   isGiphyActive: boolean;
   isMicActive: boolean;
+  isMessengerFullScreen?: boolean;
```
ratik21, line 47: why do we have this in the State AND in the Properties?

src/components/message-input/giphy/styles.scss
```diff
@@ -0,0 +1,54 @@
...
+
+  &--fullscreen {
+    right: 65px;
+  }
+
+  &-grid {
+    height: 440px;
+    overflow-y: overlay;
+    margin-top: 5px;
+    .loader {
```
ratik21, line 21: so just another css question here.. why use `.classname {..}` & not `&-classname` OR `__classname` (not sure)..?

src/components/message-input/index.tsx
```diff
@@ -1,32 +1,34 @@
...
+import classNames from 'classnames';
 import './styles.scss';
-import { MediaType } from '../../store/messages';
-import AttachmentCards from '../../platform-apps/channels/attachment-cards';
 
 export interface PublicProperties extends PublicPropertiesContainer {}
 
 export interface Properties extends PublicPropertiesContainer {
   viewMode: ViewModes;
+  isMessengerFullScreen?: boolean;
```
ratik21, line 31: I see that we're using `isFullScreen` above for the same. Maybe use that here as well (if possible)?

ratik21 (commented):
mostly css doubts

src/components/message-input/index.tsx
domw30: Good spot. Removed.

src/components/message-input/giphy/styles.scss
domw30: Old styles from a previous dev. I only moved the file into a new folder here, didn't take a look at the css. Something to check out on another task :)

src/components/message-input/index.tsx
domw30: I think for the `message-input`, `isMessengerFullscreen` gives it more context of what the attribute is for. Rather than assuming what exactly is `isFullScreen`

src/components/message-input/styles.scss
domw30: Ha good question. I couldn't get this working that way round, very odd! For the time it was worth, I just left it as is for now.

src/components/message-input/index.tsx
domw30: Haha, always my friend ! 

Removed.

src/components/message-input/styles.scss
domw30: We use !important when we want to make sure that a certain declaration will always apply, no matter what other stylesheets or styles might also be applicable to that HTML element.

Here !important is being used to override the top and background properties of &__suggestions and the font-size of the textarea. These styles are getting applied from elsewhere in the CSS and we want to ensure these specific styles apply in these cases i.e. for suggestions class.

Use of !important should be limited though as it can make debugging difficult at times.

src/components/message-input/emoji-picker/emoji-picker.test.tsx
```diff
@@ -1,6 +1,6 @@
 import { shallow } from 'enzyme';
 import React from 'react';
-import { ViewModes } from '../../shared-components/theme-engine';
+import { ViewModes } from '../../../shared-components/theme-engine';
```
dalefukami, line 3: Love these little enhancements but next time can you submit them as a separate PR either before or after the actual changes.

src/components/message-input/giphy/styles.scss
```diff
@@ -0,0 +1,54 @@
...
+  bottom: 65px;
+  right: 74px;
+  background-color: themeDeprecated.$background-color-tertiary;
+  padding: 5px;
+  border-radius: 5px;
+  border: 1px solid themeDeprecated.$input-arrow-color;
+
+  &--fullscreen {
+    right: 65px;
+  }
```
dalefukami, line 15: It feels very strange that a component at this low level needs to know whether it's fullscreen or not. Feels like we may have previously defined the positioning poorly and this is just a bandaid. Positioning

Let's leave it for now but keep this in mind for the future. Full screen mode should impact the major layout pieces but not lower level pieces directly. I think there's a missing concept here or just a funky lower level layout thing that we're not using properly.

dalefukami (commented):
Just a couple of notes for future changes. I'll leave it to Ratik to approve since he's already gone through it all :)

src/components/message-input/styles.scss
ratik21: ahhh, so you're basically overriding that style..got it. Thanks!

ratik21 (approved):

PR #759, opened 2023-07-10
Add pageload saga tests

ratik21:


src/store/page-load/saga.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    defaultZnsRoute: 'wilder',
+    defaultApp: 'nfts',
+  },
+}));
+
+jest.mock('../../lib/feature-flags', () => ({
+  featureFlags: {
+    allowPublicZOS: false,
+  },
+}));
```
ratik21, line 20: I wasn't able to conditionally mock it. Like in a test i wanted to set this to `true`, so i can test if the `initializePublicLayout` function is called, but i wasn't able to "invoke" this from within a test. Seems to work above the `describe('..', { .. })` level only i think 🤔

dalefukami:
```
const featureFlags = { allowWeb3Registration: false };
jest.mock('../../lib/feature-flags', () => ({
  featureFlags: featureFlags,
}));
```
Then in the tests:
```
featureFlags.allowWeb3Registration = true;
```

src/store/page-load/saga.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    // login
+    let history = new StubHistory('/login');
+    const { storeState: loginStoreState } = await expectSaga(saga)
+      .withReducer(rootReducer, initialState as any)
+      .provide(stubResponses(history, true))
+      .run();
+
+    // redirected from /login to /0.wilder.nfts
+    expect(history.replace).toHaveBeenCalledWith({ pathname: '/0.wilder/nfts' });
+    expect(loginStoreState.pageload.isComplete).toBe(false);
```
dalefukami, line 47: Interesting, I would have expected this to be `true` at the end if everything was successful. We're done with verifying the initial page load no matter which final condition is reached, right?

dalefukami (approved):
Thanks for adding this!

src/store/page-load/saga.test.ts
ratik21: thanks, works! Added one more test

src/store/page-load/saga.test.ts
ratik21: Now that i think about it, it should be `true`. Previously we were just setting the value & returning. I have updated the code and the test.

PR #760, opened 2023-07-10
fix: update class name to fix message input margin

domw30:
### What does this do?
- fixes the margin for the compose a message input.

- this change is to fix the margin to what is currently in production. This was not an intended design change, therefore the fix is required.

### How do I test this?
- open the app (not in full screen) and check the message input has equal margins as per screenshots below.

#### Things to consider:
  1. Does this change any APIs?

Before:
Screenshot 2023-07-10 at 10 28 36

Screenshot 2023-07-10 at 10 28 26

After:
Screenshot 2023-07-10 at 10 28 18

### DIFF
--- a/src/platform-apps/channels/styles.scss
+++ b/src/platform-apps/channels/styles.scss
@@ -67,7 +67,7 @@ $scrollbar-width: 5px;
   flex-direction: column;
   height: 100vh;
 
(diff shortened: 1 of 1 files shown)


ratik21 (approved):
thanks for fixing

PR #762, opened 2023-07-10
feat: conversation item preview mentions as plaintext

domw30:
### What does this do?
- adds a `negative` variant to the content highlighter so mentions are displayed in plain text if variant is set to `negative`.
- adds tests for content-highlighter

- as per figma desings.

### How do I test this?
- open app, navigate to a conversation and tag a user when composing a message. Send message and check UI in conversation item in sidekick.

#### Things to consider:
  1. Does this change any APIs?

Before:

Screenshot 2023-07-10 at 12 24 43

After:

Screenshot 2023-07-10 at 12 24 38

### DIFF
--- a/src/components/messenger/list/conversation-item.tsx
+++ b/src/components/messenger/list/conversation-item.tsx
@@ -153,7 +153,7 @@ export class ConversationItem extends React.Component<Properties> {
             </div>
             <div className={c('content')}>
               <div className={c('message')} is-unread={isUnread}>
-                <ContentHighlighter message={this.message} />
+                <ContentHighlighter message={this.message} variant='negative' />
               </div>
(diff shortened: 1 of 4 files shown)


src/components/messenger/list/conversation-item.tsx
```diff
@@ -153,7 +153,7 @@ export class ConversationItem extends React.Component<Properties> {
             </div>
             <div className={c('content')}>
               <div className={c('message')} is-unread={isUnread}>
-                <ContentHighlighter message={this.message} />
+                <ContentHighlighter message={this.message} variant='negative' />
```
dalefukami, line 156: Not sure `negative` holds any meaning to me...but I also can't come up with a suggestion. haha

dalefukami (approved):

PR #763, opened 2023-07-10
feat: add sender prefix to message preview

domw30:
### What does this do?
- adds sender prefix to message preview in conversation list item.
- updates `conversation-item` tests

- as per figma designs. [designs](https://www.figma.com/file/aETpyuG2AKjNcQtCjldcX0/ZERO-Messenger-%2F-Web?type=design&node-id=9931-278566&mode=design&t=2ZAsx1xZv4S9lMUZ-4)

### How do I test this?
- open messenger and check prefix of last sender in the sidekick conversation item component.

#### Things to consider:
  1. Does this change any APIs?

Before:
Screenshot 2023-07-10 at 12 36 19

After:
Screenshot 2023-07-10 at 12 36 17

### DIFF
--- a/src/components/messenger/list/conversation-item.tsx
+++ b/src/components/messenger/list/conversation-item.tsx
@@ -107,8 +107,12 @@ export class ConversationItem extends React.Component<Properties> {
 
     const { messagePreview, lastMessage } = this.props.conversation;
     if (messagePreview) {
+      const isAdminMessage = lastMessage?.admin && Object.keys(lastMessage.admin).length > 0;
+      if (isAdminMessage) return messagePreview;
+
       const isUserLastMessageSender = lastMessage?.sender?.userId === this.props.myUserId;
       const lastSenderDisplayName = isUserLastMessageSender ? 'You' : lastMessage.sender.firstName;
+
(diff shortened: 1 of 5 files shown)


src/components/messenger/list/conversation-item.tsx
```diff
@@ -107,7 +107,9 @@ export class ConversationItem extends React.Component<Properties> {
 
     const { messagePreview, lastMessage } = this.props.conversation;
     if (messagePreview) {
-      return messagePreview;
+      const isUserLastMessageSender = lastMessage?.sender?.userId === this.props.myUserId;
+      const lastSenderDisplayName = isUserLastMessageSender ? 'You' : lastMessage.sender.firstName;
```
dalefukami, line 111: Just want to double check if you verified this with admin messages. Admin messages have a sender but it's really a fake thing so I don't think we want any "sender" prefix for those ones.

dalefukami (approved):

src/components/messenger/list/conversation-item.tsx
domw30: Good shout. I've adjusted this.

PR #780, opened 2023-07-11
Remove `ChatConnect` (Connection Lost message)

ratik21:
### What does this do?

Removes this message

image

NOTE: I haven't removed the actual component, just it's usage (since we may need it later)


domw30 (approved):

dalefukami:
FWIW, I would remove the whole component. We can always find it again in the git history.

PR #781, opened 2023-07-12
feat: improve suggestions ui and add avatar

domw30:
### What does this do?
- updates suggestions ui to follow app theme
- adds scroll functionality to suggestions list
- wires up profile image (avatar) to user suggestions items
- resolves breaking tests

- as per figma design/requirements

### How do I test this?
- open chat and type '@' to trigger the mentions. Suggestions list will render as shown in demo below.

#### Things to consider:
  1. Does this change any APIs?

DEMO:

https://github.com/zer0-os/zOS/assets/39112648/ee58ab02-2c55-4732-8dc4-d758e73d4980

### DIFF
--- a/src/components/message-input/index.test.tsx
+++ b/src/components/message-input/index.test.tsx
@@ -155,29 +155,30 @@ describe('MessageInput', () => {
   });
 
   it('searches for matching users via userMentionSearch function', async function () {
-    const getUsersForMentions = async (_searchString) => Promise.resolve([{ id: '1', display: 'dale' }]);
+    const getUsersForMentions = async (_searchString) =>
+      Promise.resolve([{ id: '1', display: 'dale', profileImage: 'http://example.com' }]);
     const wrapper = subject({ getUsersForMentions });
 
     const searchResults = await userSearch(wrapper, 'da');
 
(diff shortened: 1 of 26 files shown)


src/components/message-input/index.test.tsx
```diff
@@ -155,29 +155,30 @@ describe('MessageInput', () => {
...
+    expect(searchResults).toEqual([{ display: 'dale', id: '1', profileImage: 'http://example.com' }]);
   });
 
   it('sorts by search string index', async function () {
     const getUsersForMentions = async (_searchString) =>
       Promise.resolve([
-        { id: 'd-2', display: '2-dale' },
-        { id: 'd-3', display: '3--dale' },
-        { id: 'd-1', display: 'dale' },
+        { id: 'd-2', display: '2-dale', profileImage: 'http://example-2.com' },
```
dalefukami, line 170: FYI, we should use only `example.com` here (not `example-1`, etc). `example.com` is the only reserved domain that nobody can register and we wouldn't want to be accidentally downloading something from somewhere else. So, `http://example.com/1` , etc, would be more appropriate.

Note: I'm sure I/we have made this mistake all over the place so if you do run into others it'd be nice to fix up.

dalefukami (approved):

PR #789, opened 2023-07-13
fix: message input hidden text when multiline

domw30:
### What does this do?
- fixes bug raised by product where some multiline text can be hidden when typing into the message input. 
- removes some hidden borders
- correctly sets font size and line height of text in input

- ui and ux enhancement
- bug fix

### How do I test this?
- type a range of long and short strings into the message input and check the correct behaviour is occurring. No text should ever be hidden unless scroll is triggered at max height of 200px.

#### Things to consider:
  1. Does this change any APIs?

Before: 

https://github.com/zer0-os/zOS/assets/39112648/dd9df5d4-5fd6-4acc-8b4d-625e5b360016

After:

https://github.com/zer0-os/zOS/assets/39112648/d9477d6c-0a5e-4866-b3ae-c73f2cfab370

### DIFF
--- a/src/components/message-input/styles.scss
+++ b/src/components/message-input/styles.scss
@@ -39,7 +39,7 @@
     display: flex;
     align-items: flex-end;
     gap: 16px;
-    align-self: stretch;
+    width: 100%;
   }
 
   &__icon-outer {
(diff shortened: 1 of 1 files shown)


src/components/message-input/styles.scss
```diff
@@ -142,7 +143,7 @@
       white-space: pre-wrap;
       background: transparent;
       overflow-wrap: break-word;
-      border: 1px solid transparent;
+      border: none !important;
```
dalefukami, line 146: Is this related to the input sizing?

src/components/message-input/styles.scss
```diff
@@ -155,11 +156,8 @@
     textarea {
       color: theme.$color-greyscale-12;
       word-break: break-word;
-      font-size: 14px !important;
-      line-height: 17px;
-      font-weight: 400;
       outline: none;
-      border: 1px solid transparent;
+      border: none;
```
dalefukami, line 160: Same question here

dalefukami (approved):

PR #794, opened 2023-07-13
Group conversation messages shouldn't fill the max width

dalefukami:
### What does this do?

Fixes the issue where group conversation messages were filling the max width available rather than being limited to the size of the content.

Before:
image

After:
image


ratik21 (approved):
🎉

PR #799, opened 2023-07-13
Zos 564 entering a special character eg (`/`) in conversations search breaks the app

ratik21:
IF you enter `/` or `*`..etc it breaks the whole app. 

Before

https://github.com/zer0-os/zOS/assets/33264364/e37383b4-93dc-4b79-930a-c6bf82a8a1bb

After

https://github.com/zer0-os/zOS/assets/33264364/ba7a46b5-1e24-4808-a1e4-d903846f1524

### DIFF
--- a/src/components/messenger/list/utils.tsx
+++ b/src/components/messenger/list/utils.tsx
@@ -11,11 +11,3 @@ export function lastSeenText(user): string {
 
   return 'Last Seen: ' + (user.lastSeenAt ? fromNow(user.lastSeenAt) : 'Never');
 }
-
-export function escapeRegExp(string) {
-  if (typeof string !== 'string') {
-    return;
-  }
-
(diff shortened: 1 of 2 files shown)


src/components/messenger/list/utils.tsx
```diff
@@ -11,3 +11,11 @@ export function lastSeenText(user): string {
 
   return 'Last Seen: ' + (user.lastSeenAt ? fromNow(user.lastSeenAt) : 'Never');
 }
+
+export function escapeRegExp(string) {
```
dalefukami, line 15: `import escapeRegExp from "lodash/escapeRegExp"`

Maybe we should use lodash? I imagine they'd maintain this better than we would.

dalefukami (approved):

src/components/messenger/list/utils.tsx
ratik21: i didn't knew it existed. Thanks!

PR #821, opened 2023-07-19
Zos 539: Display zero as large number and usd value as subtext in rewards modal

ratik21:
Design Changes

Before

image

After:

image

image

image

image

EMPTY STATE

image


dalefukami (approved):
Looks great. Only questions is whether we need to call our api or if we can just call the public page directly.

PR #822, opened 2023-07-19
documentation: update README with styling and SCSS usage details

domw30:
Proposed style documentation for README (or to be stored elsewehere).


dalefukami (approved):
Love it, thanks for doing this. Do we want to make a separate `docs` folder so we can split this specific kind of stuff up?

PR #827, opened 2023-07-19
Add a dev panel for message status testing

dalefukami:
### What does this do?

Adds a dev panel to test the various message states. As you can see, very little effort was spent styling :)

It's frequently difficult to get into the various states a message can be in when it comes to the sending flow. This adds a hidden dev panel that allows a user to toggle the various states so we can quickly see how it impacts the view.

### How do I test this?

Enable via `window.FEATURE_FLAGS.enableDevPanel = true` and refresh

https://github.com/zer0-os/zOS/assets/43770/bf313fbf-96da-40d6-a514-dce8b2e2f392


domw30 (approved):
This is nice! Very handy 💯

ratik21 (approved):

PR #828, opened 2023-07-20
Cache `node_modules` (reduces build time by a min)

ratik21:
### What does this do?

Caches `node_modules` in the github action, which will be reused by subsequent pushed. The cache is **rebuilt** if there is a change in OS, `package-lock.json`, the `node version`.

Before (notice the times on the right)

image

After

image

Time took by `npm install` before 
image

Time took by `npm install` after (this is where the cache "hits")
image

Cache:
image


.github/workflows/development.yml
```diff
@@ -50,11 +65,25 @@ jobs:
...
+        run: |
+          echo "version=$(node -v)" >> $GITHUB_OUTPUT
+
+      - name: Get node_modules cache
+        uses: actions/cache@v3.0.2
+        id: node_modules
+        with:
+          path: |
+            **/node_modules
+          key: ${{ runner.os }}-node_modules-${{ hashFiles('**/package-lock.json') }}-${{ steps.node.outputs.version }}
```
ratik21, line 85: I wanted to use a composite action here (a common setup file), since this is repatead in each workflow

Something like:
```yml
name: Common Setup

on:
  workflow_run:
    workflows: ['Deploy Development', 'run-build', 'run-lint', 'run-tests', 'run-typecheck']
    types:
      - requested

jobs:
  common-setup:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout code
        uses: actions/checkout@v3

      - name: Setup Node.js
        uses: actions/setup-node@v3
        with:
          node-version: '16'

      - name: Get node version
        id: node
        run: echo "version=$(node -v)" >> $GITHUB_OUTPUT

      # rebuild the cache if OS, package-lock.json or node version changes
      - name: Get node_modules cache
        uses: actions/cache@v3.0.2
        id: node_modules
        with:
          path: |
            **/node_modules
          key: ${{ runner.os }}-node_modules-${{ hashFiles('**/package-lock.json') }}-${{ steps.node.outputs.version }}
```

And then in the actual workflow use:
```yml
- uses: actions/checkout@v2
- uses: ./.github/workflows/common/setup
```

I tried a few things but that didn't work 🤔

dalefukami (approved):
Awesome, thanks!

If you wanna try one more thing I believe the api repo has a re-usable action example. I'm guessing you already tried exactly what it does but just in case it triggers an idea.

ratik21:
@dalefukami thanks for the reference, i was able to create a reusable action.

Action
```yml
name: 'Common Setup'

description: 'Setup nodejs and cache'

runs:
  using: 'composite'
  steps:
    - name: Checkout code
      uses: actions/checkout@v3

    - name: Setup Node.js
      uses: actions/setup-node@v3
      with:
        node-version: '16'

    - name: Get node version
      id: node
      shell: bash
      run: echo "version=$(node -v)" >> $GITHUB_OUTPUT

    # rebuild the cache if OS, package-lock.json or node version changes
    - name: Get node_modules cache
      uses: actions/cache@v3.0.2
      id: node_modules
      with:
        path: |
          **/node_modules
        key: ${{ runner.os }}-node_modules-${{ hashFiles('**/package-lock.json') }}-${{ steps.node.outputs.version }}
```

now in the workflow, we can reduce the code and just use the common setup (eg. in typecheck):

```yml
name: run-typecheck
on:
  pull_request: {}

concurrency:
  group: ${{ github.workflow }}-${{ github.head_ref }}
  cancel-in-progress: true

jobs:
  run-typecheck:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - name: Common Setup
        uses: ./.github/actions/common

      - run: echo '//npm.pkg.github.com/:_authToken=${{secrets.GITHUB_TOKEN}}' >> .npmrc
      - run: npm install
      - name: Build project
        run: npx tsc
```

PR #840, opened 2023-07-21
Zos 578: Show error when messenger gets stuck in perma skeleton loader state

ratik21:
Messenger (full screen):

https://github.com/zer0-os/zOS/assets/33264364/e684c03c-dbd9-41a2-a554-0284ad40f368

Channel on zOS:

https://github.com/zer0-os/zOS/assets/33264364/4286c6d6-fa63-4da7-be67-5fdce761bf04

Messenger (normal):

https://github.com/zer0-os/zOS/assets/33264364/64a76adb-0b07-4c0a-89d7-ea2e302c705c

### DIFF
--- a/src/components/chat-view-container/chat-view.test.tsx
+++ b/src/components/chat-view-container/chat-view.test.tsx
@@ -243,17 +243,46 @@ describe('ChatView', () => {
     expect(wrapper).not.toHaveElement('ChatSkeleton');
   });
 
+  it('renders failure message if message load failed', () => {
+    let wrapper = subject({
+      messages: MESSAGES_TEST,
+      messagesFetchStatus: MessagesFetchState.FAILED,
+      hasLoadedMessages: true,
+    });
+    expect(wrapper.find('.channel-view__failure-message').text()).toContain('Failed to load new messages. Try Reload');
+
+    wrapper = subject({
+      messages: MESSAGES_TEST,
+      messagesFetchStatus: MessagesFetchState.FAILED,
+      hasLoadedMessages: false,
+      name: 'channel-name',
+    });
+
+    expect(wrapper.find('.channel-view__failure-message').text()).toContain(
+      'Failed to load conversation with channel-name. Try Reload'
+    );
+
+    wrapper = subject({
+      messages: MESSAGES_TEST,
+      messagesFetchStatus: MessagesFetchState.FAILED,
+      hasLoadedMessages: false,
+      name: '',
+      otherMembers: [{ firstName: 'ratik', lastName: 'jindal' } as any],
+    });
+
+    expect(wrapper.find('.channel-view__failure-message').text()).toContain(
+      'Failed to load your conversation with ratik. Try Reload'
+    );
+  });
+
   it('renders failure message if message load failed, and fetches messages on reload', () => {
     const fetchMessages = jest.fn();
-
     let wrapper = subject({
       messages: MESSAGES_TEST,
       messagesFetchStatus: MessagesFetchState.FAILED,
       fetchMessages,
       id: 'channel-id',
     });
-
-    expect(wrapper.find('.channel-view__failure-message').text()).toContain('Failed to load messages. Try Reload');
 
     wrapper.find('.channel-view__try-reload').simulate('click');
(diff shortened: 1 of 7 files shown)


src/components/chat-view-container/chat-view.tsx
```diff
@@ -240,10 +243,28 @@ export class ChatView extends React.Component<Properties, State> {
...
+              <ChatSkeleton conversationId={this.props.id} />
+            )}
+
             <div ref={this.bottomRef} />
           </div>
         </InvertedScroll>
+
+        {this.props.messagesFetchStatus === MessagesFetchState.FAILED && (
+          <div className='channel-view__failure-message'>
+            Failed to load messages.&nbsp;
```
ratik21, line 256: The message is not per design (`Failed to load messages`), and will change. I have asked eth/kelly for a message to show when you're in a group conversation or a normal zOS channel

src/components/chat-view-container/chat-view.test.tsx
```diff
@@ -220,15 +223,31 @@ describe('ChatView', () => {
     expect(wrapper.find(`.${className}`).exists()).toBe(true);
   });
 
-  it('renders skeleton if messages have not been loaded yet', () => {
-    const wrapper = subject({ messages: MESSAGES_TEST, hasLoadedMessages: false });
+  it('renders skeleton if messages are being loaded', () => {
```
dalefukami, line 226: This doesn't sound right. This means that every time we switch conversations we'll re-show the skeleton again, I think. We definitely don't want that. If the messages have been loaded previously we just want to render them immediately. That's one thing that gives the illusion speed.

src/components/chat-view-container/chat-view.tsx
```diff
@@ -240,10 +243,28 @@ export class ChatView extends React.Component<Properties, State> {
...
-            {!this.props.hasLoadedMessages && <ChatSkeleton conversationId={this.props.id} />}
+            {this.props.messagesFetchStatus === MessagesFetchState.IN_PROGRESS && (
+              <ChatSkeleton conversationId={this.props.id} />
+            )}
+
             <div ref={this.bottomRef} />
           </div>
         </InvertedScroll>
+
+        {this.props.messagesFetchStatus === MessagesFetchState.FAILED && (
```
dalefukami, line 254: I think maybe we need to be careful here. What if we've already previously loaded the messages and then we open a conversation? We show the existing messages but we also then fetch in the background to kinda ensure we get the latest stuff. But, if that one fails I don't think we want to clear the entire conversation. We might not even want to show the actual error....maybe we don't care...or maybe we want a more subtle indicator.

dalefukami (changes requested):

src/components/chat-view-container/chat-view.test.tsx
ratik21: I had missed that we're doing that. Updated the code to only show the failure message ONLY If the message fetch has failed.

src/components/chat-view-container/chat-view.tsx
ratik21: > But, if that one fails I don't think we want to clear the entire conversation. We might not even want to show the actual error....maybe we don't care...or maybe we want a more subtle indicator.

I don't know...what the right answer is here. For me personally i agree that we shouldn't clear the entire conversation. BUT i think not showing "any" error is a more buggy state..since we know that the user has missed some latest messages..but the user doesn't know that! 

I think in this case we should show an error at the bottom, with keeping the previous conversation as is. And example message could be "Some messages failed to load. Try Reloading".

ratik21: In this PR i am doing the same in theory (keeping the previous conversation as is). But ..just not able to reproduce the state so that i actually see it on my local (i.e allowing the first message fetch, and then failing the next ones for the same channel). Will try more.

dalefukami: Yep, that's the kind of thing Eth should have input on.

ratik21:
Screens as per new design:

image

image

image

image

src/components/chat-view-container/chat-view.tsx
ratik21: looks like this now

image

image

dalefukami (approved):

PR #846, opened 2023-07-24
feat: add edit message icon tooltips (save / discard)

domw30:
### What does this do?
- adds tooltips to the edit message action icons.

- as per figma design.

### How do I test this?
- open messenger. hover over a message you have sent and click edit. hover over the 'x' icon to see 'discard changes', and hover over the check icon to see 'save changes'

#### Things to consider:
  1. Does this change any APIs?

https://github.com/zer0-os/zOS/assets/39112648/98806753-32f9-4f1c-95de-021d9e652e32

### DIFF
--- a/src/components/message/edit-message-actions/styles.scss
+++ b/src/components/message/edit-message-actions/styles.scss
@@ -14,14 +14,3 @@
     color: theme.$color-greyscale-12;
   }
 }
-
-.zui-tooltip__content {
-  padding: 10px 8px;
-  border-radius: 8px;
-  font-size: 14px !important;
-}
-
-.zui-tooltip__arrow {
-  width: 16px;
-  height: 8px;
-}
--- a/src/components/message/edit-message-actions/styles.scss
+++ b/src/components/message/edit-message-actions/styles.scss
@@ -14,3 +14,14 @@
     color: theme.$color-greyscale-12;
   }
 }
+
+.zui-tooltip__content {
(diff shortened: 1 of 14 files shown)


src/components/message/edit-message-actions/styles.scss
```diff
@@ -15,3 +14,14 @@
     color: theme.$color-greyscale-12;
   }
 }
+
+.zui-tooltip__content {
```
dalefukami, line 18: I don't think we should be overriding zui component styles. This could cause every tooltip in the whole app (even all the zapps) to take on these properties.

domw30: Okay makes sense. I'll need to go into zUI and make some changes for these

dalefukami: Are these what we want _all_ tooltips to look like? If so, we can just delete these from here and merge this and then update zUI as a follow up.

domw30: This is the component in the design system - https://www.figma.com/file/wTOVNZonl9MmEzORwBHDuq/zDS-Web-%26-Global?type=design&node-id=1641-3344&mode=design&t=TXIBEvHDcJ0gmcZN-4. 

I think this would mean updating zUI. However, I can remove these overriding styles for now and just use what zUI provides for the time being

dalefukami (approved):

PR #850, opened 2023-07-25
feat: remove antiphonal text from searches returning no results

domw30:
### What does this do?
- removes antiphonal text as shown in the demo below.

- requested from product/n3o.

### How do I test this?
- enter characters into one of the search inputs in the sidekick until no users are returned. Check that there is nothing rendered when this happens.

#### Things to consider:
  1. Does this change any APIs?

Before:

https://github.com/zer0-os/zOS/assets/39112648/891d901d-e2ab-4bc9-a3ea-d59fa468f806

After: 

https://github.com/zer0-os/zOS/assets/39112648/04b16d7f-b68d-4e9d-b6c9-6722be8f8f54

### DIFF
--- a/src/components/messenger/autocomplete-members/index.tsx
+++ b/src/components/messenger/autocomplete-members/index.tsx
@@ -7,7 +7,6 @@ import './styles.scss';
 import '../list/styles.scss';
 
 import { highlightFilter, itemToOption } from '../lib/utils';
-import classNames from 'classnames';
 
 export interface Properties {
   search: (query: string) => Promise<Item[]>;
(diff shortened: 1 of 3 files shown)


src/components/messenger/autocomplete-members/index.tsx
```diff
@@ -65,12 +64,6 @@ export class AutocompleteMembers extends React.Component<Properties, State> {
           inputClassName={'autocomplete-members__search-input'}
         />
 
-        {this.state.results?.length === 0 && this.state.results !== '' && (
-          <div className={classNames('messages-list__empty', 'messages-list__empty-top-padding')}>
```
ratik21, line 69: do we need the `messages-list__empty` in the styles now?

domw30: Great spot! Thanks @ratik21

ratik21 (approved):

PR #852, opened 2023-07-25
fix: rewards icon button in the sidebar

domw30:
### What does this do?
- Update the space between the word “Rewards” and the Rewards icon to be 4 px
- Update the hover text to no longer be bold on hover
- Update the text to no longer be bold when active
- Update the hover state to match the specification in the above Figma file
- Orange badge should hide as soon as rewards modal is opened
- Update rewards bar buttons as per design
- Updates rewards button icon status design
- Adds test coverage

- as per design (https://www.figma.com/file/aETpyuG2AKjNcQtCjldcX0/ZERO-Messenger-%2F-Web?type=design&node-id=6797-369406&mode=dev)

### How do I test this?
- open full messenger and check rewards bar/rewards button. Compare with design linked above.

#### Things to consider:
  1. Does this change any APIs?

*Updated design*

https://github.com/zer0-os/zOS/assets/39112648/5ad75599-9491-473d-bcb9-151b413225e0

### DIFF
--- a/src/components/rewards-bar/index.test.tsx
+++ b/src/components/rewards-bar/index.test.tsx
@@ -92,6 +92,28 @@ describe('rewards-bar', () => {
     expect(wrapper.find(TooltipPopup).prop('open')).toBeFalse();
   });
 
+  it('displays rewards icon status when conditions are met', function () {
+    const wrapper = subject({
+      showNewRewards: true,
+      isMessengerFullScreen: true,
+    });
+
+    wrapper.setState({ isRewardsPopupOpen: false });
+
+    expect(wrapper.find('.rewards-bar__rewards-icon__status')).toHaveLength(1);
+  });
+
+  it('does not display rewards icon status when rewards pop up is open', function () {
+    const wrapper = subject({
+      showNewRewards: false,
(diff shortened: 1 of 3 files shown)


src/components/rewards-bar/index.test.tsx
```diff
@@ -92,6 +92,28 @@ describe('rewards-bar', () => {
     expect(wrapper.find(TooltipPopup).prop('open')).toBeFalse();
   });
 
+  it('displays rewards icon status when conditions are met', function () {
+    const wrapper = subject({
+      showNewRewards: true,
+      isMessengerFullScreen: true,
+    });
+
+    wrapper.setState({ isRewardsPopupOpen: false });
```
dalefukami, line 101: It'd be preferable to perform the actual action that causes this state to get updated rather than have the test "just know" the internal state of the component. So, find the button and simulate the click event kinda thing.

dalefukami (approved):

src/components/rewards-bar/index.test.tsx
domw30: Nice suggestion thanks! I have updated :)

PR #853, opened 2023-07-25
feat: populate messages from the bottom up

domw30:
### What does this do?
- populates messages (where the container uses `InvertedScroll` from the bottom up.

- as per figma designs / product request.

### How do I test this?
- open messenger fullscreen or collapsed direct message and check that messages are populated from the bottom up - as per screenshots below.

#### Things to consider:
  1. Does this change any APIs?

Before: 

Screenshot 2023-07-25 at 15 42 57

After:

Screenshot 2023-07-25 at 15 43 27

### DIFF
--- a/src/components/inverted-scroll/styles.scss
+++ b/src/components/inverted-scroll/styles.scss
@@ -3,6 +3,10 @@
   overflow-x: hidden;
 
   .scroll-container__children {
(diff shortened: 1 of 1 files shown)


dalefukami (approved):
Nice! I'm glad that ended up working out.

PR #857, opened 2023-07-27
Zos 600: Enable users to edit their profile

ratik21:
### What does this do?

Enables users to edit their profiles. 

SC:

image

image

image

Flow:

https://github.com/zer0-os/zOS/assets/33264364/0ae90a9c-770d-454c-956a-1ac703c7089b

### DIFF
--- a/src/components/edit-profile/container.tsx
+++ b/src/components/edit-profile/container.tsx
@@ -1,0 +1,63 @@
+import * as React from 'react';
+import { RootState } from '../../store/reducer';
+import { connectContainer } from '../../store/redux-container';
+import { EditProfile } from '.';
+import { editProfile, setChangesSaved } from '../../store/edit-profile';
+import { Container as RegistrationContainer } from '../../authentication/create-account-details/container';
+export interface PublicProperties {
+  onClose?: () => void;
+}
+
+export interface Properties extends PublicProperties {
+  isLoading: boolean;
+  errors: {
+    image?: string;
+    name?: string;
+    general?: string;
+  };
+  changesSaved: boolean;
+  currentDisplayName: string;
+  currentProfileImage: string;
+  editProfile: (data: { name: string; image: File }) => void;
+  setChangesSaved: (value: boolean) => void;
+}
+
+export class Container extends React.Component<Properties> {
+  static mapState(state: RootState) {
+    const {
+      editProfile,
+      authentication: { user },
+    } = state;
+    return {
+      isLoading: editProfile.loading,
+      errors: RegistrationContainer.mapErrors(editProfile.errors),
+      changesSaved: editProfile.changesSaved,
+      currentDisplayName: user?.data?.profileSummary.firstName,
+      currentProfileImage: user?.data?.profileSummary.profileImage,
+    };
+  }
+
+  static mapActions(_props: Properties): Partial<Properties> {
+    return { editProfile, setChangesSaved };
+  }
+
+  componentWillUnmount(): void {
+    this.props.setChangesSaved(false);
+  }
+
+  render() {
+    return (
+      <EditProfile
+        isLoading={this.props.isLoading}
+        onEdit={this.props.editProfile}
+        errors={this.props.errors}
+        onClose={this.props.onClose}
+        changesSaved={this.props.changesSaved}
+        currentDisplayName={this.props.currentDisplayName}
+        currentProfileImage={this.props.currentProfileImage}
+      />
+    );
+  }
+}
+
+export const EditProfileContainer = connectContainer<PublicProperties>(Container);
--- a/src/components/edit-profile/container.tsx
+++ b/src/components/edit-profile/container.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { RootState } from '../../store/reducer';
 import { connectContainer } from '../../store/redux-container';
 import { EditProfile } from '.';
-import { editProfile, setChangesSaved } from '../../store/edit-profile';
+import { editProfile, startProfileEdit } from '../../store/edit-profile';
 import { Container as RegistrationContainer } from '../../authentication/create-account-details/container';
 export interface PublicProperties {
   onClose?: () => void;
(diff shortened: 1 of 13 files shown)


src/components/settings-menu/index.test.tsx
```diff
@@ -22,13 +23,23 @@ describe('settings-menu', () => {
     expect(wrapper).toHaveElement(DropdownMenu);
   });
 
+  it('calls onLogout prop when logout item is selected', function () {
```
dalefukami, line 26:
```suggestion
  it('opens profile dialog', function () {
```

src/store/edit-profile/index.ts
```diff
@@ -0,0 +1,38 @@
+import { createAction, createSlice, PayloadAction } from '@reduxjs/toolkit';
+
+export enum SagaActionTypes {
+  EditProfile = 'profile/edit',
+}
+
+export type EditProfileState = {
+  loading: boolean;
+  errors: string[];
+  changesSaved: boolean;
```
dalefukami, line 10: :thinking: would the `loading` and `changesSaved` combo be better represented as a `state` enum? Like: `NONE`, `IN_PROGRES`, `SUCCESS` or something along those lines.

src/store/edit-profile/api.ts
```diff
@@ -0,0 +1,27 @@
...
+  profileImage?: string;
+}) {
+  try {
+    const response = await patch(`/api/profiles/${profileId}`).send({ profileId, firstName: name, profileImage });
+    return {
+      success: true,
+      response: response.body,
+    };
+  } catch (error: any) {
+    if (error?.response?.status === 400) {
```
dalefukami, line 19: Doesn't seem like we do anything with these errors in the saga so maybe we should just not catch the error here and let it throw. Did we confirm that errors for this endpoint come back in this format? (moot point if we remove this)

src/components/settings-menu/index.tsx
```diff
@@ -66,39 +85,48 @@ export class SettingsMenu extends React.Component<Properties, State> {
...
+            {
+              id: 'header',
+              label: this.renderSettingsHeader(),
+              onSelect: () => {},
+            },
+            {
+              className: 'edit_profile',
+              id: 'edit_profile',
+              label: this.renderSettingsOption(<IconUser1 />, 'Edit Profile'),
+              onSelect: () => this.openEditProfileDialog(),
```
dalefukami, line 101:
```suggestion
              onSelect: this.openEditProfileDialog,
```

No need for the lambda. Might as well just pass the reference to the function.

dalefukami (approved):
Nice, looks pretty good. Just a few comments. I'll let you decide which ones should be handle before merging and which can be follow ups.

src/components/settings-menu/index.tsx
```diff
@@ -66,39 +85,48 @@ export class SettingsMenu extends React.Component<Properties, State> {
...
-        trigger={
-          <Avatar
-            isActive={this.state.isDropdownOpen}
-            size={'medium'}
-            type={'circle'}
-            imageURL={this.props.userAvatarUrl}
-          />
-        }
-      />
+      <div>
```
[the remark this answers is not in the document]
ratik21: we don't. updated to fragment

src/store/authentication/saga.ts
```diff
@@ -125,7 +126,22 @@ export function* clearUserState() {
   ]);
 }
 
+function* listenForUserProfileUpdates() {
+  const authChannel = yield call(getAuthChannel);
+  while (true) {
+    const { payload } = yield take(authChannel, Events.UserProfileUpdated);
```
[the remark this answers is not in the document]
ratik21: works! Sorry i forgot.

src/store/edit-profile/saga.ts
```diff
@@ -0,0 +1,53 @@
...
+    if (response.success) {
+      yield call(publishUserProfileEdited, { name, profileImage });
+      return;
+    } else {
+      yield put(setErrors([ProfileDetailsErrors.UNKNOWN_ERROR]));
+    }
+  } catch (e) {
+    yield put(setErrors([ProfileDetailsErrors.UNKNOWN_ERROR]));
+  } finally {
+    yield put(setChangesSaved(true));
```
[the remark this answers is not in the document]
ratik21: makes sense. updated.

src/store/edit-profile/api.ts
ratik21: > Did we confirm that errors for this endpoint come back in this format?

We don't have specific 400 Errors for the PATCH endpoint, no. We can simply throw here.

src/components/edit-profile/container.tsx
```diff
@@ -0,0 +1,63 @@
...
+      currentProfileImage: user?.data?.profileSummary.profileImage,
+    };
+  }
+
+  static mapActions(_props: Properties): Partial<Properties> {
+    return { editProfile, setChangesSaved };
+  }
+
+  componentWillUnmount(): void {
+    this.props.setChangesSaved(false);
```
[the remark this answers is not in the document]
ratik21: done, thanks for suggestion.

src/store/edit-profile/index.ts
ratik21: done

src/store/authentication/channels.ts
```diff
@@ -4,6 +4,7 @@ import { call } from 'redux-saga/effects';
 export enum Events {
   UserLogin = 'user/login',
   UserLogout = 'user/logout',
+  UserProfileUpdated = 'user/profile-updated',
```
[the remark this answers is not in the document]
ratik21: hmm, makes sense. I don't think any existing multicast channel is good enough. (Currently we have buses for `channels`/`conversations`, `eventChannels` for sendbird events, and the `authChannel` for authentication). We'll likely need to add a new bus i think. I'll discuss more async, & implement this one as a follow up.

PR #858, opened 2023-07-27
feat: update invite screen and layout

domw30:
### What does this do?
- updates invite screen to follow latest designs

- as per figma designs

### How do I test this?
- visit zOS `/get-access` page. Check against designs [here](https://www.figma.com/file/aETpyuG2AKjNcQtCjldcX0/ZERO-Messenger-%2F-Web?type=design&node-id=9745-269425&mode=design&t=tEwW07rdhHz6gJWy-4)

#### Things to consider:
  1. Does this change any APIs?

Screenshot 2023-07-27 at 11 44 04

Screenshot 2023-07-27 at 11 44 09

Screenshot 2023-07-27 at 11 50 08


src/authentication/footer/footer.tsx
```diff
@@ -0,0 +1,25 @@
...
+  render() {
+    return (
+      <div className={c('')}>
+        <div>
+          <span>Already on ZERO? </span>
+          <Link to='/login'>Log in</Link>
+        </div>
+
+        <div>
+          <span>Dont have an invite code? </span>
```
ratik21, line 19:
```suggestion
          <span>Don't have an invite code? </span>
```

src/authentication/footer/styles.scss
```diff
@@ -0,0 +1,12 @@
+@import '../styles';
+
+.invite-footer {
+  @include base();
```
ratik21, line 4: (doubt) Why do we need to incluse "base" in the footer file? Instead shouldn't we include "footer as base()" in the main `styles.scss` file?

ratik21 (approved):

src/authentication/footer/footer.tsx
domw30: I'll check this with Eth as this is what is shown in figma :)

src/authentication/footer/styles.scss
domw30: yeah we can do that for sure. Updated

src/authentication/footer/styles.scss
```diff
@@ -0,0 +1,5 @@
+@import '../styles';
+
+.invite-footer {
+  @include footer();
```
dalefukami, line 4: Is anything else using the `footer` mixin? If not, we should just put the styles here. The intent of the mixins is when multiple pages need the same styling but if all those pages are just gonna use the the same component then we don't need the mixin.

domw30: Nice, I changed it to this from Ratiks suggestion but I'll re-add the styles here and remove the footer mixin. Thanks Dale!

src/authentication/validate-invite/index.tsx
```diff
@@ -69,48 +68,49 @@ export class Invite extends React.Component<Properties, State> {
...
+              error={this.state.renderAlert}
+            />
+
+            {this.state.renderAlert &&
+              this.props.inviteCodeStatus !== InviteCodeStatus.VALID &&
+              this.renderAlert(this.props.inviteCodeStatus)}
+          </div>
 
           <Button
+            className={c('submit-button')}
```
dalefukami, line 101: Is there a reason to add this class? Doesn't seem like we're using it anywhere.

domw30: Ah, I will be using it in the follow up PRs for this story, but not using in this PR.

It will replace the `> *:last-child {}` here in `authentication/_styles.scss`

```
@mixin form {
  display: flex;
  flex-direction: column;
  align-items: center;
  gap: 32px;

  // Button should be the last element of the form.
  // Once button component has the ability to select the large option this should
  // no longer be needed.
  > *:last-child {
    height: 40px;
    padding: 16px;

    div {
      text-transform: none;
      margin: 0;
    }
  }
}
```

src/authentication/footer/styles.scss
dalefukami: Yeah, there's just so many ways to share code it's hard to choose the "right" one. lol

PR #862, opened 2023-07-27
fix: invite screen error handling

domw30:
### What does this do?
- improves/fixes the handling of errors on the invite screen.
- alert remains visible when there has been an error.
- adds the error state of the input when rendering alert.
- disables button when invite code has not been edited as per design.

- to improve the UI/UX of the invite screen.

### How do I test this?
- navigate to /get-access. enter invalid code. then enter valid code.

#### Things to consider:
  1. Does this change any APIs?

OUTCOME:

https://github.com/zer0-os/zOS/assets/39112648/07620972-5ca8-47a6-a4c2-4917fa73c8ac

### DIFF
--- a/src/authentication/validate-invite/index.tsx
+++ b/src/authentication/validate-invite/index.tsx
@@ -20,12 +20,14 @@ const MAX_INVITE_CODE_LENGTH = 14;
 interface State {
   inviteCode: string;
   renderAlert: boolean;
+  lastSubmittedInviteCode: string;
 }
 
 export class Invite extends React.Component<Properties, State> {
   state: State = {
     inviteCode: '',
     renderAlert: false,
+    lastSubmittedInviteCode: '',
   };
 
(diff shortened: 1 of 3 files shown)


src/authentication/validate-invite/index.tsx
```diff
@@ -34,17 +36,10 @@ export class Invite extends React.Component<Properties, State> {
 
   submitForm = async (e) => {
     e.preventDefault();
-    this.setState({ renderAlert: true });
+    this.setState({ renderAlert: true, lastSubmittedInviteCode: this.state.inviteCode });
     this.props.validateInvite({ code: this.state.inviteCode });
   };
 
-  componentDidUpdate(_prevProps: Readonly<Properties>, prevState: Readonly<State>): void {
```
dalefukami, line 41: So happy when we can remove `componentDidUpdate`!

PR #863, opened 2023-07-27
Remove redundant event listener for creating conversations

dalefukami:
### What does this do?

Removes an extra event listener that already existed

We had 2 listeners to the same event under the same conditions which was causing us to send 2 api calls for the same thing.

### How do I test this?

Create new conversations in the various ways (particularly one on ones)

### DIFF
--- a/src/store/create-conversation/saga.ts
+++ b/src/store/create-conversation/saga.ts
@@ -56,7 +56,6 @@ export function* createConversation(action) {
 
 export function* saga() {
   yield fork(authWatcher);
-  yield fork(handleCreateConversation);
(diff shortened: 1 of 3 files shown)


src/store/create-conversation/saga.ts
```diff
@@ -56,7 +56,6 @@ export function* createConversation(action) {
 
 export function* saga() {
   yield fork(authWatcher);
-  yield fork(handleCreateConversation);
```
dalefukami, line 59: You can see below we're already listening for this event. No reason to listen here.

PR #864, opened 2023-07-27
Clear the search state when a conversation is selected

dalefukami:
### What does this do?

When using the quick search in the conversation list, clear the filter state when a conversation is selected.

It's very confusing for the search results to remain there as if nothing has happened.


domw30 (commented):
👍

PR #868, opened 2023-07-31
fix: sign up and log in layout

domw30:
### What does this do?
- fixes the layout to a reasonable standard while developing in sign up / log in flows.

- ensure the pages are still usable and easy to develop on top of.

### How do I test this?
- open the application and run through the sign up / log in flows.

#### Things to consider:
  1. Does this change any APIs?

Screenshot 2023-07-31 at 10 14 10

Screenshot 2023-07-31 at 10 14 32

Screenshot 2023-07-31 at 11 02 13

Screenshot 2023-07-31 at 11 02 24

Screenshot 2023-07-31 at 11 08 07

Screenshot 2023-07-31 at 11 11 15

Screenshot 2023-07-31 at 11 11 22

### DIFF
--- a/src/authentication/create-account-details/index.tsx
+++ b/src/authentication/create-account-details/index.tsx
@@ -1,13 +1,13 @@
 import * as React from 'react';
 
-import { Alert, Button, Input } from '@zero-tech/zui/components';
-
-import './styles.scss';
-import { bem } from '../../lib/bem';
 import { ImageUpload } from '../../components/image-upload';
 import { IconUpload2 } from '@zero-tech/zui/icons';
+import { Alert, Button, Input } from '@zero-tech/zui/components';
 
-const c = bem('create-account-details');
+import { bemClassName } from '../../lib/bem';
+import './styles.scss';
+
+const cn = bemClassName('create-account-details');
 
 export interface Properties {
   isLoading: boolean;
@@ -58,11 +58,13 @@ export class CreateAccountDetails extends React.Component<Properties, State> {
 
   render() {
     return (
-      <div className={c('')}>
-        <h3 className={c('heading')}>CREATE YOUR ACCOUNT</h3>
-        <div className={c('sub-heading')}>Step 2 of 2: What should we call you?</div>
-        <form className={c('form')} onSubmit={this.publishOnCreate}>
-          <div className={c('image-upload')}>
+      <div {...cn('')}>
+        <div {...cn('heading-container')}>
+          <h3 {...cn('heading')}>Create Account</h3>
+          <div {...cn('sub-heading')}>Enter your details</div>
+        </div>
+        <form {...cn('form')} onSubmit={this.publishOnCreate}>
+          <div {...cn('image-upload')}>
             <ImageUpload
               onChange={this.trackImage}
               icon={this.renderImageUploadIcon()}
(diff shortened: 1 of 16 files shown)


src/authentication/footer/footer.tsx
```diff
@@ -15,10 +21,12 @@ export class Footer extends React.Component {
...
-        <div>
-          <span>Dont have an invite code? </span>
-          <a href='https://www.zine.live/#/portal/signup'>Subscribe for updates</a>
-        </div>
+        {this.props.stage === RegistrationStage.ValidateInvite && (
+          <div>
+            <span>Dont have an invite code? </span>
+            <a href='https://www.zine.live/#/portal/signup'>Subscribe for updates</a>
+          </div>
+        )}
```
ratik21, line 29: why not have this in the invite component itself, if we're showing it only there? 🤔

src/invite.tsx
```diff
@@ -45,16 +45,28 @@ export class Container extends React.Component<Properties> {
...
           {this.props.stage === RegistrationStage.SelectMethod && <SelectMethodContainer />}
+
           {this.props.stage === RegistrationStage.EmailAccountCreation && <CreateEmailAccountContainer />}
           {this.props.stage === RegistrationStage.WalletAccountCreation && <CreateWalletAccountContainer />}
+
           {this.props.stage === RegistrationStage.ProfileDetails && <CreateAccountDetailsContainer />}
           {this.props.stage === RegistrationStage.Done && <Redirect to='/' />}
 
-          <Footer />
+          {this.props.stage !== RegistrationStage.ProfileDetails && <Footer stage={this.props.stage} />}
```
ratik21, line 69: a `nit` but i would just return `null` inside the `Footer`, since we're already passing in `this.props.stage`.

ratik21 (approved):
lgtm. added some minor comments

src/authentication/create-account-details/index.tsx
```diff
@@ -58,11 +58,13 @@ export class CreateAccountDetails extends React.Component<Properties, State> {
 
   render() {
     return (
-      <div className={c('')}>
-        <h3 className={c('heading')}>CREATE YOUR ACCOUNT</h3>
-        <div className={c('sub-heading')}>Step 2 of 2: What should we call you?</div>
-        <form className={c('form')} onSubmit={this.publishOnCreate}>
-          <div className={c('image-upload')}>
+      <div {...cn('')}>
+        <div {...cn('heading-container')}>
```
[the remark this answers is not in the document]
domw30: Importing `bemClassName` here and assigning `cn = bemClassName('')`. 

So in this case `{...cn('')}` is basically short for `className={''}`. Removes the need to add `className` attribute - makes things a little cleaner/less cluttered.

src/authentication/footer/footer.tsx
domw30: This area of the footer code will be updated in the follow up screens, this is a simple first iteration of this component 🙂 

Also, think it's better to keep anything footer related within the footer component so its easier to follow - as well as grouping css styles in one container rather than multiple areas.

src/authentication/create-account-details/index.tsx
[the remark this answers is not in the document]
dalefukami: Yeah, it's a little dicey as far as the short names go :) but in my mind `c` is "class" so it creates just the class string where `cn` was "classname" and it creates the classname as an object that we can "spread" onto the element. It saves writing out the `classname=` portion as Dom noted.

src/invite.tsx
dalefukami: If you return null in `Footer` then it will still not be present in the DOM. I.E., Ratik's suggestion would result in the same DOM.

PR #870, opened 2023-07-31
Optimistically render new conversations

dalefukami:
### What does this do?

Optimistically renders a new conversation with stub data

Many things still need to be tweaked but this is a usable starting point:
* Don't allow attempting to send a message until created
* Get the real user name into the local state from the search

To provide instantaneous feedback that something is happening and reduce the feel of lag.

### How do I test this?

Create a conversation with someone you don't have a conversation with yet.


src/store/channels-list/saga.createConversation.test.ts
```diff
@@ -0,0 +1,241 @@
...
+      testSaga(createConversation, ['user-1'], 'name', { name: 'file' } as File)
+        .next()
+        .call(createOptimisticConversation, ['user-1'], 'name', { name: 'file' } as File)
+        .finish();
+    });
+
+    it('sets the optimistic conversation to the active one', async () => {
+      testSaga(createConversation, ['user-1'], 'name', { name: 'file' } as File)
+        .next()
+        // .call(createOptimisticConversation)
```
domw30, line 41: are these comments here for info purposes, sort of like helper comments? so `.next({ id: 'optimistic-id' }))`  is `.call(createOptimisticConversation)`?

domw30 (approved):
Nice 👍

src/store/channels-list/saga.createConversation.test.ts
dalefukami: Yeah, since these are specific unit tests that rely on the exact order of calls these were just to help you know which calls we're expecting but aren't really critical to the test. I'm not sure I like these tests as individual tests so I might just collapse them all down to a single test and then these would go away.

PR #871, opened 2023-07-31
feat: replace select method stage with toggle group component

domw30:
### What does this do?
- removes select method stage from registration saga.
- replaces the stage with toggle group component.

- as per sign-up / log-in designs

### How do I test this?
- navigate to `/get-access` page and follow the sign up flow.

#### Things to consider:
  1. Does this change any APIs?

Removed Select Method Stage - 

https://github.com/zer0-os/zOS/assets/39112648/8662a997-f9c4-47ae-b5cd-0d144c6741eb


src/store/registration/index.ts
```diff
@@ -102,9 +101,11 @@ const slice = createSlice({
     },
     registerWithEmail: (state, _action: PayloadAction<null>) => {
       state.stage = RegistrationStage.EmailAccountCreation;
+      state.errors = [];
```
domw30, line 104: required to clear errors if toggling from creating a web3 account to email or vice versa

src/invite.tsx
```diff
@@ -21,9 +21,20 @@ const cn = bemClassName('invite-main');
...
+  registerWithWallet: () => Action<string>;
+}
+
+export interface State {
+  selectedOption: string;
 }
 
 export class Container extends React.Component<Properties> {
+  state: State = {
+    selectedOption: 'web3',
```
dalefukami, line 35: Not sure if we even need this since we have the Stages that we could key off of? Not worth changing unless it's a 5 minute swap.

PR #873, opened 2023-08-01
refactor: create account method container/presentation-component structure

domw30:
### What does this do?
- refactors the create account method container to follow the container / presentation-component structure that we are using in the authentication folder.
- adds test coverage for the container and component.
- refactors other areas of the code for a tidy up.

- to follow existing practices/conventions.
- to improve the code quality.
- to add test coverage to the container and component.
- to ensure registration state can be accessed for the ToggleGroup component (i.e. is loading - display elements / hide elements).

### How do I test this?
- ensure all tests are passing and functionality for the sign-up flow is working as expected by visiting the `/get-access` page.

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/authentication/create-account-method/container.tsx
+++ b/src/authentication/create-account-method/container.tsx
@@ -29,8 +29,6 @@ export class Container extends React.Component<CreateAccountMethodProps> {
   }
 
   handleSelectionChange = (selectedOption: string) => {
-    this.setState({ selectedOption });
-
     if (selectedOption === 'web3') {
       this.props.registerWithWallet();
     } else if (selectedOption === 'email') {
--- a/src/authentication/create-account-method/container.tsx
+++ b/src/authentication/create-account-method/container.tsx
@@ -1,0 +1,52 @@
+import React from 'react';
+
+import { Action } from 'redux';
+import { RootState } from '../../store/reducer';
+import { connectContainer } from '../../store/redux-container';
+import { RegistrationStage, registerWithEmail, registerWithWallet } from '../../store/registration';
+
+import { CreateAccountMethod } from '.';
+
+export interface CreateAccountMethodProps {
+  stage: RegistrationStage;
+  isConnecting: boolean;
+
+  registerWithEmail: () => Action<string>;
+  registerWithWallet: () => Action<string>;
+}
+
+export class Container extends React.Component<CreateAccountMethodProps> {
+  static mapState(state: RootState): Partial<CreateAccountMethodProps> {
+    const { registration } = state;
+    return {
+      stage: registration.stage,
+      isConnecting: registration.loading,
+    };
+  }
+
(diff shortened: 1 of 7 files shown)


src/authentication/create-account-method/container.tsx
```diff
@@ -0,0 +1,52 @@
...
+      isConnecting: registration.loading,
+    };
+  }
+
+  static mapActions(_props: CreateAccountMethodProps): Partial<CreateAccountMethodProps> {
+    return { registerWithEmail: registerWithEmail, registerWithWallet: registerWithWallet };
+  }
+
+  handleSelectionChange = (selectedOption: string) => {
+    this.setState({ selectedOption });
```
dalefukami, line 32: :thinking: I'm surprised there isn't a typescript error here as we didn't define the state?

Actually... do we use it?

src/authentication/create-account-method/index.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+    const wrapper = subject({ onSelectionChange });
+
+    wrapper.find('ToggleGroup').simulate('selectionChange', 'web3');
+    expect(onSelectionChange).toHaveBeenCalledWith('web3');
+  });
+
+  it('renders CreateEmailAccountContainer when stage is EmailAccountCreation', () => {
+    const wrapper = subject({ stage: RegistrationStage.EmailAccountCreation });
+    expect(wrapper.find(CreateEmailAccountContainer)).toHaveLength(1);
+    expect(wrapper.find(CreateWalletAccountContainer)).toHaveLength(0);
```
dalefukami, line 36: There's a helper assertion that could work here:
```
expect(wrapper).toHaveElement(CreateEmailAccountContainer);
expect(wrapper).not.toHaveElement(CreateWalletAccountContainer)
```
I find those express the intent a little better in these situations.

dalefukami (approved):
Nice, well done.

src/authentication/create-account-method/index.tsx
```diff
@@ -1,8 +1,10 @@
-import React from 'react';
-import { ToggleGroup } from '@zero-tech/zui/components';
+import * as React from 'react';
```
ratik21, line 1: to confirm, is this intentional change? Everywhere i see `import React from 'react';` being used

ratik21 (approved):

src/authentication/create-account-method/container.tsx
domw30: good catch thanks, it was not being used!

src/authentication/create-account-method/index.tsx
domw30: Yeah I would have changed this just so I didn't have to explicitly import `FC` for `React.FC` on line 20. 

I probably should have used `React.Component` for a class component rather than functional to stay in line with everything else. My favoured Functional Component approach has crept through here haha. I'll update at another time.

PR #879, opened 2023-08-02
feat: update wallet select for sign-up / log-in flow 

domw30:
### What does this do?
- some minor updates to the update wallet select component

- as per figma design

### How do I test this?
- run through the sign-up/log-in flow.

#### Things to consider:
  1. Does this change any APIs?

BEFORE:
Screenshot 2023-08-02 at 11 16 29

AFTER
Screenshot 2023-08-02 at 11 15 54

The Wallet Select Modal when in app (not messenger) will remain the same due to adding the `walletSelectTitle` prop:
Screenshot 2023-08-02 at 11 24 40

### DIFF
--- a/src/authentication/create-wallet-account/styles.scss
+++ b/src/authentication/create-wallet-account/styles.scss
@@ -13,10 +13,6 @@
   }
 
   &__select-wallet {
-    background-color: theme.$color-primary-2;
-    border: 1px solid theme.$color-primary-5;
-    border-radius: 8px;
-    padding: 16px;
     margin-bottom: 8px;
   }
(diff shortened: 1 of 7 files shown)


domw30:
To Note- 

Figma design:
Screenshot 2023-08-02 at 11 37 05

Current implementation:
Screenshot 2023-08-02 at 11 37 54

The styles here differ slightly, re the background colour and the follow up hover effect colours for this. I have started a discussion with design to understand what colours to use. Once I receive confirmation I will update accordingly.

dalefukami (approved):

PR #882, opened 2023-08-02
feat: name and avatar screen improvements

domw30:
### What does this do?
- updates the ui of the name and avatar screen

- as per figma design

### How do I test this?
- navigate to the sign-up flow. sign up via one of the create account methods. when on the name and avatar screen, compare with [figma](https://www.figma.com/file/aETpyuG2AKjNcQtCjldcX0/ZERO-Messenger-%2F-Web?type=design&node-id=7794-272813&mode=dev) design

#### Things to consider:
  1. Does this change any APIs?

Final UI:

Screenshot 2023-08-02 at 13 42 21

Screenshot 2023-08-02 at 13 44 04

### DIFF
--- a/src/authentication/create-account-details/index.test.tsx
+++ b/src/authentication/create-account-details/index.test.tsx
@@ -63,18 +63,4 @@ describe('CreateAccountDetails', () => {
 
     expect(wrapper.find('Alert').prop('children')).toEqual('invalid');
   });
-
-  it('disables button when name is empty', function () {
-    const wrapper = subject({});
-
-    expect(wrapper.state('name')).toEqual('');
-    expect(wrapper.find('Button').prop('isDisabled')).toEqual(true);
-  });
-
-  it('enables button when name is not empty', function () {
-    const wrapper = subject({});
-
-    wrapper.find('Input[name="name"]').simulate('change', inputEvent('Jack Black'));
-    expect(wrapper.find('Button').prop('isDisabled')).toEqual(false);
-  });
 });
--- a/src/authentication/create-account-details/index.test.tsx
+++ b/src/authentication/create-account-details/index.test.tsx
@@ -63,4 +63,18 @@ describe('CreateAccountDetails', () => {
 
     expect(wrapper.find('Alert').prop('children')).toEqual('invalid');
   });
+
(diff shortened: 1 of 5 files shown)


domw30:
Please note: background color of image upload component differs in the figma design to current implementation. This is another question I have raised with design as to what the color in the design actually is - unable to find in figma. Shall update once I have received confirmation of color.

src/authentication/create-account-details/index.tsx
```diff
@@ -68,24 +68,26 @@ export class CreateAccountDetails extends React.Component<Properties, State> {
...
               errorMessage={this.props.errors.image}
             />
           </div>
           {this.imageError && <Alert variant='error'>{this.imageError}</Alert>}
           <Input
             {...cn('input')}
-            label='What is your name?'
-            helperText='This will be your name that is visible to others on Zero'
+            label='Display Name'
+            wrapperClassName='create-account-details__input-wrapper'
```
domw30, line 79: not sure how best to write the classNames here when using `bemClassName`, so I have added the full className for now.

dalefukami: Yeah, they're just tools that make the general 99% case shorter. If you wanted you could add:
```
const c = bem('create-account-details');
```
at the top of the file alongside the `cn` definition. Then this becomes:
`wrapperClassName={c('input-wrapper')}`

I.E., `bem` and `bemClassName` aren't mutually exclusive and if they're defined close to each other then the duplication is localized.

src/authentication/create-account-details/index.test.tsx
```diff
@@ -63,4 +63,18 @@ describe('CreateAccountDetails', () => {
 
     expect(wrapper.find('Alert').prop('children')).toEqual('invalid');
   });
+
+  it('disables button when name is empty', function () {
```
dalefukami, line 67: Are we sure we want this? In previous discussions with Ethan we decided to always keep the button enabled and the errors would tell the user what's required. The problem now is that there's nothing telling the user which fields are required and why the button is disabled.

dalefukami (approved):
Approving but deal with the disabled button.

PR #883, opened 2023-08-02
feat: login with wallet error state and connecting state updates

domw30:
### What does this do?
- adjusts the position of the login with wallet container when connecting wallet via the login with wallet option (as shown in the demo video below) - this position should only change for login with wallet option, not the login with email option, hence the condition added.
- updates the style and content of the error states as per figma design

- as per figma designs.

### How do I test this?
- navigate to `/login` page and run through the create wallet account flow. Check against the [figma](https://www.figma.com/file/aETpyuG2AKjNcQtCjldcX0/ZERO-Messenger-%2F-Web?type=design&node-id=9777-315854&mode=dev) designs

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/authentication/web3-login/index.tsx
+++ b/src/authentication/web3-login/index.tsx
@@ -1,13 +1,13 @@
 import * as React from 'react';
 
-import { Alert, Button } from '@zero-tech/zui/components';
-import { IconLoading2 } from '@zero-tech/zui/icons';
+import { Alert } from '@zero-tech/zui/components';
 import { WalletSelect } from '../../components/wallet-select';
 
+import { bemClassName } from '../../lib/bem';
 import './styles.scss';
-import { bem } from '../../lib/bem';
+import { Web3LoginErrors } from '../../store/login';
 
-const c = bem('web3-login');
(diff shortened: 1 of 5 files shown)


src/authentication/web3-login/index.tsx
```diff
@@ -21,22 +21,25 @@ export class Web3Login extends React.Component<Web3LoginProperties, Web3LoginSta
...
+        <div {...cn('login')}>
+          <div {...cn('select-wallet')}>
+            <WalletSelect isConnecting={isConnecting} onSelect={onSelect} />
           </div>
-        ) : (
-          <div className={c('login')}>
-            <div className={c('select-wallet')}>
-              <WalletSelect isConnecting={false} onSelect={onSelect} />
+          {error && (
+            <div {...cn('error-container')}>
```
domw30, line 36: Looks to me like the error styles for this flow differ to other alert error styles in the app i.e. font-size and line-height spacing. Other uses of Alert don't require the container for the padding here either. Will note to look at zUI and will ping design another question related to this.

PR #890, opened 2023-08-03
Render error if the conversation fails to create

dalefukami:
### What does this do?

Renders an error if an optimistically rendered conversation fails to create

User feedback

image

### DIFF
--- a/src/components/chat-view-container/chat-view.tsx
+++ b/src/components/chat-view-container/chat-view.tsx
@@ -23,6 +23,11 @@ import './styles.scss';
 import { ChatSkeleton } from './chat-skeleton';
 import { createMessageGroups, getMessageRenderProps } from './utils';
 import { MessagesFetchState } from '../../store/channels';
+import { bemClassName } from '../../lib/bem';
+
+// Note: this is the component convention. Existing styles reference channel-view which
+// is old and can be migrated to this component.
+const cn = bemClassName('chat-view');
 
 interface ChatMessageGroups {
   [date: string]: MessageModel[];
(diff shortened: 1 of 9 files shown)


src/components/chat-view-container/styles.scss
```diff
@@ -3,6 +3,18 @@
 
 // Note: there are some classes in various other places (src/platform-apps/channels/styles.scss)
 
+.chat-view {
+  &__error {
+    padding: 8px;
+    padding-bottom: 32px;
```
domw30, line 9: does this `padding-bottom` get overridden here by the `padding` above?

src/components/chat-view-container/chat-view.tsx
```diff
@@ -23,6 +23,11 @@ import './styles.scss';
 import { ChatSkeleton } from './chat-skeleton';
 import { createMessageGroups, getMessageRenderProps } from './utils';
 import { MessagesFetchState } from '../../store/channels';
+import { bemClassName } from '../../lib/bem';
+
+// Note: this is the component convention. Existing styles reference channel-view which
+// is old and can be migrated to this component.
```
domw30, line 29: nice 👍

domw30 (commented):
looks good

src/components/chat-view-container/styles.scss
dalefukami: Because it's second the 32 overrides the top one. It's effectively the same as:
```
padding: 8px 8px 32px 8px;
```
but I find it more clear when reading. I'm not so sure if it makes sense though. Thought I'd play with it and see how it felt.

PR #891, opened 2023-08-03
Don't replace optimistic messages when conversation poll happens

dalefukami:
### What does this do?

Ensure we don't remove optimistic conversations from the list when the automatic poll happens

To make sure the UI doesn't rip an "in progress" conversation out of the UI for the user

### How do I test this?

Force a conversation creation to take a long time (longer than 1 minute). Verify that when the background poll happens the conversation does not disappear.


domw30 (commented):
👍

PR #896, opened 2023-08-04
Fix blank screen issue in case of an incomplete user

ratik21:


dalefukami (approved):

dalefukami:
It does feel like maybe something is slightly out of place that we're touching the "pageLoad complete" flag somewhere other than the page load.

PR #897, opened 2023-08-04
Fix: search results no longer returning all group conversations containing members matching search input

ratik21:
Now after searching:

(returns both direct + group match)
image

(direct matches are prioritized first, and then group conversation matches by otherMember names)
image

image


domw30 (approved):
Looks good to me 🚀

PR #898, opened 2023-08-04
Zos 469: Adjust preview text behavior in conversations list sidebar when files images are received

ratik21:
Now:
image
image
image


dalefukami (commented):
Are we sure we literally wanted "they"?... doesn't it make more sense to put the user name like other messages?

image

ratik21:
> Are we sure we literally wanted "they"?... doesn't it make more sense to put the user name like other messages?

🤦‍♂️ my bad. Updated.

image

image

dalefukami (approved):

PR #899, opened 2023-08-04
feat: add fade-in animation to invite code module

domw30:
### What does this do?
- checks if data is loading to render elements at the same time. Prior to this the elements were rendering at different times making the elements fade-in at different times.
- fades in the elements as per figma.
- this also reverts the colour of the invites left logic

- as per figma design. This is to soften the rendering of the the elements

### How do I test this?
- once logged in click the invite button to bring up the invite module.

#### Things to consider:
  1. Does this change any APIs?

BEFORE:

https://github.com/zer0-os/zOS/assets/39112648/52a78d85-8c64-4efe-afa2-e3039ad7ae15

AFTER:

https://github.com/zer0-os/zOS/assets/39112648/8a7cad06-fefc-488f-85ae-f56d2b5a62fa

### DIFF
--- a/src/components/invite-dialog/index.test.tsx
+++ b/src/components/invite-dialog/index.test.tsx
@@ -14,6 +14,7 @@ describe('InviteDialog', () => {
       maxUses: 0,
       isUserAMemberOfWorlds: false,
       clipboard: { write: () => null },
+      isLoading: false,
       ...props,
     };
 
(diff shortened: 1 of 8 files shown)


src/components/invite-dialog/index.test.tsx
```diff
@@ -85,14 +91,14 @@ describe('InviteDialog', () => {
     expect(onClose).toHaveBeenCalled();
   });
 
-  it('displays text in green if no invites remaining', function () {
+  it('displays text in green if invites remaining', function () {
```
dalefukami, line 94: We should wait for Ethan to confirm this is what he wants. The designs have the text green when the invites remaining hits zero.

dalefukami (approved):

src/components/invite-dialog/index.test.tsx
domw30: Will do :) thanks!

PR #901, opened 2023-08-07
fix: edit profile minor ui tweaks

domw30:
### What does this do?
- some very minor ui tweaks

- noticed some position and spacing inconsistencies 

### How do I test this?
- click profile avatar and select `edit profile`

#### Things to consider:
  1. Does this change any APIs?
  
 
BEFORE:

Screenshot 2023-08-07 at 12 28 11

AFTER:

Screenshot 2023-08-07 at 12 28 30

### DIFF
--- a/src/components/edit-profile/styles.scss
+++ b/src/components/edit-profile/styles.scss
@@ -29,7 +29,11 @@
   }
 
   &__close {
-    height: 28px;
+    display: flex;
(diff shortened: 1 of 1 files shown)


ratik21 (approved):
Thanks! This was on my mind. We also need to fix the display for "alerts" (general error + input errors). I can do that :)

PR #902, opened 2023-08-07
fix: sign up and log in flow feedback 

domw30:
### What does this do?
- removes wallet icon backgrounds
- update toggle group logic (show when creating account via email)
- updates alert messages
- adds option to log in footer
- updates success variant for password validity
- some ui tweaks

- as per product QA feedback/figma designs.

### DIFF
--- a/src/authentication/create-account-method/index.tsx
+++ b/src/authentication/create-account-method/index.tsx
@@ -23,14 +23,15 @@ export const CreateAccountMethod: React.FC<CreateAccountMethodProps> = ({ stage,
     { key: 'email', label: 'Email' },
   ];
 
-  const selectedOption = stage === RegistrationStage.WalletAccountCreation ? 'web3' : 'email';
   const isWalletAccountCreationStage = stage === RegistrationStage.WalletAccountCreation;
+  const selectedOption = isWalletAccountCreationStage ? 'web3' : 'email';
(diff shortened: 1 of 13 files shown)


src/authentication/create-account-method/index.tsx
```diff
@@ -23,14 +23,15 @@ export const CreateAccountMethod: React.FC<CreateAccountMethodProps> = ({ stage,
...
   const isWalletAccountCreationStage = stage === RegistrationStage.WalletAccountCreation;
+  const selectedOption = isWalletAccountCreationStage ? 'web3' : 'email';
+  const shouldRenderToggleGroup = isConnecting && isWalletAccountCreationStage;
 
   return (
     <div {...cn('', isConnecting && isWalletAccountCreationStage && 'is-connecting')}>
       <h3 {...cn('heading')}>Create Account</h3>
 
-      {!isConnecting && (
+      {!shouldRenderToggleGroup && (
```
dalefukami, line 34: This is confusing to me. "If not should render...then render"?

dalefukami (approved):

ratik21 (approved):

src/authentication/create-account-method/index.tsx
domw30: Oops sorry yeah that does not make any sense. Updated.

PR #909, opened 2023-08-08
Integrate matrix

jtulloch:
### What does this do?
Adds initial support for running zOS against a matrix homeserver.

To begin the move towards a distributed chat infrastructure.

### How do I test this?
Run or get access to a matrix homeserver. Create a user and get the credentials. Set them in .env. Fire it up.

#### Things to consider:
  1. Does this change any APIs?


src/lib/chat/matrix-client.ts
```diff
@@ -0,0 +1,161 @@
...
+    return this.connectionAwaiter;
+  };
+
+  private addConnectionAwaiter() {
+    this.connectionAwaiter = new Promise((resolve) => {
+      this.connectionResolver = resolve;
+    });
+  }
+
+  private clearConnectionAwaiter() {
```
dalefukami, line 132: This might be unused. Not sure how important it is.

dalefukami (approved):

src/lib/chat/matrix-client.ts
jtulloch: hmm. good point. i don't think it's too important as things stand. it'll maybe come up if we start re-connecting, but i'll just kill it for now.

PR #912, opened 2023-08-08
Add tests around toLocalChannel

dalefukami:
### What does this do?

Adds tests around the existing behaviour for `toLocalChannel`

### DIFF
--- a/src/store/channels-list/utils.ts
+++ b/src/store/channels-list/utils.ts
@@ -28,8 +28,8 @@ export const toLocalChannel = (input): Partial<Channel> => {
     otherMembers: input.otherMembers || [],
     lastMessage: input.lastMessage || null,
     lastMessageCreatedAt: input.lastMessage?.createdAt || null,
-    isChannel: input.isChannel === undefined ? true : input.isChannel,
-    groupChannelType: input.groupChannelType || '',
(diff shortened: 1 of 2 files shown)


src/store/channels-list/utils.ts
```diff
@@ -28,8 +28,8 @@ export const toLocalChannel = (input): Partial<Channel> => {
     otherMembers: input.otherMembers || [],
     lastMessage: input.lastMessage || null,
     lastMessageCreatedAt: input.lastMessage?.createdAt || null,
-    isChannel: input.isChannel === undefined ? true : input.isChannel,
-    groupChannelType: input.groupChannelType || '',
+    isChannel: (input.isChannel ?? true) || !!input.isChannel,
+    groupChannelType: input.groupChannelType || '', // Is this the right default to use?
```
domw30, line 32: Hm shouldn't it be 'private' or 'public' as default?

domw30 (approved):

ratik21 (approved):

src/store/channels-list/utils.ts
dalefukami: Yep, it should but it's been like this since someone wrote it so I didn't want to mix functionality changes with refactoring.

PR #913, opened 2023-08-09
refactor: improve messenger list file structure and style

domw30:
### What does this do?
- use bemClassName and remove className attributes in messenger list files.
- one minor fix spotted whilst refactoring - selected users position/alignment.
- split each component into its own directory.
- extracted all styles from a massive single scss file into isolated component scss files in their own directory folders. 
- tidied up all imports and reduced duplication.
- manually tested local changes against development to prevent any breaks slipping through.

- remain consistent with style guide.
- this makes it 100x easier to navigate through the messenger list component styles.

### How do I test this?
Comparison with development:
Screenshot 2023-08-09 at 12 03 00

Screenshot 2023-08-09 at 12 02 11

Screenshot 2023-08-09 at 12 02 54

Screenshot 2023-08-09 at 12 02 36

#### Things to consider:
  1. Does this change any APIs?


ratik21 (approved):

domw30:
Thanks for the review @ratik21. I'll see what @dalefukami thinks as well before hitting the green button :)

src/components/messenger/list/conversation-item/conversation-item.tsx
```diff
@@ -1,19 +1,23 @@
 import * as React from 'react';
```
dalefukami, line 1: We should probably stick to conventions and name this file `index.tsx`. Same for all the other ones.

dalefukami (approved):

src/components/messenger/list/conversation-item/conversation-item.tsx
domw30: Okay. What would you name the `index.ts` file if renaming `conversation-item.tsx` to `index.tsx` ?

dalefukami: I wouldn't have one. None of our other components have one.

domw30: Okay no probs, the idea was an attempt to improve imports like 'import { menu } from './menu/menu`. I'll change those back. Thanks!

dalefukami: Yeah, the issue is that doesn't follow any of the preexisting conventions and it's a pretty significant change to the file structure. If we want to propose new conventions those should come in separate PRs where the merits of the new proposal can be discussed outside the distraction of other changes.

domw30: Nice yeah agree, doesn't make sense to have one component structured differently to the rest, understood that stirs up the existing conventions. Made those changes and checked everything is running okay :)

dalefukami: :+1: 

Love the PR by the way. Thanks for cleaning up!

PR #919, opened 2023-08-11
feat: add reset password store

domw30:
### What does this do?
- adds reset-password saga
- adds reset-password saga tests
- adds reset-password api (this is likely to change - away from auto 0)
- adds reset-password actions and reducer
- updates store root reducer
- updates store root saga

- to introduce and wire up the reset password flow 

### How do I test this?
- run tests
- main testing once wired up with frontend container and component.

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/store/login/index.ts
+++ b/src/store/login/index.ts
@@ -5,6 +5,7 @@ export enum SagaActionTypes {
   EmailLogin = 'login/emailLogin',
   Web3Login = 'login/web3Login',
   SwitchLoginStage = 'login/switchLoginStage',
+  ResetPassword = 'reset-password',
 }
 
 export type LoginState = {
--- a/src/store/login/index.ts
+++ b/src/store/login/index.ts
@@ -16,7 +16,7 @@ export type LoginState = {
 export enum LoginStage {
   EmailLogin = 'email',
   Web3Login = 'web3',
-  ResetPassword = 'reset-password',
+  RequestPasswordReset = 'request-password-reset',
   Done = 'done',
 }
 
--- a/src/store/login/index.ts
+++ b/src/store/login/index.ts
@@ -5,7 +5,6 @@ export enum SagaActionTypes {
   EmailLogin = 'login/emailLogin',
   Web3Login = 'login/web3Login',
   SwitchLoginStage = 'login/switchLoginStage',
-  ResetPassword = 'reset-password',
 }
 
 export type LoginState = {
@@ -17,6 +16,7 @@ export type LoginState = {
 export enum LoginStage {
   EmailLogin = 'email',
   Web3Login = 'web3',
+  ResetPassword = 'reset-password',
   Done = 'done',
 }
 
--- a/src/store/request-password-reset/index.ts
+++ b/src/store/request-password-reset/index.ts
@@ -2,8 +2,8 @@ import { PayloadAction, createAction, createSlice } from '@reduxjs/toolkit';
 
 export enum SagaActionTypes {
   RequestPasswordReset = 'requestPasswordReset',
-  EnterRequesetPasswordResetPage = 'enterRequesetPasswordResetPage',
-  LeaveRequesetPasswordResetPage = 'leaveRequesetPasswordResetPage',
+  EnterRequestPasswordResetPage = 'enterRequestPasswordResetPage',
+  LeaveRequestPasswordResetPage = 'leaveRequestPasswordResetPage',
 }
 
 export enum RequestPasswordResetStage {
@@ -30,8 +30,8 @@ export const initialState: RequestPasswordResetState = {
 };
 
 export const requestPasswordReset = createAction<{ email: string }>(SagaActionTypes.RequestPasswordReset);
-export const enterResetPasswordPage = createAction(SagaActionTypes.EnterRequesetPasswordResetPage);
-export const leaveResetPasswordPage = createAction(SagaActionTypes.LeaveRequesetPasswordResetPage);
+export const enterResetPasswordPage = createAction(SagaActionTypes.EnterRequestPasswordResetPage);
+export const leaveResetPasswordPage = createAction(SagaActionTypes.LeaveRequestPasswordResetPage);
 
 const slice = createSlice({
   name: 'request-password-reset',
--- a/src/store/request-password-reset/index.ts
+++ b/src/store/request-password-reset/index.ts
@@ -1,10 +1,10 @@
 import { PayloadAction, createAction, createSlice } from '@reduxjs/toolkit';
 
 export enum SagaActionTypes {
-  ResetPassword = 'resetPassword',
+  RequestPasswordReset = 'requestPasswordReset',
 }
 
-export type ResetState = {
+export type RequestPasswordResetState = {
   loading: boolean;
   errors: string[];
   emailSubmitted: boolean;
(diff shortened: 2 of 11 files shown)


src/store/reset-password/saga.ts
```diff
@@ -0,0 +1,52 @@
+import { resetPassword as resetPasswordApi } from './api';
+import { setLoading, setErrors, SagaActionTypes, setEmailSubmitted } from './index';
+import { ResetPasswordErrors } from '.';
+import { call, put, take } from 'redux-saga/effects';
+
+export function validateResetPassword({ email }) {
```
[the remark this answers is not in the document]
dalefukami: Ah, I see now that this isn't actually the reset password. It's just the request for the link.

src/store/login/index.ts
```diff
@@ -16,6 +16,7 @@ export type LoginState = {
 export enum LoginStage {
   EmailLogin = 'email',
   Web3Login = 'web3',
+  ResetPassword = 'reset-password',
```
dalefukami, line 19: I think we're going to run into a fair bit of confusion between requesting the password reset email and actually resetting the password. Maybe "RequestPasswordReset" for these various things? (can be handled in follow ups)

src/store/reset-password/saga.ts
```diff
@@ -0,0 +1,52 @@
...
+      }
+      yield put(setErrors(errors));
+    }
+
+    yield put(setLoading(false));
+  } while (!success);
+}
+
+export function* saga() {
+  yield resetPasswordPage();
```
dalefukami, line 51: I'm not sure this pattern is quite right. This seems like you'd only be able to reach the reset password stage a single time. What happens if I request a password reset and then click the `login` link on the footer and then click the link to try to come back and reset it again? My guess is the user would be stuck.

We also don't really want to be waiting for password reset events if you're actually logged into the app. It's cleaner to only be listening for these events when we expect them to happen (i.e., when we've gotten to that page)

src/store/reset-password/saga.test.ts
```diff
@@ -0,0 +1,75 @@
+import { expectSaga } from 'redux-saga-test-plan';
+import { call } from 'redux-saga/effects';
+import { resetPasswordPage } from './saga';
+import { resetPassword as resetPasswordApi } from './api';
+import { setLoading, setErrors, setEmailSubmitted, SagaActionTypes } from './index';
+import { ResetPasswordErrors } from '.';
+
+describe('resetPasswordPage', () => {
```
dalefukami, line 8: I have some notes about these tests but I think I'm going to record a video to try to explain/show what I mean.

src/store/reset-password/saga.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    const validationErrors = validateResetPassword(action.payload);
+
+    if (validationErrors.length > 0) {
+      yield put(setErrors(validationErrors));
+      success = false;
+      continue;
+    }
+
+    const result = yield call(resetPasswordApi, action.payload);
```
dalefukami, line 31: I think we need a try/catch somewhere in here because it's possible for the api request to outright fail and that would throw an exception.

dalefukami (commented):
Looks like a great start! Some things we need to do to make it a little more robust but overall we're on the right track.

src/store/login/index.ts
domw30: Yeah agreed this makes more sense! Updated :)

src/store/reset-password/index.ts
```diff
@@ -0,0 +1,44 @@
+import { PayloadAction, createAction, createSlice } from '@reduxjs/toolkit';
+
+export enum SagaActionTypes {
+  ResetPassword = 'resetPassword',
+}
+
+export type ResetState = {
+  loading: boolean;
+  errors: string[];
+  emailSubmitted: boolean;
```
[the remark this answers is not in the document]
domw30: Added a stage attribute to handle this. Thanks Dale!

src/store/reset-password/index.ts
```diff
@@ -0,0 +1,44 @@
...
+};
+
+export enum ResetPasswordErrors {
+  UNKNOWN_ERROR = 'UNKNOWN',
+  EMAIL_REQUIRED = 'EMAIL_REQUIRED',
+  EMAIL_NOT_FOUND = 'EMAIL_NOT_FOUND',
```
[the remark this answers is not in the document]
domw30: Removed `EMAIL_NOT_FOUND`. We'll only show a general error is something fails, but not indicate to the user that the email was not found. Email required will just handle an empty input. Thanks!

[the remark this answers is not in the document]
ratik21: @dalefukami @domw30 shouldn't we follow up on this for the `login` component as well? I think we also send back a "email not found" response there (if email didn't exist while logging in)

src/store/request-password-reset/index.ts
```diff
@@ -0,0 +1,53 @@
...
+  LeaveRequestPasswordResetPage = 'leaveRequestPasswordResetPage',
+}
+
+export enum RequestPasswordResetStage {
+  SubmitEmail = 'submit-email',
+  Done = 'done',
+}
+
+export type RequestPasswordResetState = {
+  loading: boolean;
```
dalefukami, line 15: Do we need the loading flag if we have a stage now?

src/store/reset-password/index.ts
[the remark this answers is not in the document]
dalefukami: Do we still need the loading flag? The idea was to replace both flags with the stage.

src/store/reset-password/index.ts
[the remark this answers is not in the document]
dalefukami: Yep, we probably should

src/store/request-password-reset/index.ts
```diff
@@ -0,0 +1,53 @@
...
+  stage: RequestPasswordResetStage;
+  errors: string[];
+};
+
+export enum ResetPasswordErrors {
+  UNKNOWN_ERROR = 'UNKNOWN',
+  API_ERROR = 'API_ERROR',
```
dalefukami, line 22: Does this error hold any meaning? I.E., are we reacting differently anywhere in the app? If not, we don't really need it. It's also a little misleading because the api returns specific errors and this is being used in the case where it's likely something other than the api failed.

src/store/reset-password/index.ts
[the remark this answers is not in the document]
domw30: My thinking was we would need loading to handle loading state of the submit button? I could be missing something though 🤔

src/store/request-password-reset/saga.ts
```diff
@@ -0,0 +1,59 @@
...
+
+      if (result.success) {
+        yield put(setStage(RequestPasswordResetStage.Done));
+        break;
+      } else {
+        let errors = [];
+        if (result.response === 'UNKNOWN_ERROR') {
+          errors.push(ResetPasswordErrors.UNKNOWN_ERROR);
+        }
+        yield put(setErrors(errors));
```
dalefukami, line 37: This seems off. What if the response sends a different type of error....then we'd end up with an empty error list.

```suggestion
        yield put(setErrors([ResetPasswordErrors.UNKNOWN_ERROR]));
```

src/store/request-password-reset/saga.ts
```diff
@@ -0,0 +1,59 @@
...
+      }
+    } catch (error: any) {
+      console.error(error);
```
dalefukami, line 40: Remove

dalefukami (approved):
Comments can be handled in follow ups.

PR #920, opened 2023-08-11
feat: add reset password container and component

domw30:
### What does this do?
- adds reset password container
- adds reset password container tests
- adds reset password component
- adds reset password component tests
- adds reset password styling
- wires up use of reset password api, saga, actions and reducer

- required to use reset password store in ui

### How do I test this?
- run tests
- to view locally, you can switch out one of the containers in `invite.tsx` for the reset password container and `npm run start`.

#### Things to consider:
  1. Does this change any APIs?

Screenshot 2023-08-11 at 09 30 01

Screenshot 2023-08-11 at 09 30 07

Screenshot 2023-08-11 at 09 30 28


src/authentication/reset-password/container.tsx
```diff
@@ -0,0 +1,70 @@
...
+      switch (error) {
+        case ResetPasswordErrors.EMAIL_NOT_FOUND:
+          errorObject.email = 'Email not found';
+          break;
+        case ResetPasswordErrors.EMAIL_REQUIRED:
+          errorObject.email = 'Email is required';
+          break;
+        case ResetPasswordErrors.UNKNOWN_ERROR:
+          errorObject.general = 'An unknown error has occurred';
+          break;
```
ratik21, line 45: maybe we can remove this case altogether, since it will go to default anyways?

src/authentication/reset-password/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+import { Button, Input, Alert } from '@zero-tech/zui/components';
+
+import classNames from 'classnames';
+import { bem, bemClassName } from '../../lib/bem';
+
+import './styles.scss';
+
+const c = bem('reset-password');
+const cn = bemClassName('reset-password');
```
ratik21, line 11: using both `c` & `cn` gets a little confusing. Is it possibe to use only one?

ratik21 (approved):

src/authentication/reset-password/container.tsx
domw30: Agree! Updated.

src/authentication/reset-password/index.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  it('renders general errors', function () {
+    const wrapper = subject({ errors: { general: 'invalid' } });
+
+    expect(wrapper.find('Alert').prop('children')).toEqual('invalid');
+  });
+
+  it('renders success message when email submitted', function () {
+    const wrapper = subject({ emailSubmitted: true });
+    wrapper.setState({ email: 'test@example.com' });
```
dalefukami, line 53: Would be better to simulate the actual user experience rather than force internal state. So, change the input, and then `setProps` to be emailSubmitted.

src/authentication/reset-password/index.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+  it('disables the submit button if isLoading or email is empty', function () {
+    let wrapper = subject({ isLoading: true });
+    expect(wrapper.find('Button').prop('isDisabled')).toEqual(true);
+
+    wrapper = subject({});
+    wrapper.setState({ email: '' });
+    expect(wrapper.find('Button').prop('isDisabled')).toEqual(true);
+  });
+
+  it('tracks email input correctly', function () {
```
dalefukami, line 68: I'm not sure this test is useful. Nothing really cares what the internal state of the component is. If this is important, how would the user notice the impact?

src/authentication/reset-password/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+        An email containing a reset password link has been emailed to: {this.state.email}
+      </div>
+    );
+  }
+
+  renderForm() {
+    return (
+      <form {...cn('form')} onSubmit={this.handleResetPassword}>
+        <Input
+          {...cn('input')}
```
dalefukami, line 66: I really don't like how we have to override the styles for all of these components. Are we sure these are all necessary? `input`, `submit-button`, `error`.... seems like something's wrong if we are constantly forceably overriding our UI lib styles.

dalefukami (approved):

PR #921, opened 2023-08-11
feat: update log in container and component to handle reset-password

domw30:
### What does this do?
- updates log in container to handle reset password selection
- updates log in container to handle switch back to login selection and resets email submission
- updates log in container tests to cover the above
- updates log in component to handle ui for reset password switch and footer
- updates log in component tests
- updates log in styling

- to be able to switch the ui to reset password flow from log in screens
- wires up the reset password container and password

### How do I test this?
- run ui and navigate to email log in screen. click the `reset` button in the footer to navigate to the reset password flow.

#### Things to consider:
  1. Does this change any APIs?

https://github.com/zer0-os/zOS/assets/39112648/189004fb-31a9-456f-8bb7-2873efffb93f


src/pages/login/login-component.tsx
```diff
@@ -56,39 +59,62 @@ export class LoginComponent extends React.Component<LoginComponentProperties> {
...
     const isWeb3LoginStage = stage === LoginStage.Web3Login;
+    const isResetPasswordStage = stage === LoginStage.ResetPassword;
     const selectedOption = isWeb3LoginStage ? 'web3' : 'email';
 
     return (
       <>
         <ThemeEngine theme={Themes.Dark} />
         <div {...cn('')}>
-          <main {...cn('content')}>
+          <main {...cn('content', isResetPasswordStage && 'is-reset-stage')}>
```
ratik21, line 102:
```suggestion
          <main {...cn('content', isResetPasswordStage && 'reset-stage')}>
```
I don't think the classname shoud also be `is-reset-stage`

src/pages/login/login-container.tsx
```diff
@@ -28,12 +30,29 @@ export class LoginContainer extends React.Component<LoginContainerProperties> {
...
+        break;
+      case 'email':
+        switchLoginStage(LoginStage.EmailLogin);
+        break;
+      case 'reset':
+        switchLoginStage(LoginStage.ResetPassword);
+        break;
+      case 'backToLogin':
+        switchLoginStage(LoginStage.EmailLogin);
+        this.props.setEmailSubmitted(false);
```
ratik21, line 51: I don't follow this part. Could you please explain?

src/pages/login/login-component.tsx
domw30: Is it possible to expand on why? Happy to change, but I'm not sure `reset-stage` reflects the condition in the styles as well as using `is`.  I think it mirrors the condition `isResetPasswordStage` clearly, but again, happy to change if that is misleading.

src/pages/login/login-container.tsx
domw30: Yeah sure! 

In `src/pages/login/login-component.tsx`, line:

```
<button {...cn('button-link')} onClick={() => this.props.handleSelectionChange('backToLogin')}>
```

If the button is clicked, we set the `selectedOption` to `backToLogin`. So in the switch statement above, if the case is equal to `backToLogin` then two things happen:
1. The login stage is set to `EmailLogin` this means the UI will display the log in with email screen. 
2. The state for 'has an email been submitted' is set back to its initial state `false`. This means that if the user navigates back to the reset screen, they will not see the success message, but instead the form to enter their email.

Let me know if that's not clear !

PR #922, opened 2023-08-14
fix: elements displayed in block body

domw30:
### What does this do?
- fixes a current issue where elements displayed in the message block body are forced onto new lines when links are included in the message. 
- small change that inherits the display, resulting in a correct lay out of elements in the message block body.

- bug found in production. messages are not being displayed as expected. See screenshots below.

### How do I test this?
- open messenger and compose messages with a number of elements including highlight and tagging, and urls etc.

#### Things to consider:
  1. Does this change any APIs?

BEFORE:
- channels
Screenshot 2023-08-14 at 09 42 26

- collapsed chat
Screenshot 2023-08-14 at 09 42 54

- full screen
Screenshot 2023-08-14 at 09 43 31

AFTER:
- channels
Screenshot 2023-08-14 at 09 44 06

- collapsed chat
Screenshot 2023-08-14 at 09 44 54

- full screen
Screenshot 2023-08-14 at 09 45 11

### DIFF
--- a/src/components/message/styles.scss
+++ b/src/components/message/styles.scss
@@ -65,7 +65,7 @@
     font-size: 14px;
     line-height: 17px;
     color: theme.$color-greyscale-12;
-    display: inherit;
+    display: flex;
     flex-direction: column;
     gap: 4px;
 
--- a/src/components/message/styles.scss
+++ b/src/components/message/styles.scss
@@ -65,7 +65,7 @@
     font-size: 14px;
     line-height: 17px;
     color: theme.$color-greyscale-12;
-    display: flex;
+    display: inherit;
     flex-direction: column;
     gap: 4px;
 
--- a/src/components/content-highlighter/index.tsx
+++ b/src/components/content-highlighter/index.tsx
@@ -69,7 +69,7 @@ export class ContentHighlighter extends React.Component<Properties> {
             },
           }}
         >
-          {this.renderContent(this.props.message)}
+          <div>{this.renderContent(this.props.message)}</div>
         </Linkify>
       );
     } else {


ratik21 (approved):
CSS is magic!

src/components/message/styles.scss
```diff
@@ -65,7 +65,7 @@
     font-size: 14px;
     line-height: 17px;
     color: theme.$color-greyscale-12;
-    display: flex;
+    display: inherit;
     flex-direction: column;
```
dalefukami, line 69: 1. Looks like 'block-body' contains a number of different elements. Did we verify rendering looks correct for all of the cases where other things are rendered and not just the message?
2. Can we remove the flex-direction and gap? Those only apply if you are display: flex, I think.
3. Can we just remove `display` completely here? Why does it need to inherit from the parent? I think that'd make it difficult to figure out what it's actually rendering as.

domw30: Hey Dale!

1. I went through and checked the elements and all looked okay to me. I compared with development. From what I could see the elements render okay. Is there anything that comes to mind that would potentially be broken in the UI ? I just had another quick check and appears okay.

2. Good shout yeah I can remove those.

3. Yeah, same as above, I should be able to just remove it here. Will make that change. Thanks!

domw30: This was partly why I didn't want to merge straight away just in case I may have missed something, felt like an update that was 'too easy' to not have caused any unwanted changes in the ui 😬

dalefukami: > Is there anything that comes to mind that would potentially be broken in the UI ?

I guess mainly the gap that was there. I imagine that was probably the reason for it being flex in the first place.

dalefukami (approved):

PR #926, opened 2023-08-14
Use message list for preview rendering in conversation list

dalefukami:
### What does this do?

Uses the message list as the first option when rendering the message preview in the conversation list.

The lastMessage property on a conversation is really meant to be a shortcut for rendering previews without fetching a conversation's messages. However, in cases where we actually have the message data, using the most recent message in the list makes more sense and saves a lot of juggling of trying to manage the lastMessage property.

https://github.com/zer0-os/zOS/assets/43770/75f73777-f1f1-4136-b021-9bd30cec2230

https://github.com/zer0-os/zOS/assets/43770/d2af8cba-d0b5-4056-8c36-83969d693f10

### DIFF
--- a/src/components/messenger/list/index.test.tsx
+++ b/src/components/messenger/list/index.test.tsx
@@ -342,24 +342,51 @@ describe('messenger-list', () => {
       ]);
     });
 
-    test('messagePreview', () => {
-      const state = subject([
-        {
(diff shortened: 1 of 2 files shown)


domw30 (commented):
👍

PR #928, opened 2023-08-14
Wrap callbacks to hide event information

dalefukami:
### What does this do?

Wraps some callbacks connected to click events to prevent serialization of the event to redux

Turns out these warnings actually cost quite a bit of time. I started noticing it when navigating the conversation panels as shown in the videos.

https://github.com/zer0-os/zOS/assets/43770/a95a90a1-5e1b-42ba-9aa1-ff2da0335a9f

https://github.com/zer0-os/zOS/assets/43770/3d50e631-d0b3-4b9e-a88b-f81c2064441d


domw30 (commented):
👍

PR #932, opened 2023-08-15
Simplify the channel message polling when in public ZOS mode

dalefukami:
### What does this do?

Simplifies the logic for polling for new messages in public ZOS mode by just always running a poll. If there's no user and we have an active channel then they must be viewing it in public mode.

Simplifies the code as we don't have to react to all the various state changes

### How do I test this?

When in public ZOS mode verify that new messages that are sent from someone else show up within the poll interval

### DIFF
--- a/src/store/messages/saga.ts
+++ b/src/store/messages/saga.ts
@@ -24,6 +24,7 @@ import { Events as ChatEvents, getChatBus } from '../chat/bus';
 import { ChannelEvents, conversationsChannel } from '../channels-list/channels';
 import { Uploadable, createUploadableFile } from './uploadable';
 import { chat } from '../../lib/chat';
+import { activeChannelIdSelector } from '../chat/selectors';
 
 export interface Payload {
   channelId: string;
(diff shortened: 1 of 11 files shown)


domw30 (commented):
nice tidy up 🧹

src/store/messages/saga.ts
```diff
@@ -477,9 +461,15 @@ export function* getPreview(message) {
...
-  while (yield select(rawShouldSyncChannels(action.payload.channelId))) {
-    yield call(fetchNewMessages, action);
+function* pollForPublicChannelMessages() {
+  while (true) {
+    const currentUser = yield select(currentUserSelector());
+    const isPublicZOS = !currentUser?.id;
+    const activeChannelId = yield select(activeChannelIdSelector);
+
+    if (isPublicZOS && activeChannelId) {
+      yield call(fetchNewMessages, activeChannelId);
```
ratik21, line 471: same question here, curious about how the polling would work i am a logged in user, but in multiple networks (so i can see the normal view with all the public + private channels)

src/store/messages/saga.ts
```diff
@@ -477,9 +461,15 @@ export function* getPreview(message) {
   return yield call(getLinkPreviews, link[0].href);
 }
 
-function* syncChannelsTask(action) {
-  while (yield select(rawShouldSyncChannels(action.payload.channelId))) {
-    yield call(fetchNewMessages, action);
+function* pollForPublicChannelMessages() {
+  while (true) {
+    const currentUser = yield select(currentUserSelector());
+    const isPublicZOS = !currentUser?.id;
```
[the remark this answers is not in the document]
dalefukami: Yeah, public ZOS is when you're not logged in. You can actually see public channels, etc, but you're not a user so you can't do things like...send messages, etc. It also means you can't get real time message events because we don't know who you are.

src/store/messages/saga.ts
dalefukami: It won't poll. You don't have to because real time events handle everything.

PR #933, opened 2023-08-16
Searching conversations list should reset scroll position

ratik21:
### What does this do?

Before:

https://github.com/zer0-os/zOS/assets/33264364/aa3df655-9ce3-4070-adee-4c2d3c9084e0

After

https://github.com/zer0-os/zOS/assets/33264364/3721fc00-a14c-4fef-be18-507fcaf706a5

### DIFF
--- a/src/components/scrollbar-container/index.test.tsx
+++ b/src/components/scrollbar-container/index.test.tsx
@@ -50,15 +50,4 @@ describe('ScrollbarContainer', () => {
     wrapper.setState({ showPanel: false });
     expect(wrapper.find('.scrollbar-container__panel')).toHaveLength(0);
   });
-
-  it('resets scroll position when children change', () => {
-    const wrapper: any = subject();
-    const scrollContainerRef = { current: { scrollTop: 100 } }; // Simulate the ref object
-    wrapper.instance().scrollContainerRef = scrollContainerRef;
-
-    const newProps = { children: <div>New Content</div> };
-    wrapper.setProps(newProps);
-
-    expect(scrollContainerRef.current.scrollTop).toBe(0); // Check if scrollTop was reset
-  });
 });
--- a/src/components/scrollbar-container/index.test.tsx
+++ b/src/components/scrollbar-container/index.test.tsx
@@ -10,19 +10,6 @@ describe('ScrollbarContainer', () => {
     hasPanel: false,
   };
 
-  const initialDocument = global.document;
-
-  beforeEach(() => {
-    // @ts-ignore
-    global.document = {
-      getElementById: jest.fn(() => ({ scrollTop: 100 } as any)),
-    };
-  });
-
-  afterEach(() => {
-    global.document = initialDocument;
-  });
-
   const subject = (props: Partial<Properties> = {}) => {
     const allProps = { ...defaultProps, ...props };
     return shallow(<ScrollbarContainer {...allProps} />);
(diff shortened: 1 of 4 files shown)


src/components/scrollbar-container/index.tsx
```diff
@@ -34,6 +34,11 @@ export class ScrollbarContainer extends React.Component<Properties, State> {
   }
 
   componentDidUpdate(prevProps: Properties) {
+    if (prevProps.children !== this.props.children) {
+      const div = document.getElementById('scrollbar-container__content');
```
dalefukami, line 38: We should use a ref here rather than fetching by id.

dalefukami: Haha, actually, there's already a ref for that node.... we should be able to use `this.scrollContainerRef.current`

src/components/scrollbar-container/index.test.tsx
```diff
@@ -50,4 +63,12 @@ describe('ScrollbarContainer', () => {
     wrapper.setState({ showPanel: false });
     expect(wrapper.find('.scrollbar-container__panel')).toHaveLength(0);
   });
+
+  it('resets scroll position when new children are provided', () => {
```
dalefukami, line 67: Is this the general scroll component? I'm not sure we want ALL things in a scroll component to reset its scroll position.

ratik21: The only other place it's used is the "channels list", in which case this will get triggered only when we add a new channel. So i had put it here, in general component. Is that..not correct? 🤔

src/components/scrollbar-container/index.tsx
dalefukami: In this case they both happened to work but they are not the same at all.

Using the document `getElementById` queries the whole DOM and will get all elements that have id. If you have more than one scroll container then you don't know which one you'll get. Using `ref`s is a React mechanism for getting the actual rendered DOM element associated with the component. It should be used very rarely because you're supposed to manage a component in "react-y" ways. But for certain things you do sometimes need access. This is one of them.

src/components/scrollbar-container/index.test.tsx
dalefukami: I was actually thinking about this a bit more. Will this impact the whole conversation list? If it does then it seems like there might be some weird user experiences if you've scrolled down to look at a conversation and then someone starts a new conversation with you...the children would change and your list would scroll to the top.

src/components/scrollbar-container/index.tsx
ratik21: got it.

src/components/scrollbar-container/index.test.tsx
ratik21: Ok, i got it working from the `conversation-list-panel` component. So basically whenever a search input changes, and we're displays the "results", only then we scroll to top. But i had to use `refs` for this (like i had to define a new one).

src/components/scrollbar-container/index.tsx
```diff
@@ -23,6 +23,11 @@ export class ScrollbarContainer extends React.Component<Properties, State> {
     this.checkScrollBottom = this.checkScrollBottom.bind(this);
   }
 
+  // used when we have "new" results while searching conversations, to scroll to the top
```
dalefukami, line 26: Probably don't need this comment. Anything that renders with a scrollbar can use this if we want.

dalefukami (approved):

PR #935, opened 2023-08-16
Fix scss deprecation warning

dalefukami:
### What does this do?

Fixes a sass deprecation warning


domw30 (commented):
Oh nice was this for the warning that was shown in the console when starting up the app? 👍

PR #936, opened 2023-08-16
Remove unused zApps

dalefukami:
### What does this do?

Removes the non-chat zApps

Since these are not actively in use they just end up costing us maintenance time with no benefit. Remove them all for now.


domw30 (commented):
💪

PR #937, opened 2023-08-17
feat: add reset-password route

domw30:
### What does this do?
- adds `/reset-password` route/page.
- updates authentication footer to be used in request password reset ui
- redirects authenticated user away from `/reset-password`.
- updates tests
- resets request password reset when enter request password reset page

- to handle reset-password flow on its own route/page.

### How do I test this?
- run the app. go to login page. select email. select `reset` in footer. you will be directed to the `/reset-password` page.

#### Things to consider:
  1. Does this change any APIs?

`/reset-password`

https://github.com/zer0-os/zOS/assets/39112648/35e53ec2-b7f2-47cb-ad2c-199bb5b83e86

### DIFF
--- a/src/authentication/footer/footer.tsx
+++ b/src/authentication/footer/footer.tsx
@@ -5,11 +5,12 @@ import { RegistrationStage } from '../../store/registration';
 
 import { bem } from '../../lib/bem';
 import './styles.scss';
+import { RequestPasswordResetStage } from '../../store/request-password-reset';
 
 const c = bem('authentication-footer');
 
 export interface Properties {
-  stage: RegistrationStage;
+  stage: RegistrationStage | RequestPasswordResetStage;
 }
 
 export class Footer extends React.Component<Properties> {
@@ -18,10 +19,13 @@ export class Footer extends React.Component<Properties> {
       return null;
     }
 
+    const loginPromptText =
+      this.props.stage === RequestPasswordResetStage.SubmitEmail ? 'Back to ' : 'Already on ZERO? ';
+
     return (
       <div className={c('')}>
         <div>
-          <span>Already on ZERO? </span>
+          <span>{loginPromptText}</span>
           <Link to='/login'>Log in</Link>
         </div>
 
--- a/src/authentication/request-password-reset/index.tsx
+++ b/src/authentication/request-password-reset/index.tsx
@@ -98,20 +98,21 @@ export class RequestPasswordReset extends React.Component<Properties, State> {
 
   render() {
     const { stage } = this.props;
+    const isResetPasswordDone = stage === RequestPasswordResetStage.Done;
 
     return (
       <>
         <div {...cn('')}>
           <div {...cn('heading-container')}>
             <h3 {...cn('heading')}>Reset Password</h3>
-            {stage === RequestPasswordResetStage.Done && this.renderSuccessMessage()}
+            {isResetPasswordDone && this.renderSuccessMessage()}
 
-            {stage !== RequestPasswordResetStage.Done && (
+            {!isResetPasswordDone && (
               <div {...cn('sub-heading')}>Enter your ZERO account email and we’ll send a reset link</div>
             )}
           </div>
 
-          {stage !== RequestPasswordResetStage.Done && this.renderForm()}
+          {!isResetPasswordDone && this.renderForm()}
         </div>
         <Footer stage={this.props.stage} />
       </>
--- a/src/authentication/request-password-reset/index.tsx
+++ b/src/authentication/request-password-reset/index.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 
+import { Footer } from '../footer/footer';
 import { Button, Input, Alert } from '@zero-tech/zui/components';
 
 import { bem, bemClassName } from '../../lib/bem';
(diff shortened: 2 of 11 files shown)


src/reset-password.tsx
```diff
@@ -0,0 +1,52 @@
...
+    return (
+      <>
+        <ThemeEngine theme={Themes.Dark} />
+        <div {...cn('')}>
+          <div {...cn('logo-container')}>
+            <ZeroLogo />
+          </div>
+
+          <RequestPasswordResetContainer />
+          {/* TODO: Add ResetPasswordConfirmation - this will redirect user to login*/}
```
domw30, line 45: Will be adding the ResetPasswordConfirmation container here and logic to display the request or the confirmation

src/store/request-password-reset/saga.ts
```diff
@@ -43,6 +43,10 @@ export function* requestPasswordReset() {
 export function* watchRequestPasswordReset() {
   while (true) {
     yield take(SagaActionTypes.EnterRequestPasswordResetPage);
+
+    // set the stage back to the initial stage
+    yield put(setStage(RequestPasswordResetStage.SubmitEmail));
```
ratik21, line 48: Does `RequestPasswordResetStage.SubmitEmail` mean that the email is "submitted" OR "about to be submitted"? Wondering from the comment you had on line 47 🤔

src/authentication/footer/footer.tsx
```diff
@@ -18,10 +19,13 @@ export class Footer extends React.Component<Properties> {
       return null;
     }
 
+    const loginPromptText =
+      this.props.stage === RequestPasswordResetStage.SubmitEmail ? 'Back to ' : 'Already on ZERO? ';
```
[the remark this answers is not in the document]
domw30: Screenshot 2023-08-22 at 10 13 35

Yep thats there :)

src/authentication/request-password-reset/index.tsx
```diff
@@ -99,18 +100,21 @@ export class RequestPasswordReset extends React.Component<Properties, State> {
...
-
-          {stage !== RequestPasswordResetStage.Done && (
-            <div {...cn('sub-heading')}>Enter your ZERO account email and we’ll send a reset link</div>
-          )}
-        </div>
+      <>
+        <div {...cn('')}>
+          <div {...cn('heading-container')}>
+            <h3 {...cn('heading')}>Reset Password</h3>
+            {stage === RequestPasswordResetStage.Done && this.renderSuccessMessage()}
```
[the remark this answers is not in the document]
domw30: Nice. Updated!

src/reset-password.scss
```diff
@@ -0,0 +1,35 @@
...
+  overflow-x: hidden;
+  width: 100vw;
+  height: 100vh;
+  top: 0;
+  left: 0;
+  z-index: 10;
+
+  > * {
+    flex-shrink: 0;
+  }
```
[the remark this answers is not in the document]
domw30: Correct. I think this should be handled in another PR though as per previous guidance on refactoring changes out of scope of task.

src/store/request-password-reset/saga.ts
domw30: `RequestPasswordResetStage.SubmitEmail` means the Submit Email stage - 'about to be submitted'.

Two stages for the Request Password Reset Stage:

1. Submit Email - user is required to submit an email address.
2. Done - user has submitted the email and the request password reset is done / complete.

ratik21: Can we rename `SubmitEmail` to `RequestEmail`/`RequestSubmitEmail`? I think that would be more clear, the stage wasn't clear to me upon first look

src/reset-password.scss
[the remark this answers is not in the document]
ratik21: I am not sure if i follow, would like more clarity on this though (just to be sure). I agree that we should have separate PR's for refactoring the code we're touching. But in this particular case we're "adding" a new file, no? If we import the scss from the `invite.scss` here, and re-use it, is that considered into "refactoring" as well? 🤔 

just a doubt, not a big deal anyways. we can absolutely do it in a follow up PR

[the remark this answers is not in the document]
domw30: Yeah I think you are also right here. I decided not to use the `invites.scss` file for this PR as really we would want to rename `invites.scss` if we will be re-using it elsewhere - or move it to another folder etc. So for review purposes I thought best to add another in relation to the reset-password to prevent making changes not directly related to `reset-password` i.e. the invite files. But maybe I should have just imported the `invites.scss` file instead, that also would have been fairly simple to follow!

src/store/request-password-reset/saga.ts
domw30: I think `RequestPasswordResetStage. RequestEmail` & `RequestPasswordResetStage. RequestSubmitEmail ` are also a little confusing. 

How about `RequestPasswordResetStage.EnterEmail` ?

ratik21: `RequestPasswordResetStage.EnterEmail` sounds good. But..how are `RequestEmail` & `SubmitEmail` different?

ratik21 (approved):

PR #942, opened 2023-08-21
Zos 611: Send toast notification if the file a user is trying to attach exceeds the size limit

ratik21:
### What does this do?

Pops up a toast notification if the file size exceeds the limit (currently 10MB)

https://github.com/zer0-os/zOS/assets/33264364/1d4dcc5b-3667-434b-9c80-68c821ac01cd

### DIFF
--- a/src/components/message-input/menu/menu.tsx
+++ b/src/components/message-input/menu/menu.tsx
@@ -1,10 +1,12 @@
 import React from 'react';
 import Dropzone from 'react-dropzone';
-import { dropzoneToMedia, Media } from '../utils';
+import { bytesToMB, dropzoneToMedia, Media } from '../utils';
 import { IconPaperclip } from '@zero-tech/zui/icons';
 import { IconButton } from '../../icon-button';
 
 import './styles.scss';
+import { ToastNotification } from '@zero-tech/zui/components';
+import { config } from '../../../config';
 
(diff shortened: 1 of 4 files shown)


dalefukami (approved):

src/components/message-input/menu/menu.tsx
```diff
@@ -1,38 +1,78 @@
...
   };
 
+  renderToastNotification = () => {
+    const maxSize = bytesToMB(config.cloudinary.max_file_size);
+
+    return (
+      <ToastNotification
+        title={''}
+        description={`File exceeds ${maxSize} size limit`}
+        actionTitle=''
```
domw30, line 41: I think with your changes to the component in zUI you can remove the `actionTitle` attribute ?

domw30 (approved):
Looks good. Just need to add the `viewportClassName` attribute and add `padding: 64px` to the stylesheet for correct positioning.

The default duration for the toast is 10 seconds, not sure if its worth asking product/design if this is the desired amount of time for this toast? I'd assume so!

ratik21:
> The default duration for the toast is 10 seconds, not sure if its worth asking product/design if this is the desired amount of time for this toast? I'd assume so!

Currently the toast appears instantly when the file is rejected 🤔 , there is no delay

PR #945, opened 2023-08-22
fix: align giphy container

domw30:
### What does this do?
- fixes the position of the giphy container.

- as per figma designs.

### How do I test this?
- open messenger and click giphy icon.

#### Things to consider:
  1. Does this change any APIs?

BEFORE
Screenshot 2023-08-22 at 11 43 28

AFTER
Screenshot 2023-08-22 at 11 42 53

### DIFF
--- a/src/components/message-input/giphy/styles.scss
+++ b/src/components/message-input/giphy/styles.scss
@@ -3,8 +3,8 @@
 .giphy__container {
   width: 500px;
   position: absolute;
-  bottom: 65px;
-  right: 74px;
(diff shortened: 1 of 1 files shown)


domw30:
The container here could be improved and hacky positioning with `left`, `bottom` removed with a greater amount of time spent on somer refactoring, but not sure how much time we actually want to spend on this. Quick change to fix for now. If we'd prefer a full refactor please let me know

dalefukami (approved):

PR #946, opened 2023-08-22
Remove unused profileId for mentions

dalefukami:
### What does this do?

This removes a couple of references to mentioned users' profile id.

It seems like at one point in time there may have been a profileId associated with a mentioned user but based on the api at the moment that is never set.


domw30 (commented):
Nice clean up!

PR #948, opened 2023-08-22
Remove unused property from edit message data

dalefukami:
### What does this do?

Removes a property (`mentionedUserIds`) from the Edit Message payload that we're not using and was always send as undefined. It is passed as an independent parameter already.


ratik21 (approved):

dalefukami:
Uggg... this is actually a bigger mess than I original thought. But, in any case, sending undefined here is wrong and sending the props as it currently was is also wrong and sending a list of ids as strings in that prop is also wrong. So, going to continue and merge this and it can be restructured when the endpoint is fixed.

PR #949, opened 2023-08-22
Fix the Message model type

dalefukami:
### What does this do?

Fixes the message model type to match what's actually happening


domw30 (commented):
👍 🤦

PR #950, opened 2023-08-23
feat: implement scroll to bottom on message send

domw30:
### What does this do?
- when a message is sent the chat view container will scroll to the bottom to the most recently sent message.

- so when replying to a message, the chat view container will scroll to the bottom to the most recently sent message.

### How do I test this?
- open a chat, scroll to a higher message, reply to that message and press send.

#### Things to consider:
  1. Does this change any APIs?
  

 

https://github.com/zer0-os/zOS/assets/39112648/fe44d015-ca14-4089-8b39-4dafe9bc1b07

### DIFF
--- a/src/components/chat-view-container/chat-view.test.tsx
+++ b/src/components/chat-view-container/chat-view.test.tsx
@@ -138,6 +138,22 @@ describe('ChatView', () => {
     expect(wrapper.find(InvertedScroll).hasClass('channel-view__inverted-scroll')).toBe(true);
   });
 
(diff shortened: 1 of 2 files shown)


ratik21 (approved):
to confirm - we also scroll to bottom when i hit "enter" to send a new message, right?

domw30:
> to confirm - we also scroll to bottom when i hit "enter" to send a new message, right?

Hey! Yeah the keyboard event will call the `scrollToBottom` - had a quick check to be sure :)

PR #955, opened 2023-08-24
fix: decrease z-index of reply card

domw30:
### What does this do?
- fixes a minor z-index issue for the reply card.

- some overlap I noticed -

Screenshot 2023-08-24 at 18 43 43

### How do I test this?
- open zOS and in a channel, select to reply to a message. When the reply card is visible, open a direct one to one conversation from the conversation list. The reply card should not be layered over the direct-message container.

#### Things to consider:
  1. Does this change any APIs?

https://github.com/zer0-os/zOS/assets/39112648/15eae57f-b8da-4aef-b429-d538c5e4a42a

### DIFF
--- a/src/components/reply-card/styles.scss
+++ b/src/components/reply-card/styles.scss
@@ -7,7 +7,7 @@
 
 .reply-card {
   position: relative;
-  z-index: 10;
+  z-index: 9;
   margin-bottom: 8px;
   &__container {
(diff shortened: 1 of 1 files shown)


ratik21 (approved):

domw30:
Closing pull request due to `reply-card` changes made in another task

PR #956, opened 2023-08-25
fix: remove outdated fix scroll implementation

domw30:
### What does this do?
Proposing the following changes:
- Removes the outdated fixScroll function from the InvertedScroll component.
- Simplifies the scrolling logic, leading to more predictable behaviour.

- To resolve the unexpected "scroll to bottom" behavior seen when editing messages or loading older messages.
- To improve the performance of the InvertedScroll component by eliminating redundant and excessive function calls.
- To make the component more maintainable by reducing complexity.

### How do I test this?
- Open messenger and when a conversation has loaded, immediately attempt to edit a message, or scroll up to load older messages. The scroll to bottom should no longer occur.

#### Things to consider:
  1. Does this change any APIs?

Whats happening - 

https://github.com/zer0-os/zOS/assets/39112648/c3c6fd10-9691-4ce5-8f4f-d11af388e2e3

I identified that the root cause was tied to the fixScroll function in our InvertedScroll component (code that was added 9 months ago). This function was continuously being called up to a 1000 times (for roughly 5-10 seconds) after a particular event, such as a new element containing InvertedScroll was initialised.

Appears to be unnecessary and also created unpredictable and undesirable scrolling behaviours. If any interactions were made on the page during the time fixScroll was excessively executing, unexpected "scroll to bottom" events would be triggered.

It seems the desired functionality is already well-handled by the `adjustScrollPositionForContentChanges` function, so, I've made the decision to remove the outdated fixScroll related code.

https://github.com/zer0-os/zOS/assets/39112648/3353e492-cfa2-4636-9246-4af906411689

### DIFF
--- a/src/components/inverted-scroll/index.tsx
+++ b/src/components/inverted-scroll/index.tsx
@@ -1,10 +1,6 @@
 import React from 'react';
 import classNames from 'classnames';
-import debounce from 'lodash.debounce';
 import './styles.scss';
-
-const SCROLL_HEIGHT_FIXER_DELAY_MS = 5;
-const SCROLL_HEIGHT_FIXER_ITERATIONS = 1000;
 
 export interface Properties {
   className?: string;
@@ -54,8 +50,6 @@ export class InvertedScroll extends React.Component<Properties, undefined> {
 
   scrollToBottom() {
     this.scrollWrapper.scrollTop = this.scrollWrapper.scrollHeight;
-
-    this.scrollFixer();
   }
 
   setScrollWrapper = (element: HTMLElement) => {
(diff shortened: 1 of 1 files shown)


ratik21 (approved):
Could you please test loading messaging in channels, DM's & group DM's even when we're "recieving" new messages. Just to confirm if there's a case we didn't miss out on before removing this code.

domw30:
> Could you please test loading messaging in channels, DM's & group DM's even when we're "recieving" new messages. Just to confirm if there's a case we didn't miss out on before removing this code.

To confirm - @ratik21 and I tested this together to ensure we're still getting the correct behaviours.

domw30:
Thanks for the review @ratik21, I'm hesitant to merge this one today - would like to get @dalefukami's thoughts as well in case we've missed something this code was supporting.

dalefukami (approved):
My guess is that this had something to do with delayed loading of url previews/images/etc which would change the height of the scrollable content and since we're using an inverted scroll that all gets kinda dicey. It's possible this is no longer an issue because of new display techniques that weren't available when this was originally written (copied from zero-web likely). In any case, I'm good with removing it and fixing whatever issue may arise.

dalefukami (changes requested):
Looks like this has broken some functionality:


https://github.com/zer0-os/zOS/assets/43770/b9370184-f87a-4071-9530-fcb64e808a0b

domw30:
> Looks like this has broken some functionality:
> 
>  brokenscroll.mp4

Ah yes good spot, I'll look into this.

domw30:
@dalefukami - I re-introduced the `fixScroll` functionality here but reduced the number set for `SCROLL_HEIGHT_FIXER_ITERATIONS` from `1000` to `200` so it is not so aggressive. Reducing the number of iterations appears to resolve the unpredictable scrolling, and continues to scroll to bottom on load.

dalefukami:
I kinda feel like we're hacking this thing without truly understanding what it's doing. Shrinking it to 200 maybe works perfectly for the exact network conditions that you tested on but what if the network is slower or there are 50 large images to download? How do you know that 200 is the right number?

dalefukami:
Closing in favor of #972

PR #958, opened 2023-08-29
Fix autoscroll behavior occurring in conversation window with edit message function

ratik21:
### What does this do?

Currently the 'x' or 'tick' icon gets hidden (check before video), if trying to edit a message in the bottom of the screen/viewport. Fixed now (check after video).

I used the `clientHeight` & `ref.getBoundingClientRect()` methods for this to calculate the distance 😅 

Before:

https://github.com/zer0-os/zOS/assets/33264364/bd95f24a-5d41-439c-9b37-a2c273ce2f10

After:

https://github.com/zer0-os/zOS/assets/33264364/170c2d1a-d37c-4abf-a845-5ce7e04282fd

### DIFF
--- a/src/components/chat-view-container/chat-view.tsx
+++ b/src/components/chat-view-container/chat-view.tsx
@@ -177,6 +177,7 @@ export class ChatView extends React.Component<Properties, State> {
               showSenderAvatar={this.props.showSenderAvatar}
               showTimestamp={messageRenderProps.showTimestamp}
               showAuthorName={messageRenderProps.showAuthorName}
+              scrollContainerRef={this.scrollContainerRef}
               {...message}
             />
           </div>
--- a/src/components/message/edit-message-actions/edit-message-actions.test.tsx
+++ b/src/components/message/edit-message-actions/edit-message-actions.test.tsx
@@ -12,6 +12,7 @@ describe('EditMessageActions', () => {
       secondaryTooltipText: '',
       onEdit: jest.fn(),
       onCancel: jest.fn(),
+      scrollContainerRef: React.createRef(),
       ...props,
     };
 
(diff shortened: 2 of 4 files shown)


dalefukami (commented):
Did you happen to look into `scrollIntoView` functionality? It seems like maybe we could use default javascript functionality rather than trying to do calculations and controlling scroll ourselves.

ratik21:
> Did you happen to look into `scrollIntoView` functionality? It seems like maybe we could use default javascript functionality rather than trying to do calculations and controlling scroll ourselves.

It works. But i guess it goes back to the `fixScroll` implementation which Dom is working on. If you directly try to "edit" a message without waiting for 5-10s, it will start pushing the scroll the bottom. Otherwise this works well. 

So i guess if that is resolved, this should be good to go.

ratik21:
This is how it works now with the updated functionality:

https://github.com/zer0-os/zOS/assets/33264364/26504782-81b1-4001-987c-97fd1899e973


And this is the behaviour if i try to edit the message immediately (without waiting for `fixScroll` to finish): 

https://github.com/zer0-os/zOS/assets/33264364/13ac4903-e4dc-4a21-8bb2-7dc01338a540



(the latter is currently happening on dev env)

domw30:
> This is how it works now with the updated functionality:
> 
>  Screen.Recording.2023-08-30.at.12.26.21.AM.mov 
> And this is the behaviour if i try to edit the message immediately (without waiting for `fixScroll` to finish):
> 
>  Screen.Recording.2023-08-30.at.12.30.47.AM.mov 
> (the latter is currently happening on dev env)

uhh yeah that bit of code `fixScroll` is so aggressive!! Will try have another look today if I can get other work wrapped up. I think `fixScroll` will end up being removed.

dalefukami (approved):
Yep, I'd consider this part resolved. The other issue is definitely a separate thing.

PR #959, opened 2023-08-29
feat: set cursor if replying to a message

domw30:
### What does this do?
- focuses on the message input when reply prop exists

- so the user can starting typing immediately without an additional mouse event.

### How do I test this?
- select reply to a message from the message menu and check if the message input is focussed.

#### Things to consider:
  1. Does this change any APIs?

https://github.com/zer0-os/zOS/assets/39112648/22149e15-2b94-4c7d-af58-54cb706f1822

### DIFF
--- a/src/components/message-input/index.tsx
+++ b/src/components/message-input/index.tsx
@@ -85,12 +85,16 @@ export class MessageInput extends React.Component<Properties, State> {
     return this.props.clipboard || windowClipboard();
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps) {
     if (this.props.onMessageInputRendered) {
       this.props.onMessageInputRendered(this.textareaRef);
     }
     if (this.state.isSendTooltipOpen && this.isSendingEnabled) {
       this.setState({ isSendTooltipOpen: false });
+    }
+
+    if (this.props.reply && !prevProps.reply) {
+      this.focusInput();
     }
   }
 
@@ -127,7 +131,7 @@ export class MessageInput extends React.Component<Properties, State> {
 
   focusInput() {
     if (this.textareaRef && this.textareaRef.current) {
-      this.textareaRef.current.focus();
+      setTimeout(() => this.textareaRef.current.focus(), 1);
     }
   }
 
@@ -340,10 +344,6 @@ export class MessageInput extends React.Component<Properties, State> {
   renderInput() {
     const hasInputValue = this.state.value?.length > 0;
     const reply = this.props.reply;
-
-    if (this.props.reply) {
-      setTimeout(() => this.focusInput(), 100);
-    }
 
     return (
       <div {...cn('container')}>
--- a/src/components/message-input/index.tsx
+++ b/src/components/message-input/index.tsx
@@ -85,16 +85,12 @@ export class MessageInput extends React.Component<Properties, State> {
     return this.props.clipboard || windowClipboard();
   }
 
-  componentDidUpdate(prevProps) {
+  componentDidUpdate() {
     if (this.props.onMessageInputRendered) {
       this.props.onMessageInputRendered(this.textareaRef);
     }
     if (this.state.isSendTooltipOpen && this.isSendingEnabled) {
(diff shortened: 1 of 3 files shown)


src/components/message-input/index.tsx
```diff
@@ -341,6 +341,10 @@ export class MessageInput extends React.Component<Properties, State> {
     const hasInputValue = this.state.value?.length > 0;
     const reply = this.props.reply;
 
+    if (this.props.reply) {
```
domw30, line 344: For some reason that I haven't quite been able to pin point, a timeout is required here to ensure the input is focussed when replying to a message. Even if we update componentDidUpdate like so 

```
  componentDidUpdate(prevProps: Properties) {
    if (this.props.reply && !prevProps.reply) {
      this.focusInput();
    }

    if (this.props.onMessageInputRendered) {
      this.props.onMessageInputRendered(this.textareaRef);
    }
    if (this.state.isSendTooltipOpen && this.isSendingEnabled) {
      this.setState({ isSendTooltipOpen: false });
    }
  }
  ```

and log 

```
 focusInput() {
    console.log('Textarea ref:', this.textareaRef.current);
    if (this.textareaRef && this.textareaRef.current) {
      this.textareaRef.current.focus();
    }
  }
  ```
 the textareaRef is logged and we can see the textarea is the current focus, but the cursor is not active. The timeout seems to fix this.

dalefukami: Hmm... I don't think it makes sense to do this in the render function. This function can get called thousands of times at any point the app decides something needs to be rerendered. This will cause the input to constantly steal focus away even if you've clicked around to somewhere else.

domw30: Spent some time digging into this more today. I believe this is *another* radix ref issue. When selecting the `reply` option in the dropdown menu, the active element (even if we set the focus to the textarea) is the dropdown menu option, or it can be the dropdown menu trigger. Seems like the timeout I have added, (replaced from render function now to `focusInput` function) is a fix for the actions that are queued in the event loop. Hacky, but seems a good workaround in this instance.

domw30: Here I am console logging  `console.log(document.activeElement, this.textareaRef.current);`. You can see the `this.textareaRef.current` is the expected textarea, but the active element remains the menu option. Selecting an emoji also uses the `focusInput` function, and as you can see in the demo here the activeElement is the textarea. A `e.stopPropagation` in `DropdownMenu` may help with this, but from digging around I believe this is due to refs controlled on the Radix side of things.

https://github.com/zer0-os/zOS/assets/39112648/07ea2a4a-7ceb-414d-a884-429e1a599d4b

dalefukami:
@domw30 you're right! It's Radix being super aggressive about claiming focus.

I pushed a proposed change to the PR which handles the over aggressiveness closer to the source. I think having it here puts it much closer to the actual problem so it's more clear why we have it and then other completely unrelated components don't need to know how to handle the situation. They can just work properly.

The other issue with using componentDidUpdate was that we'd then have to detect extra things for the situation when you already had a reply but changed it and that just starts to get unwieldy eventually. Here's a video demonstrating that problem:

https://github.com/zer0-os/zOS/assets/43770/46f6f43f-b864-481f-a9f9-427afe452509

and a video after my proposed change:

https://github.com/zer0-os/zOS/assets/43770/3e351c00-c76b-4fcf-b5a2-5c4d4167c758

src/platform-apps/channels/messages-menu/index.tsx
```diff
@@ -39,20 +39,27 @@ export class MessageMenu extends React.Component<Properties, State> {
     );
   }
 
+  // Our zUI DropdownMenu component actively steals focus.
+  // In order to allow actions to change focus without the dropdown menu stealing it back,
+  // we delay publishing the event by releasing the thread for a single tick.
+  delayEvent = (handler) => setTimeout(handler, 1);
+  onEdit = () => this.delayEvent(this.props.onEdit);
```
domw30, line 46: Nice! This looks good to me. Thanks for checking that out Dale 🚀

PR #961, opened 2023-08-31
feat: update messages to use sender avatar

domw30:
### What does this do?
- replaces div with background img with zUI Avatar component.

- should be using the re-usable Avatar component available.

### How do I test this?
- open a group chat and check the sender Avatars 

#### Things to consider:
  1. Does this change any APIs?

Screenshot 2023-08-31 at 09 38 48

### DIFF
--- a/src/components/message/styles.scss
+++ b/src/components/message/styles.scss
@@ -173,7 +173,7 @@
   &__author-avatar {
     visibility: hidden;
 
-    > div:first-child {
+    > *:first-child {
       cursor: default;
     }
   }
--- a/src/components/message/styles.scss
+++ b/src/components/message/styles.scss
@@ -172,13 +172,10 @@
 
   &__author-avatar {
     visibility: hidden;
-    background-position: center;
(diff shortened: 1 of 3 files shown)


src/components/message/styles.scss
```diff
@@ -172,13 +172,10 @@
...
     visibility: hidden;
-    background-position: center;
-    background-repeat: no-repeat;
-    background-size: cover;
-    width: 32px;
-    height: 32px;
-    overflow: hidden;
-    border-radius: 50%;
+
+    > div:first-child {
```
dalefukami, line 176: This is pretty dicey. We should never rely on the internal structure of a component library. How 'bout `> *` which is any direct descendant.

dalefukami (approved):

src/components/message/styles.scss
domw30: ah I should have gone with my instinct, I used `> *` before changing to `> div`. Shall update before merging. Thanks Dale !

PR #965, opened 2023-09-04
Remove users from search results after they've been added to a pending group chat

ratik21:
### What does this do?

If you have selected a user to "start a group chat", then it won't be displayed in the search results.

Before (user is selected & also displayed in the search results):
image

After (selectedUser is NOT displayed in the results):
image

### DIFF
--- a/src/components/messenger/autocomplete-members/index.tsx
+++ b/src/components/messenger/autocomplete-members/index.tsx
@@ -30,13 +30,13 @@ export class AutocompleteMembers extends React.Component<Properties, State> {
     }
 
     const items = await this.props.search(searchString);
-    const itemsToOption = items.map(itemToOption);
+    const options = items.map(itemToOption);
     const selectedOptions = this.props.selectedOptions || [];
-    const filteredOptions = itemsToOption.filter((o) => !selectedOptions.find((s) => s.value === o.value));
+    const filteredOptions = options.filter((o) => !selectedOptions.find((s) => s.value === o.value));
     this.setState({ results: filteredOptions });
   };
 
-  onSelect(event) {
+  itemSelected(event) {
     const clickedId = event.currentTarget.dataset.id;
     const selectedUser = this.state.results.find((r) => r.value === clickedId);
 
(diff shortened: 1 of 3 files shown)


src/components/messenger/autocomplete-members/index.tsx
```diff
@@ -29,25 +30,32 @@ export class AutocompleteMembers extends React.Component<Properties, State> {
     }
 
     const items = await this.props.search(searchString);
-
-    this.setState({ results: items.map(itemToOption) });
+    const itemsToOption = items.map(itemToOption);
```
dalefukami, line 33: I think these would be `options` now that we've mapped "items" to "options".

src/components/messenger/list/start-group-panel/index.tsx
```diff
@@ -67,7 +67,11 @@ export class StartGroupPanel extends React.Component<Properties, State> {
       <>
         <PanelHeader title='Select members' onBack={this.props.onBack} />
         <div {...cn('search')}>
-          <AutocompleteMembers search={this.props.searchUsers} onSelect={this.selectOption}>
+          <AutocompleteMembers
+            search={this.props.searchUsers}
+            onSelect={this.selectOption}
+            selectedOptions={this.state.selectedOptions}
```
dalefukami, line 73: I can't decide if this would feel more like a common component if we used `value` instead of `selectedOptions`. What do you think?

dalefukami (approved):
Just some minor thoughts.

src/components/messenger/autocomplete-members/index.tsx
ratik21: makes sense

src/components/messenger/list/start-group-panel/index.tsx
ratik21: Yeah from my perspective if we directly use `value` here then it wouldn't feel much like a common component, it may be too specific 🤔

dalefukami: Basically all standard HTML input components use `value`. That's why it could make sense to use that. It doesn't matter, I don't think. Was just thinking out loud.

PR #972, opened 2023-09-07
Fix scroll to bottom issues

dalefukami:
### What does this do?

This removes a lot of the manual scroll position management that was happening in the InvertedScroll component and replaces it with mechanisms more built into the browsers.

Notes:
* There's still an edge case when infinite scrolling up. To be handled in another story.
* Pinning to the bottom may not quite work correctly in Safari as the overflow-anchor doesn't appear to be supported.

We've been having issues with the chat window randomly scrolling to the bottom of the history. Smooth out the user experience.

### How do I test this?

Exercise various scroll states and events in a conversation:
* When opening the converstation it shoudl be scrolled to the bottom
* When scrolled to the bottom the conversation should stay pinned there. New messages "push the conversation up".
* Scroll up a bit and then receive a new message. The conversation view should remain mostly stationary.
* Scroll all the way up to trigger a fetch for more messages then quickly scroll slightly downwards. When the new data comes in the view should remain mostly stationary (Note: this scenario will get better in the next story)

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,6 @@
     "if-emoji": "^0.1.0",
     "linkify-react": "^4.0.2",
     "linkifyjs": "^4.0.2",
-    "lodash.debounce": "^4.0.8",
     "lodash.get": "^4.4.2",
     "lodash.kebabcase": "^4.1.1",
     "lodash.uniqby": "^4.7.0",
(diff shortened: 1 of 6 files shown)


domw30:
Awesome Dale! Huge improvement to the UX. I noticed when going from a full messenger chat (scrolled to bottom) to collapsed chat window, the scroll bar loses its place, but that's probably not a huge issue, or if an issue at all, just something I noticed.

PR #991, opened 2023-09-11
Mark messages read instantaneously for a channel/conversation

ratik21:
### What does this do?

Marks all messages as read instantly when you select a channel or conversation. 
image

Earlier marking messages as read felt slow/unresponsive, since we were only marking as read when you load all the messages (including background fetching). 

I thought about just "hiding" the unread count marker on the UI, but i guess if the user quickly switches b/w loading conversations/channels, then the unread count marker would dissappear and popup again & again, which won't be much ideal i think.

https://github.com/zer0-os/zOS/assets/33264364/d2156f80-9204-4fc1-a211-88f25bda123f

### DIFF
--- a/src/components/messenger/list/conversation-list-panel/index.tsx
+++ b/src/components/messenger/list/conversation-list-panel/index.tsx
@@ -27,6 +27,7 @@ export interface Properties {
   startConversation: () => void;
   onCreateConversation: (userId: string) => void;
   onConversationClick: (conversationId: string) => void;
+  markConversationAsRead: (payload: { channelId: string }) => void;
 }
 
 interface State {
@@ -116,6 +117,7 @@ export class ConversationListPanel extends React.Component<Properties, State> {
 
   openExistingConversation = (id: string) => {
     this.props.onConversationClick(id);
+    this.props.markConversationAsRead({ channelId: id });
     this.setState({ filter: '' });
   };
 
--- a/src/store/channels/saga.ts
+++ b/src/store/channels/saga.ts
@@ -5,7 +5,6 @@ import { SagaActionTypes, receive, schema, removeAll } from '.';
 import { joinChannel as joinChannelAPI, markAllMessagesAsReadInChannel as markAllMessagesAsReadAPI } from './api';
 import { takeEveryFromBus } from '../../lib/saga';
 import { Events as ChatEvents, getChatBus } from '../chat/bus';
-import { ChannelEvents, conversationsChannel } from '../channels-list/channels';
 import { currentUserSelector } from '../authentication/saga';
 import { fetch as fetchMessages } from '../messages/saga';
 import { activeChannelIdSelector } from '../chat/selectors';
@@ -45,7 +44,9 @@ export function* markAllMessagesAsRead(channelId, userId) {
 }
 
 // mark all messages as read in current active channel (only if you're not in full screen mode)
-export function* markChannelAsReadIfActive(channelId) {
+export function* markChannelAsReadIfActive(action) {
+  const { channelId } = action.payload;
+
   const activeChannelId = yield select(activeChannelIdSelector);
   if (channelId !== activeChannelId) {
     return;
@@ -62,7 +63,9 @@ export function* markChannelAsReadIfActive(channelId) {
 }
 
 // mark all messages in read in current active conversation
-export function* markConversationAsReadIfActive(channelId) {
+export function* markConversationAsReadIfActive(action) {
+  const { channelId } = action.payload;
+
   const activeConversationId = yield select((state) => state.chat.activeConversationId);
   if (channelId !== activeConversationId) {
     return;
@@ -102,14 +105,9 @@ export function* clearChannels() {
 }
 
 export function* saga() {
-  yield takeEveryFromBus(yield call(conversationsChannel), ChannelEvents.MessagesLoadedForChannel, (event) =>
-    markChannelAsReadIfActive(event.channelId)
-  );
-  yield takeEveryFromBus(yield call(conversationsChannel), ChannelEvents.MessagesLoadedForConversation, (event) =>
-    markConversationAsReadIfActive(event.channelId)
-  );
-
   yield takeLatest(SagaActionTypes.JoinChannel, joinChannel);
+  yield takeLatest(SagaActionTypes.MarkAllMessagesAsReadInChannel, markChannelAsReadIfActive);
+  yield takeLatest(SagaActionTypes.MarkAllMessagesAsReadInConversation, markConversationAsReadIfActive);
 
   yield takeEveryFromBus(yield call(getChatBus), ChatEvents.UnreadCountChanged, unreadCountUpdated);
 }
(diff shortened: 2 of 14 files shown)


src/components/messenger/list/conversation-list-panel/index.tsx
```diff
@@ -116,6 +117,7 @@ export class ConversationListPanel extends React.Component<Properties, State> {
 
   openExistingConversation = (id: string) => {
     this.props.onConversationClick(id);
+    this.props.markConversationAsRead({ channelId: id });
```
dalefukami, line 120: Seems like maybe this logic should be in the sagas. Opening the conversation should already be calling out to redux/saga to set the active conversation id. I think that can be a saga that also marks the conversation as read rather than relying on UI to remember to do both of these things every time.

ratik21: The reason i didn't do it that way because we're setting the activeConversationId in a lot of other places as well (for eg. creating a new conversation, onNewConversationCreated hook etc). On the UI side it was only 2 places where i had to place this (one here, and other when we set active channel ID). Not sure which decision is better here

dalefukami: Why wouldn't we want to mark messages as read on those other cases too?

dalefukami: I think one issue here is that the name we've used isn't truly representative of what we're doing any more. `setActiveConversationId` isn't truly what we're doing when you look at it from a system perspective. It's too technical of a name. Really what we want to do is `openConversation` or something along those lines. When you look at it that way, it makes sense that what we're trying to say is "any time we open a conversation we should mark it as read".

ratik21: > Why wouldn't we want to mark messages as read on those other cases too?

I thought when we're creating a "new" conversation we wouldn't have any messages to be "marked as read". 😅

ratik21: > I think one issue here is that the name we've used isn't truly representative of what we're doing any more. setActiveConversationId isn't truly what we're doing when you look at it from a system perspective. It's too technical of a name. Really what we want to do is openConversation or something along those lines. When you look at it that way, it makes sense that what we're trying to say is "any time we open a conversation we should mark it as read".

That makes sense i think. Let me do the updates here, and create a follow up PR for the "renaming".

dalefukami: > I thought when we're creating a "new" conversation we wouldn't have any messages to be "marked as read"

Yeah, that's probably the 1 case where we don't need to (and might fail if we don't have a proper channel id yet) but that would be the edge case, right? There could be so many ways to open a conversation. For example, the in app notifications might become clickable. In which case, we'd be opening a conversation from there and would want this behavior.

My point being, the majority case is that we'd want to do this...and we don't want to have to remember to do that everywhere.

ratik21: did the updates.

src/components/messenger/list/conversation-list-panel/index.tsx
```diff
@@ -26,7 +26,7 @@ export interface Properties {
   search: (input: string) => any;
   startConversation: () => void;
   onCreateConversation: (userId: string) => void;
-  onConversationClick: (conversationId: string) => void;
+  openConversation: (payload: { conversationId: string }) => void;
```
dalefukami, line 29: I think this still makes more sense as an event rather than a direct function call. At this level of component we should couple ourselves to the format of the redux actions. I.E., `onConversationClick` is still ok. (if anything, `onConversationSelected` since that's really what the user is doing)

src/store/channels/saga.ts
```diff
@@ -45,35 +44,44 @@ export function* markAllMessagesAsRead(channelId, userId) {
...
-  const activeChannelId = yield select(activeChannelIdSelector);
-  if (channelId !== activeChannelId) {
-    return;
-  }
-
+export function* markChannelAsRead(channelId) {
   const currentUser = yield select(currentUserSelector());
   const isMessengerFullScreen = yield select((state) => state.layout.value.isMessengerFullScreen);
-  const activeChannelInfo = activeChannelId ? yield select(rawChannelSelector(activeChannelId)) : null;
+  const activeChannelInfo = yield select(rawChannelSelector(channelId));
```
dalefukami, line 50:
```suggestion
  const channelInfo = yield select(rawChannelSelector(channelId));
```

It's not the active channel.

src/store/channels/saga.ts
```diff
@@ -102,14 +110,9 @@ export function* clearChannels() {
...
-  yield takeEveryFromBus(yield call(conversationsChannel), ChannelEvents.MessagesLoadedForChannel, (event) =>
-    markChannelAsReadIfActive(event.channelId)
-  );
-  yield takeEveryFromBus(yield call(conversationsChannel), ChannelEvents.MessagesLoadedForConversation, (event) =>
-    markConversationAsReadIfActive(event.channelId)
-  );
-
   yield takeLatest(SagaActionTypes.JoinChannel, joinChannel);
+  yield takeLatest(SagaActionTypes.OpenChannel, openChannel);
+  yield takeLatest(SagaActionTypes.OpenConversation, openConversation);
```
dalefukami, line 115: Would be good to use the new format we discussed where these parse out the `payload` so that the sagas can just take the information they need.

This sort of thing: `takeLatest(SagaActionTypes.OpenConversation, ({payload}) => yield openConversation(payload.channelId));`

That way the other sagas don't have to format the arguments in weird ways when they call `openConversation`

dalefukami (approved):
All comments are just suggestions for future work. This looks great!

src/store/channels/saga.ts
ratik21: could you please check if this makes sense 😅 : https://github.com/zer0-os/zOS/pull/991/commits/25d3ad2a6856e4604970153a6ed8f05aead344ba

dalefukami: Yep, that makes sense. Although, now that I see it again, I wonder if that function has to `yield` in order for the `takeLatest` to properly work. :thinking:

dalefukami: In this case it's fine because I wouldn't expect anything to happen rapidly enough for multiple calls to step over each other but we should definitely verify this. We might have to tweak the pattern.

src/store/channels/saga.ts
```diff
@@ -45,35 +44,42 @@ export function* markAllMessagesAsRead(channelId, userId) {
...
+
+export function* openChannel(channelId) {
+  if (!channelId) {
     return;
   }
 
-  const currentUser = yield select(currentUserSelector());
-  const activeConversationInfo = activeConversationId ? yield select(rawChannelSelector(activeConversationId)) : null;
+  yield put(setActiveChannelId(channelId));
+  yield call(markChannelAsRead, channelId);
```
dalefukami, line 73: Just realized, we probably want this to be a `spawn`. There's no reason to delay anything else for this request. Same for the other one.

ratik21: done. thanks

PR #992, opened 2023-09-11
Fix reply rendering

dalefukami:
### What does this do?

Fixes rendering of replies to shorter messages. Previous fix for long messages didn't take this into account.

Before:
image

After:
image


domw30 (commented):
👍

PR #999, opened 2023-09-12
Render link preview that does not have a thumbnail

dalefukami:
### What does this do?

Some link previews don't have a thumbnail image. This handles that scenario.


domw30 (commented):
👍

PR #1003, opened 2023-09-13
refactor: remove includesRewardsAvatar (duplicate prop)

domw30:
### What does this do?
- removes `includesRewardsAvatar` prop.
- replaces conditions using `includesRewardsAvatar` prop with `isMessengerFullScreen`.

- `includesRewardsAvatar` is used in the same way `isMessengerFullScreen`. Made sense to me to remove and replace.

```
includeRewardsAvatar: layout?.value?.isMessengerFullScreen,
isMessengerFullScreen: layout?.value?.isMessengerFullScreen,
```

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/messenger/list/index.tsx
+++ b/src/components/messenger/list/index.tsx
@@ -54,7 +54,6 @@ export interface Properties extends PublicProperties {
   includeTitleBar: boolean;
   allowClose: boolean;
   allowExpand: boolean;
-  includeRewardsAvatar: boolean;
   userName: string;
   userHandle: string;
   userAvatarUrl: string;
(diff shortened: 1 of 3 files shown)


src/components/messenger/list/index.tsx
```diff
@@ -241,12 +239,14 @@ export class Container extends React.Component<Properties, State> {
...
           <div {...cnMessageList('profile-bar')}>
-            {this.props.includeRewardsAvatar && (
+            {this.props.isMessengerFullScreen && (
               <div {...cnMessageList('avatar-container')}>{this.renderSettingsMenu()}</div>
             )}
 
             <FeatureFlag featureFlag='enableRewards'>
-              <div {...cnMessageList('rewards-container', this.props.includeRewardsAvatar && 'with-avatar')}>
+              <div
+                {...cnMessageList('rewards-container', this.props.isMessengerFullScreen && 'with-full-screen-modifier')}
```
domw30, line 248: how do we feel about `with-full-screen-modifier` ? I felt like this made more sense but if there are better suggestions please let me know 🙂

src/components/messenger/list/index.tsx
```diff
@@ -241,12 +239,14 @@ export class Container extends React.Component<Properties, State> {
 
         {this.props.stage === SagaStage.None && (
           <div {...cnMessageList('profile-bar')}>
-            {this.props.includeRewardsAvatar && (
+            {this.props.isMessengerFullScreen && (
```
dalefukami, line 242: I think this would work better as `includesAvatar` or `includesUserSettings` or whatever it is. The reasoning for this is that if we write our components with as little knowledge of what the overall app is structured like then they're simpler to move around and reuse, etc. So, this component probably shouldn't even know whether we're in full screen or not. It should only know how the app wants it to be rendered. From that perspective the higher level component is saying, "in this location with these conditions `includeUserSettings` when you're rendered".

dalefukami (approved):

PR #1004, opened 2023-09-13
refactor: add user account container for settings menu and rewards bar plus some minor naming updates

domw30:
### What does this do?
- improves the naming of containers relevant to settings menu and rewards bar.
- adds a user account container for settings menu and rewards bar.
- replaces bem with bemClassName.
- updates scss class names.

- improves readability.
- improves class naming to reflect what the containers actually are.

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/settings-menu/index.tsx
+++ b/src/components/settings-menu/index.tsx
@@ -1,12 +1,12 @@
 import * as React from 'react';
 
+import { EditProfileContainer } from '../edit-profile/container';
 import { IconLogOut3, IconUser1 } from '@zero-tech/zui/icons';
 import { Address, Avatar, DropdownMenu, Modal } from '@zero-tech/zui/components';
 
-import { bem } from '../../lib/bem';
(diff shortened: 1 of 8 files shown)


src/components/settings-menu/index.tsx
```diff
@@ -47,13 +47,13 @@ export class SettingsMenu extends React.Component<Properties, State> {
   };
 
   renderSettingsHeader() {
-    const c = bem('header');
+    const cn = bemClassName('header');
```
dalefukami, line 50: Since we're in the groove here the next step should probably be to move this to the top of the file so we're not redefining functions on every render. Also, the name doesn't match convention and would probably be pretty trivial to change?

dalefukami (approved):

src/components/settings-menu/index.tsx
domw30: Yeah sounds good I'll revisit this one in a follow up

PR #1005, opened 2023-09-13
refactor: extracts rewards button from rewards bar container

domw30:
### What does this do?
- extracts rewards button from rewards bar container.
- extracts rewards button tests from rewards bar container tests.
- adds rewards faq modal test coverage for rewards bar container tests.

- extracting the rewards button elements to keep the container tidy.
- the reason for these changes is so in a follow up PR I will be renaming RewardsBar to RewardsContainer. The idea is that RewardsContainer contains the button, the pop up and the faq modal.

### DIFF
--- a/src/components/messenger/list/index.tsx
+++ b/src/components/messenger/list/index.tsx
@@ -54,7 +54,6 @@ export interface Properties extends PublicProperties {
   includeTitleBar: boolean;
   allowClose: boolean;
   allowExpand: boolean;
-  includeRewardsAvatar: boolean;
   userName: string;
   userHandle: string;
   userAvatarUrl: string;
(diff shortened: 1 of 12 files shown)


src/components/messenger/list/styles.scss
```diff
@@ -45,13 +45,13 @@ $side-padding: 16px;
...
     align-items: center;
     justify-content: center;
-    padding-right: 8px;
     width: 100%;
     height: 64px;
     background-color: theme.$color-primary-3;
 
     &--center {
       width: unset;
+      padding-right: 8px;
```
domw30, line 54: This fixes an existing, slightly off-centre rewards button when not in full messenger view.

src/components/messenger/list/index.tsx
```diff
@@ -238,7 +238,7 @@ export class Container extends React.Component<Properties, State> {
         )}
 
         <FeatureFlag featureFlag='enableRewards'>
-          <div {...cnMessageList('rewards-container', this.props.includeUserSettings && 'center')}>
+          <div {...cnMessageList('rewards-container', !this.props.includeUserSettings && 'center')}>
```
domw30, line 241: For `center` to make sense, the bang needs to be added. This means that if `!this.props.includeUserSettings` is true, which is equal to `isMessengerFullScreen` being false, then use `center` to centre the rewards button when not in full screen.

Before this change it is the opposite.

dalefukami (approved):

PR #1006, opened 2023-09-13
refactor: rename rewards bar to rewards container

domw30:
### What does this do?
- renames `RewardsBar` to `RewardsContainer`.

- `RewardsBar` isn't really a rewards bar as such. With the additional refactoring, `Container` appears to make more sense as this contains the `RewardsButton`, `RewardsPopupContainer` and the `RewardsFAQModal`. The `RewardsBar` is no longer a bar due to extraction of other elements. This is now acting more like a container.

### DIFF
--- a/src/components/messenger/list/index.test.tsx
+++ b/src/components/messenger/list/index.test.tsx
@@ -13,7 +13,7 @@ import { Stage } from '../../../store/create-conversation';
 import { RegistrationState } from '../../../store/registration';
 import { LayoutState } from '../../../store/layout/types';
 import { RewardsState } from '../../../store/rewards';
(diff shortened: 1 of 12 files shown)


domw30:
@dalefukami - proposed name change for the `RewardsBar`. Let me know if `RewardsContainer` could be improved. I think it makes sense.

dalefukami (approved):
Yeah, the only downside is that "Container" sort of implies a certain type of thing (a wrapper that deals with the global state) and this one doesn't really do that. Although, it's very possible that this should now fetch from state rather than have to take in all these properties so it's probably a great name. :laughing:

In any case, I think this is fine.

PR #1010, opened 2023-09-13
feat: send a message with Matrix implementation

domw30:
### What does this do?
- implements ability to send a message with Matrix.

- deprecate sendbird.

### How do I test this?
- follow the instructions in the README to get an instance of Element and Matrix set up to send messages using Matrix.

#### Things to consider:
  1. Does this change any APIs?


dalefukami (approved):
Just double checked the risk spots where we might impact the Sendbird implementation. Didn't review the Matrix related code at all as we can evolve that as we go now that more people will be attacking it.

PR #1014, opened 2023-09-14
refactor: (message input) replace local icon button component 

domw30:
### What does this do?
- replaces local icon button component with zUI icon button component for message input

- in order to delete/remove IconButton (zOS local) and IconButton (zos-component-library) from the code base, we need to replace all uses of IconButton with the component from zUI.

How it looks in PROD
Screenshot 2023-09-14 at 16 56 35

How it looks after LOCAL change
Screenshot 2023-09-14 at 16 56 38


dalefukami (approved):
YES!! Love the cleanup. Thank you.

PR #1027, opened 2023-09-18
feat: (matrix) differentiate between 1-on-1 and group conversations

domw30:
### What does this do?
- Updates the Matrix Client to differentiate between 1-on-1 and group conversations.
- Updates the Channels List saga to call client.
- Updates the Sendbird Client.
- Adds test coverage / fixes failing tests.

- to differentiate between 1-on-1 and group conversations.
- to ensure Sendbird Client also still fetches conversations.

### How do I test this?
- run Matrix/Element and check one on one conversations appear in the conversation list, and channels appear in the channel list.

#### Things to consider:
  1. Does this change any APIs?

Screenshot of 1-on-1 & channels lists
Screenshot 2023-09-18 at 20 47 31

### DIFF
--- a/src/lib/chat/matrix-client.test.ts
+++ b/src/lib/chat/matrix-client.test.ts
@@ -10,9 +10,7 @@ const getMockAccountData = (data = {}) => {
   return jest.fn((eventType) => {
     if (eventType === 'm.direct') {
       return {
-        event: {
-          content: data,
-        },
+        event: { content: data },
       };
     }
     return {};
@@ -176,7 +174,10 @@ describe('matrix client', () => {
 
   describe('getAccountData', () => {
     it('get account data', async () => {
-      const getAccountData = getMockAccountData([{ id: '!abcdefg' }, { id: '!hijklmn' }]);
+      const getAccountData = getMockAccountData({
+        '@mockuser1': ['!abcdefg'],
+        '@mockuser2:': ['!hijklmn', '!opqrstuv'],
+      });
 
       const client = subject({
         createClient: jest.fn(() => getSdkClient({ getAccountData })),
@@ -207,7 +208,11 @@ describe('matrix client', () => {
     });
 
     it('fetches and returns correct account data when type is "m.direct"', async () => {
-      const getAccountData = getMockAccountData([{ id: '!abcdefg' }, { id: '!hijklmn' }]);
+      const getAccountData = getMockAccountData({
+        '@mockuser1': ['!abcdefg'],
+        '@mockuser2:': ['!hijklmn', '!opqrstuv'],
+      });
+
       const client = subject({
         createClient: jest.fn(() => getSdkClient({ getAccountData })),
       });
@@ -215,7 +220,14 @@ describe('matrix client', () => {
       await client.connect('username', 'token');
 
       const result = await client.getAccountData('m.direct');
-      expect(result).toEqual({ event: { content: [{ id: '!abcdefg' }, { id: '!hijklmn' }] } });
+      expect(result).toEqual({
+        event: {
+          content: {
+            '@mockuser1': ['!abcdefg'],
+            '@mockuser2:': ['!hijklmn', '!opqrstuv'],
+          },
+        },
+      });
     });
   });
 });
--- a/src/lib/chat/matrix-client.test.ts
+++ b/src/lib/chat/matrix-client.test.ts
@@ -6,6 +6,19 @@ const getRoom = (props: any = {}) => ({
   ...props,
 });
 
+const getMockAccountData = (data = {}) => {
+  return jest.fn((eventType) => {
+    if (eventType === 'm.direct') {
+      return {
+        event: {
+          content: data,
+        },
+      };
+    }
+    return {};
+  });
+};
+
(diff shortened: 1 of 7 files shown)


src/lib/chat/matrix-client.test.ts
```diff
@@ -119,4 +135,87 @@ describe('matrix client', () => {
...
+      const conversations = await client.getConversations();
+
+      expect(conversations).toHaveLength(1);
+      expect(conversations[0].id).toEqual('dm-id');
+    });
+  });
+
+  describe('getAccountData', () => {
+    it('get account data', async () => {
+      const getAccountData = getMockAccountData([{ id: '!abcdefg' }, { id: '!hijklmn' }]);
```
dalefukami, line 179: `[{ id: '!abcdefg' }, { id: '!hijklmn' }]` doesn't look like the same format we were seeing from the event the other day. Wasn't it more like:
```js
  {
    'some-user-id': ['!abcdefg', '!hijklmn'],
    'other-user-id': [...more-ids...]
  }
```
If so, we should update all the tests to use that format.

src/lib/chat/sendbird-client.ts
```diff
@@ -71,6 +71,16 @@ export class SendbirdClient implements IChatClient {
     }
   }
 
+  async getConversations() {
+    try {
+      const conversations = await get<any>('/directMessages/mine');
```
dalefukami, line 76: Does this mean we can delete the function in the `api` file?

src/store/channels-list/saga.test.ts
```diff
@@ -20,17 +19,29 @@ import { AsyncListStatus } from '../normalized';
...
-const MOCK_CHANNELS = [
-  { name: 'channel 1', id: 'channel_0001', url: 'channel_0001', icon: 'channel-icon', hasJoined: false },
-  { name: 'channel 2', id: 'channel_0002', url: 'channel_0002', icon: 'channel-icon', hasJoined: false },
-  { name: 'channel 3', id: 'channel_0003', url: 'channel_0003', icon: 'channel-icon', hasJoined: false },
-];
+const mockChannel = (id: string) => ({
+  id: `channel_${id}`,
+  name: `channel ${id}`,
+  icon: 'channel-icon',
+  hasJoined: false,
```
dalefukami, line 28: This should probably have `isChannel: true` just to make sure it's really considered a Channel.

src/store/channels-list/saga.test.ts
```diff
@@ -307,12 +321,20 @@ describe(fetchConversations, () => {
     await expectSaga(fetchConversations)
       .provide([
         [
-          matchers.call.fn(fetchConversationsApi),
-          MOCK_CHANNELS,
+          matchers.call(chat.get),
+          chatClient,
```
dalefukami, line 325: I think these tests could use a follow up cleaning similar to #988. Basically, we shouldn't need to clutter the tests with the default response of the mock client.

Also just a note: if you take the trailing comma off these arrays the formatter will pull them up to a single line (if they're short enough) which is far easier to read.

src/lib/chat/sendbird-client.ts
```diff
@@ -71,6 +71,16 @@ export class SendbirdClient implements IChatClient {
...
+      return (conversations.body || []).map((currentChannel) => toLocalChannel(currentChannel));
```
dalefukami, line 77:
```suggestion
      return (conversations.body || []).map(toLocalChannel);
```
You can pass the mapping function directly to `map`. There's no need to wrap it with a lambda.

dalefukami (approved):
The one required change is the one about the test data format. We should test with exactly the same format that we actually get so just double check. My memory might be incorrect. :)

src/lib/chat/matrix-client.ts
```diff
@@ -168,4 +178,35 @@ export class MatrixClient implements IChatClient {
...
+  private async getFilteredRooms(filterFunc: (roomId: string, dmConversationIds: string[]) => boolean) {
+    if (this.isDisconnected) {
+      return [];
+    }
+
+    if (this.isConnecting) {
+      await this.waitForConnection();
+    }
+
+    const accountData = await this.getAccountData('m.direct');
```
dalefukami, line 207: Looks like the sdk has enums and functions for many of these things. For example: `EventType.Direct` and `accountData.getContent()`. We should probably use these rather than poking at internal data.

src/lib/chat/matrix-client.test.ts
domw30: Yes correct! Updated. Thanks!

```
  const getAccountData = getMockAccountData({
        '@mockuser1': ['!abcdefg'],
        '@mockuser2:': ['!hijklmn', '!opqrstuv'],
      });
```

src/store/channels-list/saga.test.ts
domw30: Nice okay will do a follow up PR!

src/lib/chat/matrix-client.ts
domw30: Yes for sure, I spotted this in the element code you send me yesterday. Updated to use those!

PR #1031, opened 2023-09-20
[Matrix] Search for Users in Create Conversation Flow

ratik21:
### What does this do?

Implements user search using zero-api, but only returns the users who have "linked" their ZERO account with matrix.

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -24,7 +24,7 @@ export class MatrixClient implements IChatClient {
   private connectionResolver: () => void;
   private connectionAwaiter: Promise<void>;
 
-  constructor(private sdk = { createClient }) {}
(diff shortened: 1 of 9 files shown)


ratik21:
will open for review once #1027 is merged.

src/lib/chat/matrix-client.ts
```diff
@@ -67,6 +69,25 @@ export class MatrixClient implements IChatClient {
     return rooms.map(this.mapConversation);
   }
 
+  // async _searchMyNetworksByName(filter: string) {
```
dalefukami, line 72: We should just delete this. It's rarely good to leave commented code around.

dalefukami (approved):

domw30 (commented):
Nice work! 👍

PR #1033, opened 2023-09-20
refactor: improved readability and maintainability enhancements to channels-list saga tests

domw30:
### What does this do?
- enhancements to channels-list saga tests.

- as per previous feedback on channels-list saga work. The aim tidy up the test suite because the tests were cluttered with the default response of the mock client.

### How do I test this?
- run the command: `npm run test src/store/channels-list/saga.test.ts` and ensure all tests are passing.

#### Things to consider:
  1. Does this change any APIs?

Screenshot 2023-09-21 at 14 36 19

### DIFF
--- a/src/store/channels-list/saga.test.ts
+++ b/src/store/channels-list/saga.test.ts
@@ -1,4 +1,4 @@
-import { expectSaga, testSaga } from 'redux-saga-test-plan';
+import { testSaga } from 'redux-saga-test-plan';
 import { call, race, take } from 'redux-saga/effects';
 import * as matchers from 'redux-saga-test-plan/matchers';
 import { chat } from '../../lib/chat';
@@ -20,6 +20,7 @@ import { conversationsChannel } from './channels';
 import { multicastChannel } from 'redux-saga';
 import { ConversationStatus, denormalize as denormalizeChannel } from '../channels';
 import { StoreBuilder } from '../test/store';
+import { expectSaga } from '../../test/saga';
 
 const mockChannel = (id: string) => ({
   id: `channel_${id}`,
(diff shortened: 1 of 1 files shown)


domw30:
@dalefukami - made an attempt to clean these up using the [#988](https://github.com/zer0-os/zOS/pull/988/files) as a reference

src/store/channels-list/saga.test.ts
```diff
@@ -186,34 +134,14 @@ describe('channels list saga', () => {
...
-          ]),
-          [conversation],
-        ],
-        [
-          matchers.call.fn(delay),
-          null,
-        ],
-      ])
+    const additionalProviders = [
+      [matchers.call(chat.get), chatClient],
```
dalefukami, line 138: Shouldn't need this line in any of the tests now that it's defaulted in `subject`

src/store/channels-list/saga.test.ts
```diff
@@ -186,34 +134,14 @@ describe('channels list saga', () => {
...
+      [matchers.call([chatClient, chatClient.getChannels], rootDomainId), [channel]],
+      [matchers.call([chatClient, chatClient.getConversations]), [conversation]],
+      [matchers.call.fn(delay), null],
+    ];
+
+    const { storeState } = await subject(fetchChannelsAndConversations, {}, additionalProviders)
```
dalefukami, line 144:
```suggestion
    const { storeState } = await subject(fetchChannelsAndConversations, {})
         .provide([
             [matchers.call([chatClient, chatClient.getChannels], rootDomainId), [channel]],
             [matchers.call([chatClient, chatClient.getConversations]), [conversation]],
             [matchers.call.fn(delay), null],
         ])
```

The point of the wrapped expectSaga is to allow multiple calls to provide. That's what allows us to set up default providers in `subject` and then "add on".

src/store/channels-list/saga.test.ts
```diff
@@ -387,3 +287,12 @@ describe(fetchConversations, () => {
...
 });
+
+function subject(sagaFunction: any, args?: any, additionalProviders = []) {
+  return expectSaga(sagaFunction, args).provide([
+    ...additionalProviders,
+    [matchers.call.fn(chat.get), chatClient],
+    [matchers.call.fn(chatClient.getChannels), MOCK_CHANNELS],
+    [matchers.call.fn(chatClient.getConversations), MOCK_CONVERSATIONS],
+  ]);
+}
```
dalefukami, line 298:
```suggestion
function subject(...args: Parameters<typeof expectSaga>) {
  return expectSaga(sagaFunction, args)
    .provide([
      [matchers.call.fn(chat.get), chatClient],
      [matchers.call.fn(chatClient.getChannels), MOCK_CHANNELS],
      [matchers.call.fn(chatClient.getConversations), MOCK_CONVERSATIONS],
   ]);
}
```

See comment above. Don't take `additionalProviders` here as it's not necessary. Also, updated the function signature so typescript can help the tests get the arguments right.

src/store/channels-list/saga.test.ts
```diff
@@ -387,3 +287,12 @@ describe(fetchConversations, () => {
     ]);
   });
 });
+
+function subject(sagaFunction: any, args?: any, additionalProviders = []) {
```
dalefukami, line 291: I think I'd expect multiple `subject` functions. One per context of which saga we're testing. Not every saga is going to need the same default providers.

dalefukami (commented):
Nice work. Got pretty close.

PR #1036, opened 2023-09-21
Create conversation via Matrix

dalefukami:
### What does this do?

Adds the Matrix specific functionality to create a Conversation (in Matrix this is a room that "is_direct" and is in the user's account list of `m.direct` events)

Notes:
* This functionality works for group Conversations too but the process of creating a group conversation has other checks that need to be switched to Matrix first.
* Currently the receiving user still has to manually accept the invitation in Element. Future work will do this automatically on behalf of the user.

### How do I test this?

In zOS search for a user that has a matrix id set and start a conversation with them. You will see the conversation created and can send a message. Verify the other user received the invite by going to Element.

### DIFF
--- a/src/lib/chat/matrix/utils.ts
+++ b/src/lib/chat/matrix/utils.ts
@@ -1,0 +1,41 @@
+import { EventType, MatrixClient as SDKMatrixClient } from 'matrix-js-sdk';
+
+// Copied from the matrix-react-sdk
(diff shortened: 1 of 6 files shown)


src/lib/chat/matrix/utils.ts
```diff
@@ -0,0 +1,39 @@
+import { EventType, MatrixClient as SDKMatrixClient } from 'matrix-js-sdk';
+
+// Copied from the matrix-react-sdk
```
domw30, line 3: 💪

PR #1038, opened 2023-09-21
Fix opening a one on one from matrix

dalefukami:
### What does this do?

Missed a spot where we're assuming an "other member" exists but for Matrix convos we don't have that mapped yet.

### DIFF
--- a/src/components/messenger/chat/index.tsx
+++ b/src/components/messenger/chat/index.tsx
@@ -114,7 +114,7 @@ export class Container extends React.Component<Properties, State> {
       return '';
     }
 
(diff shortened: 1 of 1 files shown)


src/components/messenger/chat/index.tsx
```diff
@@ -114,7 +114,7 @@ export class Container extends React.Component<Properties, State> {
       return '';
     }
 
-    if (this.isOneOnOne()) {
+    if (this.isOneOnOne() && this.props.directMessage.otherMembers[0]) {
```
domw30, line 117: 👍

PR #1039, opened 2023-09-21
Make Matrix rooms private by default for now

dalefukami:
### What does this do?

Private seems like the better default until proven otherwise


src/lib/chat/matrix-client.ts
```diff
@@ -225,7 +225,7 @@ export class MatrixClient implements IChatClient {
     isOneOnOne: room.getMembers().length === 2,
     otherMembers: [],
     lastMessage: null,
-    groupChannelType: GroupChannelType.Public,
+    groupChannelType: GroupChannelType.Private,
```
domw30, line 228: Agreed, better to be private by default and change to public if desired

PR #1046, opened 2023-09-25
Move `searchMentionableUsersForChannel` to sendbird client class

ratik21:


src/platform-apps/channels/util/api.test.ts
```diff
@@ -1,12 +0,0 @@
...
-describe('userMentionSearch', () => {
-  it('maps to user mentions format', async function () {
-    const user1 = { id: 'd-1', name: 'dale', profileImage: 'http://example.com' };
-    const apiSearch = jest.fn().mockResolvedValue([user1]);
-
-    const searchResults = await searchMentionableUsersForChannel(null, null, apiSearch);
-
-    expect(searchResults).toEqual([{ id: 'd-1', display: 'dale', profileImage: 'http://example.com' }]);
-  });
-});
```
ratik21, line 12: wasn't sure about how to re-add the test here. Since i moved the entire logic into a single function in the class 🤔

dalefukami: I'm going to go with, "it doesn't matter because we're deprecating this anyway"

dalefukami (approved):

src/platform-apps/channels/util/api.test.ts
ratik21: got it

PR #1047, opened 2023-09-25
Add extra debugging for Matrix events

dalefukami:
### What does this do?

Adds debugging output for the various enum'ed events that the Matrix SDK provides

To make it easier to find which events we need to listen to during development


domw30 (commented):
Nice idea 👍

PR #1048, opened 2023-09-25
Generalize Channel mapping in Matrix client

dalefukami:
### What does this do?

Refactor channel/conversation mapping to extract a common function

To make the differences between Conversations and Channels more visible


domw30 (commented):
Nice clean up. Will need to set/handle `otherMembers` in mapConversation based on this, I imagine this will be clear with merge conflicts on my PR, shall make the changes.

PR #1049, opened 2023-09-25
Change conversation creation updates to event based

dalefukami:
### What does this do?

This changes the rendering of created conversations to be event based.

When you refresh the browser the Matrix SDK provides an old data state and then replays events. By rendering this way we ensure that both the initial creation works and the rendering upon refresh.

### How do I test this?

Create a conversation and verify that it renders. Refresh the browser and verify that it also shows up then.

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -127,7 +127,6 @@ export class MatrixClient implements IChatClient {
     const result = await this.matrix.createRoom(options);
     // Any room is only set as a DM based on a single user. We'll use the first one.
     await setAsDM(this.matrix, result.room_id, users[0].matrixId);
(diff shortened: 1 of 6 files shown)


src/lib/chat/matrix-client.ts
```diff
@@ -189,6 +189,10 @@ export class MatrixClient implements IChatClient {
       if (event.type === 'm.room.message') {
         this.events.receiveNewMessage(event.room_id, (await mapMatrixMessage(event, this.matrix)) as any);
       }
+
+      if (event.type === 'm.room.create') {
+        this.roomCreated(event);
```
domw30, line 194: Looks good. I've been thinking that maybe the user presence data should be event based too. Going to look at re-working it a little

domw30 (commented):
👍

PR #1051, opened 2023-09-26
[Matrix] Implement user mentions

ratik21:
### What does this do?

Implements user mentions using matrix

image
image

*NOTE*:
+ This implements a simple approach - we use channel.otherMembers as cache, and then filter the results based on the search string. 

**Follow ups**:
- map the `otherMembers` in the channelInfo to the zOS users.
- update the sendbird implementation for "conversations" to use otherMembers as cache as well.

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -27,7 +27,7 @@ import { ParentMessage, User } from './types';
 import { config } from '../../config';
 import { get } from '../api/rest';
 import { MemberNetworks } from '../../store/users/types';
-import { setAsDM } from './matrix/utils';
+import { getFilteredMembersForAutoComplete, setAsDM } from './matrix/utils';
 
 enum ConnectionStatus {
   Connected = 'connected',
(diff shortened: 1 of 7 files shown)


src/lib/chat/matrix-client.ts
```diff
@@ -102,8 +102,10 @@ export class MatrixClient implements IChatClient {
       .then((response) => response?.body || []);
   }
 
-  async searchMentionableUsersForChannel(_channelId: string, _search: string) {
-    return [];
+  async searchMentionableUsersForChannel(channelId: string, search: string) {
```
dalefukami, line 105: Just to summarize what we discussed in standup today, I think we can probably just search in the "other members" field of Conversations. That should always be populated. (Note: Channels probably still needs to be api based for now due to the Sendbird implementation but we don't need to implement the Matrix version for Channels yet anyway)

ratik21: yep, I have added the `otherMembers` field in the matrix-client and using that to search users for mentions in both channel & conversation

image
image

ratik21: I am using the matrixId to display the name of the user for now, and the sendbird implementation is untouched. 

Follow ups would be
- map the `otherMembers` in the channelInfo to the zOS users.
- update the sendbird implementation for "conversations" to use otherMembers as cache as well.

dalefukami: Don't bother updating the sendbird implementation.

src/lib/chat/matrix/utils.test.ts
```diff
@@ -0,0 +1,78 @@
...
+const roomMembers = [
+  {
+    userId: '@domw:zero-synapse-development.zer0.io',
+    matrixId: '@domw:zero-synapse-development.zer0.io',
+    firstName: 'domw',
+    lastName: '',
+    profileId: '',
+    isOnline: false,
+    profileImage: '',
+    lastSeenAt: '',
```
dalefukami, line 14: Do any of these values impact the tests? If not, just remove them.

dalefukami (approved):

src/lib/chat/matrix/utils.test.ts
ratik21: removed

PR #1058, opened 2023-09-28
feat: show last message in conversation list and preview details

domw30:
### What does this do?
- enhances the mapMatrixMessage to include sender details

- to have defined sender details data in the message preview (conversation list item)

### How do I test this?
- open messenger > click on a user in the conversation list that has message history > when conversation renders to top of the list you should see the senders name or 'you'. If the senders name is visible and they sent the last message, this means the sender data is being fetched correctly.

#### Things to consider:
  1. Does this change any APIs?

BEFORE
Screenshot 2023-09-28 at 11 48 21

AFTER
Screenshot 2023-09-28 at 11 47 36

DEMO - SORT ORDER

https://github.com/zer0-os/zOS/assets/39112648/279f913e-f2c5-495d-ac71-7855c357e17a

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -95,6 +95,7 @@ export class MatrixClient implements IChatClient {
 
   async getConversations() {
     const rooms = await this.getFilteredRooms(this.isConversation);
+
     for (const room of rooms) {
       await room.loadMembersIfNeeded();
       const membership = room.getMyMembership();
@@ -341,6 +342,16 @@ export class MatrixClient implements IChatClient {
       name = roomNameEvent.getContent().name;
     }
 
+    let lastMessageEvent = null;
+    const timelineEvents = room.getLiveTimeline().getEvents();
+    for (let i = timelineEvents.length - 1; i >= 0; i--) {
+      if (timelineEvents[i].getType() === EventType.RoomMessage) {
+        lastMessageEvent = timelineEvents[i];
+        break;
+      }
+    }
+    const lastMessage = this.mapMatrixEventToMessage(lastMessageEvent);
+
     return {
       id: room.roomId,
       name,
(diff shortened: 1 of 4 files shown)


domw30:
Edit - Making changes

src/lib/chat/matrix/chat-message.ts
```diff
@@ -22,13 +22,23 @@ async function extractParentMessageData(matrixMessage, sdkMatrixClient: SDKMatri
...
-    createdAt: matrixMessage.origin_server_ts,
+    id: event_id,
+    message: content.body,
+    createdAt: origin_server_ts,
     updatedAt: null,
-    sender: {},
+    sender: {
+      userId: senderId,
+      // remove condition when matrix maps to zos user
+      firstName: senderId === currentMatrixUserId ? 'You' : senderData.displayName,
```
dalefukami, line 37: Just remove this now. If we start hiding bugs then we'll never fix them.

src/lib/chat/matrix-client.ts
```diff
@@ -341,6 +342,16 @@ export class MatrixClient implements IChatClient {
...
 
+    let lastMessageEvent = null;
+    const timelineEvents = room.getLiveTimeline().getEvents();
+    for (let i = timelineEvents.length - 1; i >= 0; i--) {
+      if (timelineEvents[i].getType() === EventType.RoomMessage) {
+        lastMessageEvent = timelineEvents[i];
+        break;
+      }
+    }
+    const lastMessage = this.mapMatrixEventToMessage(lastMessageEvent);
```
dalefukami, line 353: Since we have to filter through them all anyway....is there any reason not to just populate the messages we have in the `messages` property of the conversation? Then we don't need to set the lastMessage property at all.

dalefukami (approved):

src/lib/chat/matrix-client.ts
```diff
@@ -341,6 +342,16 @@ export class MatrixClient implements IChatClient {
       name = roomNameEvent.getContent().name;
     }
 
+    let lastMessageEvent = null;
+    const timelineEvents = room.getLiveTimeline().getEvents();
```
dalefukami, line 346: Do you know how many events this actually is? What happens if there are thousands of messages?

domw30: Was just thinking the same. This could be a massive amount. I'll have a look at the suggestion above re using `messages` property of the conversation.

PR #1060, opened 2023-09-29
refactor: refactor mapToGeneralChannel for clarity and separation of concerns.

domw30:
### What does this do?
- extracts where we are fetching data within mapToGeneralChannel

- separation of concerns - tidy up

#### Things to consider:
  1. Does this change any APIs?


domw30:
Maybe we should extract some of these get methods to the matrix utils file?

PR #1063, opened 2023-10-02
refactor: get all messages from room and remove last message mapping

domw30:
### What does this do?
- gets all messages for a room and adds these to the mapToGeneralChannel. 

- suggested refactor for obtaining lastMessage data.
- removes the need to set `lastMessage` property. Our existing logic handles last message data using the messages data.

### How do I test this?
- open messenger > in the conversation list if a conversation has existing message history you should see the preview for the last/latest message sent. In addition, the conversation list items should be sorted in order based on the conversation with the latest message event.

#### Things to consider:
  1. Does this change any APIs?

Screenshot 2023-10-02 at 10 41 11

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -326,8 +326,8 @@ export class MatrixClient implements IChatClient {
   private mapToGeneralChannel(room: Room) {
     const otherMembers = this.getOtherMembersFromRoom(room).map((userId) => this.mapUser(userId));
     const name = this.getRoomName(room);
-    const lastMessageEvent = this.getLastMessageEvent(room);
-    const lastMessage = this.mapMatrixEventToMessage(lastMessageEvent);
+
+    const messages = this.getAllMessagesFromRoom(room);
 
     return {
       id: room.roomId,
@@ -338,7 +338,8 @@ export class MatrixClient implements IChatClient {
       // as zOS considers any conversation to have ever had more than 2 people to not be 1 on 1
       isOneOnOne: room.getMembers().length === 2,
       otherMembers: otherMembers,
-      lastMessage: lastMessage,
+      lastMessage: null,
+      messages,
       groupChannelType: GroupChannelType.Private,
       category: '',
       unreadCount: 0,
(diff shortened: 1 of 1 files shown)


src/lib/chat/matrix-client.ts
```diff
@@ -412,16 +413,12 @@ export class MatrixClient implements IChatClient {
...
-      }
-    }
-
-    return null;
+  private getAllMessagesFromRoom(room: Room) {
+    const timeline = room.getLiveTimeline().getEvents();
+    const messages = timeline
+      .filter((event) => event.getType() === EventType.RoomMessage)
+      .map(this.mapMatrixEventToMessage);
+    return messages;
```
ratik21, line 421: Will this return "all" messages, or is there a default limit? (like 30 messages)

src/lib/chat/matrix-client.ts
```diff
@@ -338,7 +338,8 @@ export class MatrixClient implements IChatClient {
       // as zOS considers any conversation to have ever had more than 2 people to not be 1 on 1
       isOneOnOne: room.getMembers().length === 2,
       otherMembers: otherMembers,
-      lastMessage: lastMessage,
+      lastMessage: null,
```
ratik21, line 341: also want to understand why we're not using the lastMessage now. Since we're fetching all messages, and this could be set as the last message from that list..?

src/lib/chat/matrix-client.ts
domw30: Yes I believe this will filter all events equal to `m.room.message`, map the event to message, then return all messages. @dalefukami / @ratik21 , should we add a limit? Would a limit cause any unexpected behaviour?

dalefukami: If there's a limit then we should add a limit...as long as the limit applies to the most recent messages (and not just getting the first 30. lol)

domw30: I found that **_if_** we use `createMessagesRequest` there is a limit prop we can use. This would change the set up for getting messages a little, so i will merge this PR and revisit 👍

PR #1066, opened 2023-10-03
Set the conversation name if provided

dalefukami:
### What does this do?

Sets the Matrix room name if provided when creating a conversation


domw30 (commented):
NIce 👍

PR #1067, opened 2023-10-03
Upload group conversation image

dalefukami:
### What does this do?

Uploads the group conversation cover image

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -28,6 +28,7 @@ import { config } from '../../config';
 import { get } from '../api/rest';
 import { MemberNetworks } from '../../store/users/types';
 import { getFilteredMembersForAutoComplete, setAsDM } from './matrix/utils';
+import { uploadImage } from '../../store/channels-list/api';
 
 enum ConnectionStatus {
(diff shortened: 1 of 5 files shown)


src/lib/chat/matrix-client.ts
```diff
@@ -457,6 +472,15 @@ export class MatrixClient implements IChatClient {
     return '';
   }
 
+  private getRoomAvatar(room: Room): string {
```
domw30, line 475: We could add a helper function for all of these 'get' methods at some point 🙂 will take a look if I happen to work around these

src/lib/chat/matrix-client.ts
```diff
@@ -488,4 +512,17 @@ export class MatrixClient implements IChatClient {
...
     return Object.values(content ?? {}).flat();
   }
+
+  private async uploadCoverImage(image) {
+    if (image) {
+      try {
+        return (await uploadImage(image)).url;
+      } catch (error) {
+        // For now, we will just ignore the error and continue to create the room
+        // No reason for the image upload to block the room creation
```
domw30, line 522: 👍

domw30 (commented):
Nice 👍

PR #1068, opened 2023-10-03
Remove existing channels from list When fetching full conversation list

dalefukami:
### What does this do?

Fixes a bug where the receiving users of a new conversation were seeing the conversation in the list twice.

Because of the evented nature of the data changes the room is first noted as a Channel and then eventually gets found to be a Conversation. When we were refetching the whole conversation list this would cause the room id to be in the channel (room) list twice which caused the double rendering.


domw30 (commented):
Awesome, nice fix 👍

PR #1069, opened 2023-10-03
Map ZERO users to matrix users

ratik21:
### What does this do?

Fetches and maps ZERO users <-> matrix users, when we fetch all channels + conversations. This also fixes the user mention display (if the matrix account is linked with ZERO).

image

 
Follow Up
- check on realtime events (messages, join/leave a room etc)

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -29,6 +29,7 @@ import { get } from '../api/rest';
 import { MemberNetworks } from '../../store/users/types';
 import { getFilteredMembersForAutoComplete, setAsDM } from './matrix/utils';
 import { uploadImage } from '../../store/channels-list/api';
+import { union } from 'lodash';
 
 enum ConnectionStatus {
   Connected = 'connected',
@@ -89,6 +90,9 @@ export class MatrixClient implements IChatClient {
       await room.loadMembersIfNeeded();
     }
     const channels = rooms.map(this.mapChannel);
+    await this.mapRoomMembers(channels);
+
+    return channels;
   }
 
   async getConversations() {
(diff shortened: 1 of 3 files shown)


src/lib/chat/matrix-client.ts
```diff
@@ -417,18 +425,31 @@ export class MatrixClient implements IChatClient {
...
-  private mapUser(userId: string): UserModel {
-    const user = this.matrix.getUser(userId);
-    return {
-      userId: userId,
-      matrixId: userId,
+  private mapUser(matrixId: string, zeroUser?): UserModel {
+    const user = this.matrix.getUser(matrixId);
+    let mappedUser = {
+      userId: matrixId,
+      matrixId,
```
domw30, line 432: I wonder if we would need both userId and matrixId fields/props now

src/lib/chat/matrix-client.ts
```diff
@@ -408,7 +414,9 @@ export class MatrixClient implements IChatClient {
     };
   }
 
-  private mapChannel = (room: Room): Partial<Channel> => this.mapToGeneralChannel(room);
+  private mapChannel = (room: Room): Partial<Channel> => {
```
domw30, line 417: How come the return and brackets were added here? Doesn't it work the same ?

domw30 (commented):
looks good to me 👍

src/lib/chat/matrix-client.ts
ratik21: I think it's still relevant to know the user's associated matrixId? Because there could be users who are on matrix but NOT on ZERO (yet).

src/lib/chat/matrix-client.ts
ratik21: ah, good catch. I was trying something here, but i had removed it. forgot to remove the brackets. Will update in my latest PR.

PR #1070, opened 2023-10-04
use map for zero users in `matrix-client.ts`

ratik21:
### What does this do?

Just changes the implementation a little - use a local map, and set the zero users after fetching them (from the matrixIds).

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -44,12 +44,14 @@ export class MatrixClient implements IChatClient {
 
   private accessToken: string;
   private userId: string;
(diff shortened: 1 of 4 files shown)


dalefukami (commented):
I'm not really sure it's a great idea for the Matrix client to be so tightly tied to how zOS works. This should be for interacting with the Matrix system and not other things. This data should be dealt with in the saga layer where the user data is already cached.

ratik21:
> I'm not really sure it's a great idea for the Matrix client to be so tightly tied to how zOS works. This should be for interacting with the Matrix system and not other things. This data should be dealt with in the saga layer where the user data is already cached.

moved to sagas

PR #1073, opened 2023-10-05
CreateConversation: Only send the name field if it is not null

dalefukami:
### What does this do?

Only sends the name field when creating a conversation if it is set.

Matrix rejects fields with `null` based on some type checking (i.e., it is not a string) so just don't send it.

### How do I test this?

Ensure creating 1 on 1 conversations and group conversations without a name work.


domw30 (commented):
👍

PR #1076, opened 2023-10-05
Map message senders in matrix mode

ratik21:
### What does this do?

Maps the message senders to ZERO users when we load messages for a channel/conversation. Messages are being rendered more properly now.

image
image

Also, the names on the conversation list are also of ZERO users:
image

### DIFF
--- a/src/store/messages/saga.ts
+++ b/src/store/messages/saga.ts
@@ -468,6 +468,7 @@ export function* receiveDelete(action) {
 
 export function* receiveNewMessage(action) {
   let { channelId, message } = action.payload;
+  yield call(mapMessageSenders, [message]);
 
   const channel = yield select(rawChannelSelector(channelId));
   const currentMessages = channel?.messages || [];
--- a/src/store/messages/saga.ts
+++ b/src/store/messages/saga.ts
@@ -353,6 +353,7 @@ export function* fetchNewMessages(channelId: string) {
       ],
       channelId
     );
+    yield call(mapMessageSenders, messagesResponse.messages);
 
     yield put(
       receive({
--- a/src/store/messages/saga.ts
+++ b/src/store/messages/saga.ts
@@ -25,6 +25,8 @@ import { Events as ChatEvents, getChatBus } from '../chat/bus';
 import { Uploadable, createUploadableFile } from './uploadable';
 import { chat } from '../../lib/chat';
 import { activeChannelIdSelector } from '../chat/selectors';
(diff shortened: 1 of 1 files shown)


ratik21:
Update: working while sending real-time message now

https://github.com/zer0-os/zOS/assets/33264364/33f66524-40de-4a9c-b08a-5d2fbcd9a55b

domw30 (approved):
Nice work 🚀

src/store/messages/saga.ts
```diff
@@ -90,6 +92,45 @@ const _isActive = (channelId) => (state) => {
...
+
+export function* mapMessageSenders(messages) {
+  if (!featureFlags.enableMatrix) {
+    return;
+  }
+
+  const zeroUsersMap = yield call(getZeroUsersMap);
+  messages.forEach((message) => {
+    const zeroUser = zeroUsersMap[message.sender.userId];
+    if (zeroUser) {
```
dalefukami, line 122: The assumption here is that we already know about the user. Is it possible that we've never seen this user before?

src/store/messages/saga.ts
```diff
@@ -90,6 +92,45 @@ const _isActive = (channelId) => (state) => {
...
+}
+
+export function* mapMessageSenders(messages) {
+  if (!featureFlags.enableMatrix) {
+    return;
+  }
+
+  const zeroUsersMap = yield call(getZeroUsersMap);
+  messages.forEach((message) => {
+    const zeroUser = zeroUsersMap[message.sender.userId];
```
dalefukami, line 121: Does `sender` have the `matrixId`? If so, I think it makes sense to use that here rather than the `userId`. I know they're currently the same but `matrixId` is more clear.

src/store/messages/saga.ts
ratik21: In most cases we would have the user. The logic being, we first load all the channles + conversations, which has all the "otherMembers" in a channel, which gets normalized and saved in the user state (in redux). After which we fetch the messages in channel, so assuming that a message was sent by someone who is in a channel would be in the `otherMembers` field, which would be cached.

The only case there will be a miss would be when the user has left the channel. I am looking into that currently, since it's required in displaying the message sender's name in the "parent message" as well (related to replies). 

If there's any other case which i have missed then please let me know

src/store/messages/saga.ts
ratik21: its doesn't :(

image

I will add a comment here

PR #1077, opened 2023-10-06
fix: remove edited flag on unedited messages

domw30:
### What does this do?
- removes timestamp set for `updatedAt` in `mapMatrixEventToMessage`.

- to remove the edited flag on unedited messages.

### How do I test this?
- send a message and ensure the edited flag does not appear on the message footer.

#### Things to consider:
  1. Does this change any APIs?

BEFORE

Screenshot 2023-10-06 at 10 47 59

AFTER

Screenshot 2023-10-06 at 10 50 28

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -466,7 +466,7 @@ export class MatrixClient implements IChatClient {
       message,
       isAdmin: false,
       createdAt: timestamp,
-      updatedAt: timestamp,
+      updatedAt: null,
       sender: {
         userId: senderId,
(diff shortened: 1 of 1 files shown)


domw30:
After some investigation, I believe I previously set the `updatedAt` prop with the incorrect timestamp from the message event. @dalefukami - this may cross over with your current investigation with the `createdAt` timestamp. When using the timestamp here, all messages will be marked as `edited`. I'll need to check for a more meaningful way of setting the `updatedAt` timestamp with events.

dalefukami (approved):

PR #1078, opened 2023-10-06
Set the createdAt on rooms

dalefukami:
### What does this do?

Gets the actual room created timestamp when fetching Matrix rooms.

We use createdAt as a fallback ordering in the Conversation list


domw30 (commented):
👍

PR #1079, opened 2023-10-06
Refactor: Simplify current room state getters

dalefukami:
### What does this do?

Small refactoring to simplify a few of the room state getters


domw30 (approved):
Nice tidy up! 👍

PR #1081, opened 2023-10-09
refactor: rename supportsOptimisticSend -> supportsOptimisticCreateConversation

domw30:
### What does this do?
- renames supportsOptimisticSend -> supportsOptimisticCreateConversation.

- both matrix client and sendbird client support optimistic sending, and `supportsOptimisticSend` is now only being used with `createConversation`.  

#### Things to consider:
  1. Does this change any APIs?


ratik21 (approved):
lgtm

PR #1082, opened 2023-10-09
refactor: (matrix-client) replace hardcoded values and use matrix enums

domw30:
### What does this do?
- Tidy up the matrix client, use enums and add membership state enum.
- Ensures consistency and makes use of matrix enums

- Some enums used, some enums not used. Makes sense to be consistent.

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -36,6 +36,12 @@ enum ConnectionStatus {
   Disconnected = 'disconnected',
 }
 
+enum MembershipStateType {
+  Invite = 'invite',
+  Join = 'join',
+  Leave = 'leave',
+}
+
 export class MatrixClient implements IChatClient {
   private matrix: SDKMatrixClient = null;
   private events: RealtimeChatEvents = null;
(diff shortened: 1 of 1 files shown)


src/lib/chat/matrix-client.ts
```diff
@@ -36,6 +36,12 @@ enum ConnectionStatus {
   Disconnected = 'disconnected',
 }
 
+enum MembershipStateType {
+  Invite = 'invite',
+  Join = 'join',
+  Leave = 'leave',
+}
+
```
ratik21, line 44: might be worth moving these to `utils/types.ts`?

ratik21 (approved):

src/lib/chat/matrix-client.ts
ratik21: i think we can just use `chat/matrix/types.ts`. OR `chat/types.ts` works as well, but if we have more types/enums (matrix related), we may need  to eventually move to `chat/matrix/types.ts`...

domw30: Nice sounds good! I added a `chat/matrix/types.ts` 👍

PR #1083, opened 2023-10-09
Handle case if member is not found in `state.normalized.users` while mapping senders

ratik21:
### What does this do?

It's possible that while mapping the senders of a message in a channel, that the user is NOT found in our local cache (i.e `state.normalized.users`). In this case we need to grab it's matrix ID, and call the API to fetch it's profile (as it could be the case that the user is no longer a member of channel).


domw30 (commented):
Nice 👍

PR #1084, opened 2023-10-09
Move `mapMessageSenders` from `messages/saga.ts` -> `messages/utils.matrix.ts`

ratik21:
### What does this do?

Moves a matrix related functionality to `utils.matrix.ts`. Since we use `mapMessageSenders` only in the case if matrix is enabled.


domw30 (commented):
👍

PR #1085, opened 2023-10-09
Matrix: set `parentMessage` while mapping messages which are replies

ratik21:
### What does this do?

Fetches/maps the parentMessage data correctly while parsing all messages fetched after loading a channels/conversation. 
- We first try to fetch the parent message from the list of "already loaded messages"
- if message is not found (i.e it could be an older message), in that case we call the matrix API to fetch the message.

Before:
image

image

After:
image

image

**NOTE**: This does not currently work for when replying to messages in real-time.

### DIFF
--- a/src/lib/chat/index.ts
+++ b/src/lib/chat/index.ts
@@ -38,7 +38,7 @@ export interface IChatClient {
   searchMyNetworksByName: (filter: string) => Promise<MemberNetworks[] | any>;
   searchMentionableUsersForChannel: (channelId: string, search: string, channelMembers?: UserModel[]) => Promise<any[]>;
   getMessagesByChannelId: (channelId: string, lastCreatedAt?: number) => Promise<MessagesResponse>;
-  getMessageByChannelId: (channelId: string, messageId: string) => Promise<any>;
+  getMessageByRoomId: (channelId: string, messageId: string) => Promise<any>;
   createConversation: (
     users: User[],
     name: string,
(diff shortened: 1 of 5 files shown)


src/lib/chat/index.ts
```diff
@@ -38,6 +38,7 @@ export interface IChatClient {
   searchMyNetworksByName: (filter: string) => Promise<MemberNetworks[] | any>;
   searchMentionableUsersForChannel: (channelId: string, search: string, channelMembers?: UserModel[]) => Promise<any[]>;
   getMessagesByChannelId: (channelId: string, lastCreatedAt?: number) => Promise<MessagesResponse>;
+  getMessageByChannelId: (channelId: string, messageId: string) => Promise<any>;
```
dalefukami, line 41: Should we name this `getMessageByRoomId` since it's general for both Channels and Conversations?

src/lib/chat/sendbird-client.ts
```diff
@@ -111,6 +111,10 @@ export class SendbirdClient implements IChatClient {
     return response.body;
   }
 
+  async getMessageByChannelId(_channelId: string, _messageId: string) {
+    return [];
```
dalefukami, line 115: Why an array if this is a fetch for a singular thing?

dalefukami (approved):

ratik21:
> Seems like we should be able to write some unit tests around this stuff?

yes, will do

PR #1092, opened 2023-10-11
Fix: Save new messages to state during `fetch`

ratik21:
### What does this do?

Saves any new message fetched, OR if a an existing message has an updated property (eg. `parentMessage`) to the redux state.


domw30 (commented):
👍 lgtm!

PR #1093, opened 2023-10-11
Refactor: remove `await` from `mapMatrixMessage`

ratik21:
### What does this do?

We don't need to fetch the parent message data during `mapMatrixMessage`, since we already do that while loading new messages.


domw30 (approved):
Nice!! 👍

PR #1096, opened 2023-10-11
Add tests for `mapMessageSenders` in `store/messages/utils.matrix.test.ts`

ratik21:
### What does this do?

Adds test coverage.

### DIFF
--- a/src/store/messages/utils.matrix.ts
+++ b/src/store/messages/utils.matrix.ts
@@ -48,7 +48,7 @@ export function* mapMessageSenders(messages, channelId) {
     }
   }
 
-  if (!matrixIds.length) {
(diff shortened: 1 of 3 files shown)


src/store/messages/utils.matrix.ts
```diff
@@ -48,7 +48,7 @@ export function* mapMessageSenders(messages, channelId) {
     }
   }
 
-  if (!matrixIds.length) {
+  if (matrixIds.length) {
```
ratik21, line 51: I can confirm that an idiot developer wrote this code 🙂

domw30: 😁

domw30 (commented):
Nice

PR #1097, opened 2023-10-12
feat: [matrix] delete a message 

domw30:
### What does this do?
This PR introduces an update to the chat functionality of our application to support the deletion of messages across two different chat clients: Matrix and Sendbird. It achieves the following:

- updated IChatClient interface by abstracting out the deletion method so that irrespective of the underlying chat client, the application can delete messages in a consistent manner.

- implemented the deleteMessageByRoomId method for the Matrix client, utilizing the redactEvent method from the Matrix JS SDK.

- direct implementation within the Sendbird client of the deleteMessage api. This approach ensures that the deletion logic is encapsulated within the chat client itself.

- updated messages/saga.ts function deleteMessage to fetch the active chat client and delegate the delete operation to it.

- filters out redacted messages from the matrix chat fetch results.

- adds test coverage.

- to ensure delete message works with matrix client as it does with sendbird.

### How do I test this?
- open messenger > send a message in a conversation > hover over the sent message and select the three dots menu > click delete message.

#### Things to consider:
  1. Does this change any APIs?

**Delete A Message**

https://github.com/zer0-os/zOS/assets/39112648/bb5e46c7-2cc3-406c-b0e8-424075ec3756

### DIFF
--- a/src/lib/chat/index.ts
+++ b/src/lib/chat/index.ts
@@ -11,7 +11,7 @@ export interface RealtimeChatEvents {
   reconnectStart: () => void;
   reconnectStop: () => void;
   receiveNewMessage: (channelId: string, message: Message) => void;
-  receiveDeleteMessage: (channelId: string, messageId: number) => void;
+  receiveDeleteMessage: (roomId: string, messageId: string) => void;
   onMessageUpdated: (channelId: string, message: Message) => void;
   receiveUnreadCount: (channelId: string, unreadCount: number) => void;
   onUserReceivedInvitation: (channel) => void;
@@ -54,7 +54,7 @@ export interface IChatClient {
     optimisticId?: string
   ) => Promise<MessagesResponse>;
   fetchConversationsWithUsers: (users: User[]) => Promise<Partial<Channel>[]>;
-  deleteMessageByRoomId: (channelId: string, messageId: number) => Promise<any>;
+  deleteMessageByRoomId: (roomId: string, messageId: string) => Promise<void>;
 }
 
 export class Chat {
@@ -90,8 +90,8 @@ export class Chat {
     return this.client.createConversation(users, name, image, optimisticId);
   }
 
-  async deleteMessageByRoomId(channelId: string, messageId: number): Promise<any> {
-    return this.client.deleteMessageByRoomId(channelId, messageId);
+  async deleteMessageByRoomId(roomId: string, messageId: string): Promise<void> {
+    return this.client.deleteMessageByRoomId(roomId, messageId);
   }
 
   async searchMyNetworksByName(filter: string) {
--- a/src/lib/chat/index.ts
+++ b/src/lib/chat/index.ts
@@ -54,6 +54,7 @@ export interface IChatClient {
     optimisticId?: string
   ) => Promise<MessagesResponse>;
   fetchConversationsWithUsers: (users: User[]) => Promise<Partial<Channel>[]>;
+  deleteMessageByRoomId: (channelId: string, messageId: number) => Promise<void>;
 }
 
 export class Chat {
@@ -87,6 +88,10 @@ export class Chat {
 
   async createConversation(users: User[], name: string, image: File, optimisticId: string) {
     return this.client.createConversation(users, name, image, optimisticId);
+  }
+
+  async deleteMessageByRoomId(channelId: string, messageId: number): Promise<void> {
+    return this.client.deleteMessageByRoomId(channelId, messageId);
   }
 
   async searchMyNetworksByName(filter: string) {
--- a/src/lib/chat/index.ts
+++ b/src/lib/chat/index.ts
@@ -54,7 +54,7 @@ export interface IChatClient {
     optimisticId?: string
   ) => Promise<MessagesResponse>;
   fetchConversationsWithUsers: (users: User[]) => Promise<Partial<Channel>[]>;
-  deleteMessageByRoomId: (channelId: string, messageId: number) => Promise<void>;
+  deleteMessageByRoomId: (channelId: string, messageId: number) => Promise<any>;
 }
 
 export class Chat {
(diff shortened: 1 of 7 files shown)


src/lib/chat/index.ts
```diff
@@ -54,6 +54,7 @@ export interface IChatClient {
     optimisticId?: string
   ) => Promise<MessagesResponse>;
   fetchConversationsWithUsers: (users: User[]) => Promise<Partial<Channel>[]>;
+  deleteMessageByRoomId: (channelId: string, messageId: number) => Promise<any>;
```
dalefukami, line 57: 1. The argument should probably be `roomId` :)
2. `messageId` should be a string. They aren't numbers.
3. Return type should probably be `Promise<void>` if the return value is never used.

src/lib/chat/matrix-client.ts
```diff
@@ -133,7 +133,7 @@ export class MatrixClient implements IChatClient {
   async getMessagesByChannelId(channelId: string, _lastCreatedAt?: number): Promise<MessagesResponse> {
     await this.waitForConnection();
     const { chunk } = await this.matrix.createMessagesRequest(channelId, null, 50, Direction.Backward);
-    const messages = chunk.filter((m) => m.type === EventType.RoomMessage);
+    const messages = chunk.filter((m) => m.type === EventType.RoomMessage && !m?.unsigned?.redacted_because);
```
dalefukami, line 136: Might be easier to understand if we extract the `m?.unsigned?.redacted_because` to a function `!isDeleted(m)`

src/lib/chat/sendbird-client.ts
```diff
@@ -101,6 +101,14 @@ export class SendbirdClient implements IChatClient {
     return results.map((u) => ({ id: u.id, display: u.name, profileImage: u.profileImage }));
   }
 
+  async deleteMessageByRoomId(channelId: string, messageId: number): Promise<any> {
+    try {
+      await del<any>(`/chatChannels/${channelId}/message`).send({ message: { id: messageId } });
+    } catch (error: any) {
+      console.log('Error occurred while deleting message', error?.response ?? error);
+    }
```
dalefukami, line 109: It's not a good idea to just swallow the error and do nothing. We should let this throw so that other code can decide what to do.

dalefukami (approved):
Just minor thoughts

src/lib/chat/index.ts
domw30: Nice will update.

Changing `messageId` to `string` will mean I need to update the following `messageId` to a string in the sendbird client.

```
   onMessageDeleted: (channel, messageId) => {
        const channelId = this.getChannelId(channel);

        events.receiveDeleteMessage(channelId, messageId.toString());
      },
```

domw30: Actually I'm confused here. Isn't the sendbird messageId type `number` ?

The api looked like this:
```
export async function deleteMessageApi(channelId: string, messageId: number): Promise<number> {
  const response = await del<any>(`/chatChannels/${channelId}/message`).send({ message: { id: messageId } });

  return response.status;
}
```

then in the `deleteMessage` saga we're changing to string:
```
const messageIdsToDelete = fullMessages.filter((m) => m.rootMessageId === messageId.toString()).map((m) => m.id); // toString() because message ids are currently a number
```

dalefukami: > Isn't the sendbird messageId type number ?

It is usually a number but we never should have defined it as such. It should always have been a string. Someone messed up the original model so we're stuck with juggling it. Since we're moving towards Matrix we should keep the matrix client correct and fix everything else.

domw30: Sounds good thanks! Ive changed all the types to string

src/store/messages/saga.ts
```diff
@@ -326,7 +326,7 @@ export function* deleteMessage(action) {
   const existingMessageIds = yield select(rawMessagesSelector(channelId));
   const fullMessages = yield select((state) => denormalize(existingMessageIds, state));
 
-  const messageIdsToDelete = fullMessages.filter((m) => m.rootMessageId === messageId.toString()).map((m) => m.id); // toString() because message ids are currently a number
+  const messageIdsToDelete = fullMessages.filter((m) => m.rootMessageId === messageId).map((m) => m.id);
```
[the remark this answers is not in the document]
dalefukami: Having said that, maybe we don't care. lol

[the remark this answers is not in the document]
domw30: Appeared to work when I tried it with sendbird. Just to double check..

I've changed the call in sendbird to not swallow the error, and the type to string (because `deleteMessageByRoomId`'s `messageId` type is now string as suggested :
```
async deleteMessageByRoomId(roomId: string, messageId: string): Promise<void> {
    await del<any>(`/chatChannels/${roomId}/message`).send({ message: { id: messageId } });
  }
```

And, as you commented here, remove the `toString` 🤔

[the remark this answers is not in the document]
dalefukami: I didn't say remove `toString`. I said the interface to the generic chat client call should take a string and not a number.

src/lib/chat/sendbird-client.ts
domw30: Updated

src/store/messages/saga.ts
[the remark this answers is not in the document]
dalefukami: Sorry, my point here was...there's a reason that saga code had the `toString` in that filter. It has nothing to do with the api call. That filter is figuring out which messages to delete. If you remove that `toString` there then it won't filter correctly.

src/lib/chat/matrix-client.ts
```diff
@@ -119,6 +119,10 @@ export class MatrixClient implements IChatClient {
     return dmConversationIds.includes(room.roomId) || !!room.getDMInviter();
   };
 
+  private isDeleted(m) {
```
dalefukami, line 127: Probably should use a better variable name here. What is "m"?

src/store/messages/saga.ts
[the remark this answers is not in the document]
dalefukami: In the end... it doesn't matter. We're killing sendbird anyway so just leave it. lol

dalefukami (approved):
Looks good! Thanks for the updates.

src/lib/chat/matrix-client.ts
domw30: Good shout! Updated to `event`

PR #1099, opened 2023-10-13
Fix issue where Matrix id displays in conversation and as conversation name when creating new conversation

ratik21:
**NOTE** - still need to look into the weird state behaviour we're getting with the conversation list while creating a new conversation.

https://github.com/zer0-os/zOS/assets/33264364/0634195a-2887-44d4-b25d-ec925035756d

### DIFF
--- a/src/store/channels-list/saga.ts
+++ b/src/store/channels-list/saga.ts
@@ -221,6 +221,10 @@ export function* createOptimisticConversation(userIds: string[], name: string = 
 }
 
 export function* receiveCreatedConversation(conversation, optimisticConversation = { id: '', optimisticId: '' }) {
+  if (!conversation) {
+    return;
+  }
+
   const existingConversationsList = yield select(rawConversationsList());
   const listWithoutOptimistic = existingConversationsList.filter((id) => id !== optimisticConversation.id);
 
@@ -428,12 +432,17 @@ export function* otherUserJoinedChannel(roomId: string, user: User) {
     return;
   }
 
-  // TODO: Fetch user from zOS if we don't know about them yet
+  if (user.userId === user.matrixId) {
+    user = yield select(userByMatrixId, user.matrixId);
+  }
+
   if (!channel.otherMembers.includes(user.userId)) {
+    const otherMembers = [...channel.otherMembers, user];
     yield put(
       receiveChannel({
         id: channel.id,
-        otherMembers: [...channel.otherMembers, user],
+        isOneOnOne: otherMembers.length === 1,
+        otherMembers,
       })
     );
   }
--- a/src/store/channels-list/saga.ts
+++ b/src/store/channels-list/saga.ts
@@ -441,7 +441,7 @@ export function* otherUserJoinedChannel(roomId: string, user: User) {
     yield put(
       receiveChannel({
         id: channel.id,
-        isOneOnOne: otherMembers.length === 1,
+        isOneOnOne: channel.isChannel === true ? false : otherMembers.length === 1,
         otherMembers,
       })
     );
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -410,7 +410,7 @@ export class MatrixClient implements IChatClient {
   }
 
   private async roomCreated(event) {
-    this.events.onUserJoinedChannel(this.mapChannel(this.matrix.getRoom(event.room_id)));
+    this.events.onUserJoinedChannel(this.mapConversation(this.matrix.getRoom(event.room_id)));
   }
 
   private receiveDeleteMessage = (event) => {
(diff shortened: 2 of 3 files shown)


src/store/channels-list/saga.ts
```diff
@@ -409,12 +413,17 @@ export function* otherUserJoinedChannel(roomId: string, user: User) {
...
+    user = yield select(userByMatrixId, user.matrixId);
+  }
+
   if (!channel.otherMembers.includes(user.userId)) {
+    const otherMembers = [...channel.otherMembers, user];
     yield put(
       receiveChannel({
         id: channel.id,
-        otherMembers: [...channel.otherMembers, user],
+        isOneOnOne: otherMembers.length === 1,
```
dalefukami, line 425: This is slightly incorrect. If it's a `Channel` and not a `Conversation` then it's not considered a One on One.

dalefukami: Upon second thought, I do think at this level this is actually just wrong. This would consider old group conversations where everyone but 2 people left as one on ones and that's not right. Those are still group conversations. This field should be provided by the matrix client where it knows the details of these types of things. (you can see where isOneOnOne is calculated when loading rooms)

src/store/channels-list/saga.ts
```diff
@@ -409,12 +413,17 @@ export function* otherUserJoinedChannel(roomId: string, user: User) {
     return;
   }
 
-  // TODO: Fetch user from zOS if we don't know about them yet
+  if (user.userId === user.matrixId) {
+    user = yield select(userByMatrixId, user.matrixId);
```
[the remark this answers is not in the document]
ratik21: Yes, it works. Tested with users on separate chrome profiles. 

Actually i don't think "this" code works, it will return just the matrix user. But later we're just loading the conversations (for the other user of create conversation), probably because of this event:

https://github.com/zer0-os/zOS/blob/d61983e31b96eb240bada93dddac6f29d63138d9/src/store/channels-list/saga.ts#L327

If we remove this event, then we'll need to fetch the user from the API. But right now it just works.

src/store/channels-list/saga.ts
ratik21: > This is slightly incorrect. If it's a Channel and not a Conversation then it's not considered a One on One.

will update this to only set `isOneOnOne` in case if `isChannel` is `false`.

ratik21: > This field should be provided by the matrix client where it knows the details of these types of things. (you can see where isOneOnOne is calculated when loading conversations)

This is where i am confused. We don't "load" the conversations while (or after) creating the conversation. And while creating the conversation immediately we just have the raw conversation with 0 members. Then we get the events that a member has joined channel, and an "other" member has joined channel. 

The only place we calculate the conversation's `isOneOnOne` is while mapping them (which happens while loading all conversations). Which we don't do here. I guess we could "force" the conversation list to reload here as well to get proper values. But i guess since it's all event based, i don't know when the events will "stop" (after the 1st user joins, 2nd..or 5th? ..etc). If we reload the conversation list sooner than the events haven't stopped coming through, then we'll still end up with a weird state :(

ratik21: re: group conversation, i think we shouldn't have an issue with members leaving the channel (as in that case only the LEAVE event will be triggered, not this one). But i don't know about the case if i start a group conversation with 3 people, in that case this might be wrong. I will test that also

ratik21: create group conversation:

https://github.com/zer0-os/zOS/assets/33264364/6acb6381-43b1-4e91-9679-1c4da3ea3dce

ratik21: > This is slightly incorrect. If it's a Channel and not a Conversation then it's not considered a One on One.

> Even there, I think that's wrong. A "Channel" should never be considered one on one.

@dalefukami found something interesting along this note. Currently we're **always** setting `isChannel == true` while creating a conversation. Currently [this](https://github.com/zer0-os/zOS/blob/d61983e31b96eb240bada93dddac6f29d63138d9/src/lib/chat/matrix-client.ts#L413) event is fired when after we create the room, but this maps the channel generally and always sets `isChannel` to be true. So while creating the conversation, at all event levels, we're assuming this is a channel (& not a conversation) 🤔

src/store/channels-list/saga.ts
```diff
@@ -428,12 +432,17 @@ export function* otherUserJoinedChannel(roomId: string, user: User) {
...
+  }
+
   if (!channel.otherMembers.includes(user.userId)) {
+    const otherMembers = [...channel.otherMembers, user];
     yield put(
       receiveChannel({
         id: channel.id,
-        otherMembers: [...channel.otherMembers, user],
+        isOneOnOne: channel.isChannel === true ? false : otherMembers.length === 1,
```
dalefukami, line 444: This is still incorrect. A Conversation is not a "One on One" if it _ever_ had more than two people in it. Here you'd be setting old group conversations to one on ones if everyone else left.

dalefukami (approved):

PR #1101, opened 2023-10-16
feat: fetch the users presence asynchronously (isOnline)

domw30:
### What does this do?
- replaces how we're currently setting the users `isOnline` status by fetching the correct data using matrixs `getPresence` method.

- currently using `getUser`, to set the users online status, this accidentally works as the user will be defaulted to `offline`. Using `getPresence` asynchronously will return the users correct online status. Similar to how we are getting the last seen at data.

### How do I test this?
- open messenger and check users online status in the conversation list.

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -80,10 +80,15 @@ export class MatrixClient implements IChatClient {
 
     try {
       const userPresenceData = await this.matrix.getPresence(userId);
-      const isOnline = userPresenceData?.presence === 'online';
-      const lastSeenAt = userPresenceData?.last_active_ago
-        ? new Date(Date.now() - userPresenceData.last_active_ago).toISOString()
-        : null;
+
+      if (!userPresenceData) {
+        return { lastSeenAt: null, isOnline: false };
+      }
+
+      const { presence, last_active_ago } = userPresenceData;
+
+      const isOnline = presence === 'online';
+      const lastSeenAt = last_active_ago ? new Date(Date.now() - last_active_ago).toISOString() : null;
 
       return { lastSeenAt, isOnline };
     } catch (error) {
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -508,7 +508,7 @@ export class MatrixClient implements IChatClient {
       firstName: user?.displayName,
       lastName: '',
       profileId: '',
-      isOnline: user?.presence === 'online',
+      isOnline: false,
       profileImage: user?.avatarUrl,
       lastSeenAt: '',
     };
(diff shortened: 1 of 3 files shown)


src/lib/chat/matrix-client.ts
```diff
@@ -508,7 +508,7 @@ export class MatrixClient implements IChatClient {
       firstName: user?.displayName,
       lastName: '',
       profileId: '',
-      isOnline: user?.presence === 'online',
+      isOnline: false,
```
dalefukami, line 511: No reason to provide an incorrect value here, I don't think? Just don't return this field.

src/store/channels-list/saga.ts
```diff
@@ -63,19 +63,26 @@ export function* mapToZeroUsers(channels: any[]) {
...
   }
 
   const chatClient = yield call(chat.get);
   for (let conversation of conversations) {
-    const matrixId = conversation?.otherMembers?.[0]?.matrixId;
+    const { isOneOnOne, otherMembers } = conversation;
+    const matrixId = otherMembers?.[0]?.matrixId;
 
-    if (conversation.isOneOnOne && matrixId) {
+    if (isOneOnOne && matrixId) {
```
dalefukami, line 76: Don't we still show the green dot on group conversations?

src/lib/chat/matrix-client.ts
domw30: If removed we will get a type error due to the type being `UserModel`. Should I change the method type?

dalefukami: Oh, it's required? I guess...yeah, just leave this for now.

src/store/channels-list/saga.ts
domw30: Yeah actually I don't think this will won't work for group conversations. I'll need to take another look.

src/store/channels-list/saga.ts
```diff
@@ -63,19 +63,29 @@ export function* mapToZeroUsers(channels: any[]) {
...
-        conversation.otherMembers[0].lastSeenAt = presenceData.lastSeenAt;
+      if (!presenceData) continue;
+
+      const { lastSeenAt, isOnline } = presenceData;
+
+      if (lastSeenAt) {
+        member.lastSeenAt = lastSeenAt;
+      }
+      if (isOnline) {
+        member.isOnline = isOnline;
```
dalefukami, line 88: What if `isOnline` is false? don't we need to set it to false for the user?

src/store/channels-list/saga.ts
domw30: Nice, Updated this!

Screenshot 2023-10-17 at 13 12 44

dalefukami (approved):

PR #1102, opened 2023-10-16
feat: (matrix) edit message

domw30:
### What does this do? 
Introduces the capability to handle and process edited messages. Key changes include:
- Implementing functionality to determine if a message is an edit event.
- Refactoring message handling logic to account for edited messages and update the original message content accordingly.
- Introducing mechanisms to listen for edited message events and respond with the appropriate updates.
- Refactoring related functions for better clarity and reduced code repetition.

To ensure that our matrix client can effectively handle message edits.

### How do I test this?
Open the messenger > Send a message > Click the three dots message menu > Select `Edit` > Edit the message and submit > Check message has been edited.

#### Things to consider:
  1. Does this change any APIs?

NOTE: The flicker/movement of the edited message here will be resolved in a follow up PR.

https://github.com/zer0-os/zOS/assets/39112648/a7dfe201-32a4-4cae-a469-7f939ac1e853

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -607,11 +607,7 @@ export class MatrixClient implements IChatClient {
   }
 
   private getAllMessagesFromRoom(room: Room): any[] {
-    const events = room
-      .getLiveTimeline()
-      .getEvents()
-      .reverse()
-      .map((matrixEvent) => matrixEvent.event);
+    const events = [...room.getLiveTimeline().getEvents()].reverse().map((matrixEvent) => matrixEvent.event);
     return this.processRawEventsToMessages(events);
   }
 
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -368,8 +368,6 @@ export class MatrixClient implements IChatClient {
         console.log('encryped message: ', event);
       }
 
-      this.processMessageEvent(event);
-
       if (event.type === EventType.RoomCreate) {
         this.roomCreated(event);
       }
@@ -379,6 +377,8 @@ export class MatrixClient implements IChatClient {
       if (event.type === EventType.RoomRedaction) {
         this.receiveDeleteMessage(event);
       }
+
+      this.processMessageEvent(event);
     });
     this.matrix.on(RoomMemberEvent.Membership, async (_event, member) => {
       if (member.membership === MembershipStateType.Invite && member.userId === this.userId) {
@@ -391,7 +391,7 @@ export class MatrixClient implements IChatClient {
     this.matrix.on(MatrixEventEvent.Decrypted, async (decryptedEvent: MatrixEvent) => {
       const event = decryptedEvent.getEffectiveEvent();
       if (event.type === EventType.RoomMessage) {
-        this.publishMessageEvent(event);
+        this.processMessageEvent(event);
       }
     });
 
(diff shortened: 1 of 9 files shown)


src/store/messages/saga.ts
```diff
@@ -477,6 +486,12 @@ export function* replaceOptimisticMessage(currentMessages, message) {
 export function* receiveUpdateMessage(action) {
   let { message } = action.payload;
 
+  // Check if the message already exists in the state
```
domw30, line 489: Resolves a race condition where both user action and server event trigger a UI update, causing flickering/edited message renders as received message. Implemented a check against the Redux state to determine if a message has already been updated = preventing a redundant server event from triggering a UI refresh.

src/lib/chat/matrix-client.ts
```diff
@@ -36,6 +36,10 @@ export class MatrixClient implements IChatClient {
   private events: RealtimeChatEvents = null;
   private connectionStatus = ConnectionStatus.Disconnected;
 
+  private static readonly RELATES_TO = 'm.relates_to';
+  private static readonly NEW_CONTENT = 'm.new_content';
+  private static readonly REPLACE = 'm.replace';
```
dalefukami, line 41: Didn't you recently create a Types file where something like this could be put?

src/lib/chat/matrix-client.ts
```diff
@@ -556,12 +610,13 @@ export class MatrixClient implements IChatClient {
...
-    const timeline = room.getLiveTimeline().getEvents();
-    const messages = timeline
-      .filter((event) => event.getType() === EventType.RoomMessage && !event.isRedacted())
-      .map(this.mapMatrixEventToMessage);
-    return messages;
+  private getAllMessagesFromRoom(room: Room): any[] {
+    const events = room
+      .getLiveTimeline()
+      .getEvents()
+      .reverse()
```
dalefukami, line 617: I wonder if this messes up the internal state of the room. Maybe we need to "clone" the array before reversing it?

src/lib/chat/matrix-client.ts
```diff
@@ -281,16 +354,25 @@ export class MatrixClient implements IChatClient {
...
     this.matrix.on('event' as any, async ({ event }) => {
       console.log('event: ', event);
       if (event.type === EventType.RoomEncryption) {
         console.log('encryped message: ', event);
       }
 
-      if (event.type === EventType.RoomMessage) {
-        this.publishMessageEvent(event);
-      }
+      this.processMessageEvent(event);
```
dalefukami, line 375: Did you test this in old unencrypted rooms or in an encrypted one? I think there's a decrypted event that needs to handle this :thinking:

dalefukami (approved):
Just the main question about encrypted messages/rooms

src/lib/chat/matrix-client.ts
domw30: Good shout. Extracted.

src/lib/chat/matrix-client.ts
domw30: Nice yeah to avoid potential side effects, I've cloned/copied the array before reversing it. Thanks!

src/lib/chat/matrix-client.ts
domw30: Correct, thanks for pointing that out. I reused the `processMessageEvent` to handle this.

PR #1103, opened 2023-10-17
fix: prevent double update on message edit

domw30:
### What does this do?
-  Implemented a check against the Redux state to determine if a message has already been updated = preventing a redundant server event from triggering a UI refresh.

- Resolves a race condition where both user action and server event trigger a UI update, causing flickering/edited message renders as received message.

### How do I test this?
- Send message > Edit Message > Check UI is as expected.

#### Things to consider:
  1. Does this change any APIs?
  
  
Before:
  
https://github.com/zer0-os/zOS/assets/39112648/c90bdec4-2f74-4011-b6ce-b427d8318b4c

After:

https://github.com/zer0-os/zOS/assets/39112648/ba901983-4e93-4a4e-8a4f-1d648fdbb61d

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -36,10 +36,6 @@ export class MatrixClient implements IChatClient {
   private events: RealtimeChatEvents = null;
   private connectionStatus = ConnectionStatus.Disconnected;
 
-  private static readonly RELATES_TO = 'm.relates_to';
-  private static readonly NEW_CONTENT = 'm.new_content';
-  private static readonly REPLACE = 'm.replace';
-
   private accessToken: string;
   private userId: string;
 
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -26,7 +26,7 @@ import { ParentMessage, User } from './types';
 import { config } from '../../config';
 import { get } from '../api/rest';
 import { MemberNetworks } from '../../store/users/types';
-import { ConnectionStatus, MembershipStateType } from './matrix/types';
+import { ConnectionStatus, MatrixConstants, MembershipStateType } from './matrix/types';
 import { getFilteredMembersForAutoComplete, setAsDM } from './matrix/utils';
 import { uploadImage } from '../../store/channels-list/api';
 import { SessionStorage } from './session-storage';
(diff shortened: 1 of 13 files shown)


src/store/messages/saga.ts
```diff
@@ -486,6 +486,12 @@ export function* replaceOptimisticMessage(currentMessages, message) {
 export function* receiveUpdateMessage(action) {
   let { message } = action.payload;
 
+  // Check if the message already exists in the state
+  const existingMessage = yield select(messageSelector(message.id));
+  if (existingMessage && existingMessage.message === message.message) {
+    return;
+  }
```
dalefukami, line 493: I'm not sure this is the right solution. A message could be updated for any number of reasons that aren't the message changing.

We should be able to update this message in state and if nothing has changed then it should just work. If that's not working then we're updating too much of the state or we're not providing enough of it in the event.

domw30: Hm yeah looks like that shouldn't be needed. 

The reason for the edited message rendering as a 'received' message is due to the `userId` and `senderId` differing slightly, therefore the `isOwner` boolean is false when determining if sent/edited message is from the current user. Maybe we could have a v.quick call to discuss?

Screenshot 2023-10-18 at 11 41 46

ratik21: i guess this is related to NOT mapping the user correctly maybe, which receiving the edit message event perhaps? (i.e userId/senderId don't match because one of them is zOS user id, and other one is matrixId). Could you please check/confirm?

src/lib/chat/matrix-client.ts
```diff
@@ -36,6 +36,10 @@ export class MatrixClient implements IChatClient {
   private events: RealtimeChatEvents = null;
   private connectionStatus = ConnectionStatus.Disconnected;
 
+  private static readonly RELATES_TO = 'm.relates_to';
+  private static readonly NEW_CONTENT = 'm.new_content';
+  private static readonly REPLACE = 'm.replace';
```
ratik21, line 41: a nit but i think we might want to use these values as an `enum` perhaps, otherwise this list can grow quickly :)

domw30: Ah yes these should have been removed when merging main. They should now be removed :)

domw30:
Closing this pul request in favour of : https://github.com/zer0-os/zOS/pull/1135

PR #1110, opened 2023-10-18
Fix scenario when rendering a reply for a message that hasn't been loaded yet

dalefukami:
### What does this do?

We're making the assumption that the parent message of a reply is already in state but that's not always the case. This provides null handling


domw30 (commented):
👍

PR #1113, opened 2023-10-18
Call the API route from the frontend to track messages sent by user

ratik21:


domw30 (approved):

src/lib/chat/matrix-client.ts
```diff
@@ -266,12 +266,22 @@ export class MatrixClient implements IChatClient {
 
     const messageResult = await this.matrix.sendMessage(channelId, content);
     const newMessage = await this.getMessageByRoomId(channelId, messageResult.event_id);
+    await this.recordMessageSent(channelId, newMessage.createdAt);
```
dalefukami, line 269: Maybe we don't need to `await` here? Fire and forget might be better just to not cause issues with the app if this request ever starts failing.

dalefukami (approved):

src/lib/chat/matrix-client.ts
ratik21: makes sense.

PR #1114, opened 2023-10-18
refactor: silence user presence retrieval errors

domw30:
### What does this do?
- Error handling in `getUserPresence` to silently handle errors without logging them to the console.

- To reduce console noise and ensure graceful degradation in scenarios where user presence data is unavailable or erroneous.

#### Things to consider:
  1. Does this change any APIs?

Before:
Screenshot 2023-10-18 at 21 43 08

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -74,8 +74,23 @@ export class MatrixClient implements IChatClient {
     return this.matrix.getAccountData(eventType);
   }
 
+  private hasSharedJoinedRoom(userId: string): boolean {
+    const rooms = this.matrix.getRooms();
+
+    return !!rooms.find((room) => {
+      const myMember = room.getMember(this.matrix.getUserId());
+      const targetMember = room.getMember(userId);
+
+      return myMember && myMember.membership === 'join' && targetMember && targetMember.membership === 'join';
+    });
+  }
+
   async getUserPresence(userId: string) {
     await this.waitForConnection();
+
+    if (!this.hasSharedJoinedRoom(userId)) {
+      return { lastSeenAt: null, isOnline: false };
+    }
 
     try {
       const userPresenceData = await this.matrix.getPresence(userId);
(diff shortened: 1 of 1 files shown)


src/lib/chat/matrix-client.ts
```diff
@@ -74,9 +74,24 @@ export class MatrixClient implements IChatClient {
...
+      const targetMember = room.getMember(userId);
+
+      return myMember && myMember.membership === 'join' && targetMember && targetMember.membership === 'join';
+    });
+  }
+
   async getUserPresence(userId: string) {
     await this.waitForConnection();
 
+    if (!this.hasSharedJoinedRoom(userId)) {
```
domw30, line 91: Adding this check will silence errors where we previously attempted to fetch presence data for users without a mutual room connection 'joined / accepted'. By ensuring users share a 'joined' status in a room, we reduce unnecessary `getPresence` calls.

While this mitigates the 'FORBIDDEN' error due to the absence of a shared room, the error can still be thrown for other reasons. However, this check serves as a proactive measure to prevent one of the more common causes.


We can remove this check if we don't mind console errors for unsuccessful presence data fetches due to lack of shared rooms. These are those errors:

Screenshot 2023-10-18 at 21 36 40

dalefukami: Yeah, I think  this is a little over complicated just to remove an error from the console. Let's just check for the error code in the catch statement and not log the permission error.

domw30: Okay yeah sounds good, shall update

src/lib/chat/matrix-client.ts
```diff
@@ -85,13 +85,15 @@ export class MatrixClient implements IChatClient {
...
       const { presence, last_active_ago } = userPresenceData;
-
       const isOnline = presence === 'online';
       const lastSeenAt = last_active_ago ? new Date(Date.now() - last_active_ago).toISOString() : null;
 
       return { lastSeenAt, isOnline };
-    } catch (error) {
-      console.error(error);
+    } catch (error: any) {
+      if (error && typeof error === 'object' && error.errcode !== 'M_FORBIDDEN') {
```
dalefukami, line 93: If an error throws I don't think it can ever be null. This would be sufficient:
```typescript
if (error.errcode !== 'M_FORBIDDEN') {
```

PR #1134, opened 2023-10-20
feat: remove matrix feature flag check from syncRewardsAndTokenPrice

domw30:
### What does this do?
- Removes the condition that checks the enableMatrix feature flag within the syncRewardsAndTokenPrice function.

- To enable the syncing of rewards when using matrix, ensuring it only depends on the enableRewards flag.

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/store/rewards/saga.ts
+++ b/src/store/rewards/saga.ts
@@ -64,7 +64,7 @@ export function* syncZEROPrice() {
 }
 
 export function* syncRewardsAndTokenPrice() {
-  if (!featureFlags.enableRewards || featureFlags.enableMatrix) {
+  if (!featureFlags.enableRewards) {
     return;
   }
 


src/store/rewards/saga.ts
```diff
@@ -64,7 +64,7 @@ export function* syncZEROPrice() {
 }
 
 export function* syncRewardsAndTokenPrice() {
-  if (!featureFlags.enableRewards || featureFlags.enableMatrix) {
+  if (!featureFlags.enableRewards) {
```
ratik21, line 67: not sure if fetching the token price at regular intervals does anything for us right now, in matrix. since we haven't implemented rewards yet. but your call :)

ratik21 (approved):

domw30:
As Ratik mentioned above, rewards not fully implemented yet. Closing PR as we can remove this when rewards are ready to be turned on

PR #1136, opened 2023-10-23
Update `userByMatrixId` to fetch user from API, if it doesn't exist in redux store

ratik21:
### What does this do?

Updates the `userByMatrixId` function to fetch the ZERO user using the API, if it doesn't exist in store locally. This is particularly helpful on the other end of the "create conversation" (i.e the user who the conversation was started with).

### DIFF
--- a/src/store/users/selectors.ts
+++ b/src/store/users/selectors.ts
@@ -1,5 +1,11 @@
-import { RootState } from '../reducer';
+import { call, select } from 'redux-saga/effects';
+import { getZEROUsers as getZEROUsersAPI } from '../channels-list/api';
 
-export function userByMatrixId(state: RootState, matrixId: string) {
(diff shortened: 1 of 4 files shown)


src/store/users/selectors.ts
```diff
@@ -1,5 +1,11 @@
-import { RootState } from '../reducer';
+import { call, select } from 'redux-saga/effects';
+import { getZEROUsers as getZEROUsersAPI } from '../channels-list/api';
 
-export function userByMatrixId(state: RootState, matrixId: string) {
-  return Object.values(state.normalized['users'] || {}).find((u: any) => u.matrixId === matrixId);
+export function* userByMatrixId(matrixId: string) {
```
dalefukami, line 4: I don't think we should change this selector. It should remain strictly a selector. If we need something that reaches out to multiple places we can write a helper  function to use this and then fall back to an api request.

ratik21: will update

PR #1137, opened 2023-10-23
Set the active conversation as the newly created one

ratik21:
### What does this do?

- returns the created conversation data "after" creating the conversation
- this also sets the default conversation as the newly created one. 

https://github.com/zer0-os/zOS/assets/33264364/bddeff28-87ad-402c-835b-399b22baef0c


domw30 (approved):
💯

PR #1138, opened 2023-10-23
remove `ConversationListChanged` event (fired when the `AccountData` is updated)

ratik21:
### What does this do?

Removes `conversationListChanged` event, which fires up when `AccountData` event is emitted from matrix.

There are some conversations which aren't getting set in the `accountData` while creating a new conversation. So when we get the "conversationListChanged" event at the end of the create conversation flow ([here](https://github.com/zer0-os/zOS/blob/main/src/lib/chat/matrix-client.ts#L524)), it is missing one (or maybe more) conversationId's from the list. So that conversation which is missed, is being marked as a channel ([here](https://github.com/zer0-os/zOS/blob/main/src/store/channels-list/saga.ts#L422)), due to which it disappears from the list.

Before:

https://github.com/zer0-os/zOS/assets/33264364/e7851523-19d1-432e-a2cd-fe6aae9e75ab

After:

https://github.com/zer0-os/zOS/assets/33264364/cca12624-470f-443a-bdc5-4a967ef52c60

### DIFF
--- a/src/lib/chat/index.ts
+++ b/src/lib/chat/index.ts
@@ -18,7 +18,6 @@ export interface RealtimeChatEvents {
   onUserJoinedChannel: (conversation) => void;
   invalidChatAccessToken: () => void;
   onUserLeft: (channelId: string, userId: string) => void;
-  onConversationListChanged: (conversationIds: string[]) => void;
   onUserPresenceChanged: (matrixId: string, isOnline: boolean, lastSeenAt: string) => void;
   onRoomNameChanged: (channelId: string, name: string) => void;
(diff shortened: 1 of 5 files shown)


dalefukami:
I'm having a hard time figuring out how things still work with this removed. :thinking: If you create a conversation and then refresh the browser then we probably won't render the conversation properly because it won't react to the AccountData change event and put it in the list.

I think the bigger question is, what are we doing wrong with the event that's causing it to not have all the appropriate data?

ratik21:
closing this PR, as this is resolved by [https://github.com/zer0-os/zOS/pull/1140](https://github.com/zer0-os/zOS/pull/1140)

PR #1139, opened 2023-10-24
feat: implement unread message count functionality with matrix client

domw30:
### What does this do?
- This PR implements functionality to track and update unread message counts in the application using the Matrix client. It refactors the existing event handling logic to streamline the process of receiving new messages and updates the unread count accordingly. 

- To ensure conversations are highlighted when there are unread messages, and the unread count is updated and reflected in the UI. 

### How do I test this?
- Send messages to a zOS user you are logged in with and check the conversation list item is highlighted, and, the unread counter in the conversation list item increments for every unread message. Once the conversation has been `read` the conversation list item should no longer be highlighted and a send receipt will be sent to the server which should set the unread count back to 0, in turn the UI should reflect this by no longer highlighting the conversation list item as `unread`.

#### Things to consider:
  1. Does this change any APIs?

DEMO

https://github.com/zer0-os/zOS/assets/39112648/1402902a-4ae5-40c7-a89e-a2400cb8e37f

### DIFF
--- a/src/lib/chat/index.ts
+++ b/src/lib/chat/index.ts
@@ -155,7 +155,7 @@ export class Chat {
   async markRoomAsRead(roomId: string, userId?: string): Promise<void> {
     return this.client.markRoomAsRead(roomId, userId);
   }
-  
+
   async getSecureBackup(): Promise<any> {
     return this.client.getSecureBackup();
   }
--- a/src/lib/chat/index.ts
+++ b/src/lib/chat/index.ts
@@ -62,7 +62,7 @@ export interface IChatClient {
     mentionedUserIds: string[],
     data?: Partial<EditMessageOptions>
   ): Promise<any>;
-  markRoomAsRead: (roomId: string) => Promise<void>;
+  markRoomAsRead: (roomId: string, userId?: string) => Promise<void>;
 }
 
 export class Chat {
@@ -139,8 +139,8 @@ export class Chat {
     return this.client.fetchConversationsWithUsers(users);
   }
 
-  async markRoomAsRead(roomId: string): Promise<void> {
-    return this.client.markRoomAsRead(roomId);
+  async markRoomAsRead(roomId: string, userId?: string): Promise<void> {
+    return this.client.markRoomAsRead(roomId, userId);
   }
 
   initChat(events: RealtimeChatEvents): void {
--- a/src/lib/chat/index.ts
+++ b/src/lib/chat/index.ts
@@ -62,6 +62,7 @@ export interface IChatClient {
     mentionedUserIds: string[],
     data?: Partial<EditMessageOptions>
   ): Promise<any>;
+  markRoomAsRead: (roomId: string) => Promise<void>;
 }
 
 export class Chat {
@@ -136,6 +137,10 @@ export class Chat {
 
   async fetchConversationsWithUsers(users: User[]): Promise<any[]> {
     return this.client.fetchConversationsWithUsers(users);
+  }
+
+  async markRoomAsRead(roomId: string): Promise<void> {
+    return this.client.markRoomAsRead(roomId);
   }
 
   initChat(events: RealtimeChatEvents): void {
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -513,19 +513,6 @@ export class MatrixClient implements IChatClient {
 
       await this.waitForSync();
 
-      this.matrix.once('sync' as any, (state) => {
-        if (state === 'PREPARED' || state === 'SYNCING') {
-          let rooms = this.matrix.getRooms();
-          rooms.forEach((room) => {
-            room.on(RoomEvent.UnreadNotifications, (unreadNotifications) => {
-              console.log('Unread notifications for room:', room.roomId, 'UNREAD:', unreadNotifications);
-
-              this.events.receiveUnreadCount(room.roomId, unreadNotifications?.total || 0);
-            });
-          });
-        }
-      });
-
       return opts.userId;
     }
   }
@@ -605,7 +592,7 @@ export class MatrixClient implements IChatClient {
     const avatarUrl = this.getRoomAvatar(room);
     const createdAt = this.getRoomCreatedAt(room);
     const messages = this.getAllMessagesFromRoom(room);
-    const unreadCount = room.getRoomUnreadNotificationCount(NotificationCountType.Total);
+    const unreadCount = room.getUnreadNotificationCount(NotificationCountType.Total);
 
     return {
       id: room.roomId,
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -267,7 +267,9 @@ export class MatrixClient implements IChatClient {
     // Any room is only set as a DM based on a single user. We'll use the first one.
     await setAsDM(this.matrix, result.room_id, users[0].matrixId);
 
-    return await this.mapConversation(this.matrix.getRoom(result.room_id));
+    const room = this.matrix.getRoom(result.room_id);
+    this.initializeRoomEventHandlers(room);
+    return await this.mapConversation(room);
   }
 
   async sendMessagesByChannelId(
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -124,9 +124,23 @@ export class MatrixClient implements IChatClient {
           failedToJoin.push(room.roomId);
         }
       }
+      // figure out where to register the event
+      if (room.roomId === '!gYnYrvCHcXMcoJkJUn:zero-synapse-development.zer0.io') {
+        room.on(RoomEvent.UnreadNotifications, (event) => this.logRoom(room));
+        room.on(RoomEvent.Receipt, (event) => this.logRoom(room));
+        room.on(RoomEvent.MyMembership, (event) => this.logRoom(room));
+        room.on(RoomEvent.LocalEchoUpdated, (event) => this.logRoom(room));
+        room.on(RoomEvent.Timeline, (event) => this.logRoom(room));
+        room.on(RoomEvent.Redaction, (event) => this.logRoom(room));
+      }
     }
 
     return rooms.filter((r) => !failedToJoin.includes(r.roomId)).map(this.mapConversation);
+  }
+
+  private logRoom(room) {
+    console.log('GET_COVNVO_UNREAD-TOTAL', room.getRoomUnreadNotificationCount(NotificationCountType.Total));
+    console.log('GET_COVNVO_UNREAD-HIGHLIGHT', room.getRoomUnreadNotificationCount(NotificationCountType.Highlight));
   }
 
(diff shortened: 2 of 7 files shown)


src/lib/chat/matrix-client.ts
```diff
@@ -518,6 +541,16 @@ export class MatrixClient implements IChatClient {
...
 
   private publishMessageEvent(event) {
+    const room = this.matrix.getRoom(event.room_id);
+
+    if (!room) {
+      return;
+    }
+
+    if (event.sender !== this.userId) {
+      this.events.receiveUnreadCount(event.room_id, room.getUnreadNotificationCount(NotificationCountType.Total));
```
domw30, line 551: Note : think this might have to move

dalefukami: Yeah, this doesn't seem like the right place for this stuff. Receiving a message has nothing to do with unread counts.

src/store/channels/saga.ts
```diff
@@ -61,6 +62,8 @@ export function* markConversationAsRead(conversationId) {
   const conversationInfo = yield select(rawChannelSelector(conversationId));
   if (conversationInfo?.unreadCount > 0) {
     yield call(markAllMessagesAsRead, conversationId, currentUser.id);
+    const chatClient = yield call(chat.get);
+    yield call([chatClient, chatClient.markRoomAsRead], conversationId);
```
dalefukami, line 66: Wouldn't the intent be that this call replaces the one just above? Why would we have both?

domw30: I've moved this to a more appropriate place. `markChannelsAsRead` and `markConversationAsRead` haven't needed any updates, instead I have updated `markAllMessagesAsRead` to handle/use the chatClient

src/lib/chat/matrix-client.ts
```diff
@@ -313,6 +314,28 @@ export class MatrixClient implements IChatClient {
...
+
+    const events = room.getLiveTimeline().getEvents();
+    const latestEvent = events[events.length - 1];
+
+    if (!latestEvent) {
+      return;
+    }
+
+    try {
+      await this.matrix.sendReadReceipt(latestEvent);
```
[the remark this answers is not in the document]
dalefukami: Ok, gotcha. Are the read receipts public? As in, can anyone in the room see who has read a message? Wonder if there's a setting to disable that?...

src/store/channels/saga.ts
```diff
@@ -32,7 +32,8 @@ export function* markAllMessagesAsRead(channelId, userId) {
     return;
   }
 
-  const status = yield call(markAllMessagesAsReadAPI, channelId, userId);
+  const chatClient = yield call(chat.get);
+  const status = yield call([chatClient, chatClient.markRoomAsRead], channelId, userId);
 
   if (status === 200) {
```
dalefukami, line 38: Seems like we need to generalize this status check now? Maybe the `markRoomAsRead` should return a boolean rather than http status code?

src/lib/chat/index.ts
```diff
@@ -138,6 +139,10 @@ export class Chat {
     return this.client.fetchConversationsWithUsers(users);
   }
 
+  async markRoomAsRead(roomId: string, userId?: string): Promise<void> {
```
ratik21, line 142: why is the `userId` optional here?

src/lib/chat/index.ts
```diff
@@ -138,6 +139,10 @@ export class Chat {
...
+    return this.client.markRoomAsRead(roomId, userId);
```
ratik21, line 143: if we're not `awaiting` then we can remove `async` from this function

src/lib/chat/matrix-client.ts
```diff
@@ -313,6 +314,28 @@ export class MatrixClient implements IChatClient {
...
 
+  async markRoomAsRead(roomId: string): Promise<void> {
+    const room = this.matrix.getRoom(roomId);
+
+    if (!room) {
+      return;
+    }
+
+    const events = room.getLiveTimeline().getEvents();
+    const latestEvent = events[events.length - 1];
```
ratik21, line 325: just wondering if we need to load all the events? Is there a method the fetch the latest event directly?

src/lib/chat/matrix-client.ts
```diff
@@ -313,6 +314,28 @@ export class MatrixClient implements IChatClient {
...
+
+    if (!latestEvent) {
+      return;
+    }
+
+    try {
+      await this.matrix.sendReadReceipt(latestEvent);
+      await this.matrix.setRoomReadMarkers(roomId, latestEvent.event.event_id);
+    } catch (error) {
+      console.error(`Failed to mark room with ID: ${roomId} as read. Error:`, error);
```
ratik21, line 335: just wondering, does this go to sentry right now? since we're doing most of this stuff from the frontend now (instead of the zero-api). cc: @dalefukami

src/lib/chat/matrix-client.ts
```diff
@@ -581,7 +614,7 @@ export class MatrixClient implements IChatClient {
       messages,
       groupChannelType: GroupChannelType.Private,
       category: '',
-      unreadCount: 0,
+      unreadCount: unreadCount,
```
ratik21, line 617:
```suggestion
      unreadCount,
```

src/lib/chat/index.ts
domw30: The Id is required in the sendbird client method.

src/lib/chat/index.ts
domw30: Yeah, `async` is required here

src/lib/chat/matrix-client.ts
domw30: Unfortunately not that I have found. If you come across one let us know 💪

domw30:
I've reverted the additions for listening for the RoomEvent.UnreadNotifications event. Each approach taken caused numerous bugs in the client. Initially I thought we could handle this in the sync, which would receive the received unread count, however, this introduced an array of issues. Not sure how we want to move forward with this, I'm stumped on how to correctly listen for the UnreadNotifications event without introducing these breaking changes.

dalefukami:
> I've reverted the additions for listening for the RoomEvent.UnreadNotifications event.

What was the problem? From our testing that appeared to be the only thing that actually worked consistently.

domw30:
> > I've reverted the additions for listening for the RoomEvent.UnreadNotifications event.
> 
> What was the problem? From our testing that appeared to be the only thing that actually worked consistently.

@dalefukami - Yeah this seemed like the correct room event to listen for. What I've been unable to implement successfully is the correct place to listen for this event and maintain a consistent count. We tested listening for the room event in `getConversations` which gave us the correct count, but as we discussed, not the method we want to be listening for room events and using `this.events.receiveUnreadCount`. The changes I made to listen for this room event in other areas, will return inconsistent unread counts. Reading through some forums, posts, issues etc others suggested listening for the RoomEvent.UnreadNotifications when syncing (which made some sense as this was what contained the most consistent count), but, introducing this listener when syncing caused issues. Let me know your thoughts on the above and I will revisit when back at my laptop this evening.

src/lib/chat/matrix-client.ts
```diff
@@ -118,10 +119,13 @@ export class MatrixClient implements IChatClient {
       await room.decryptAllEvents();
       await room.loadMembersIfNeeded();
       const membership = room.getMyMembership();
-      if (membership === MembershipStateType.Invite) {
-        if (!(await this.autoJoinRoom(room.roomId))) {
-          failedToJoin.push(room.roomId);
-        }
+      if (
```
dalefukami, line 122: 1. What happens if `getConversations` is called again? Do we end up doubling up on the event handlers?
2. We should also do this for Channels, I think. This is not a conversation specific thing.
3. This `if` statement is a weird nested combination of logic and functionality. I think we should probably tidy that up.

src/store/channels/saga.ts
```diff
@@ -31,16 +31,17 @@ export function* markAllMessagesAsRead(channelId, userId) {
...
 
-  if (status === 200) {
     yield put(
       receive({
         id: channelId,
         unreadCount: 0,
       })
     );
-  }
+  } catch (error) {}
```
dalefukami, line 44: It's meaningless at this point because we're killing off Sendbird but the sendbird version of the call returns the http status code (not necessarily throwing an error) so this change isn't backwards compatible.

src/lib/chat/matrix-client.ts
domw30: Hm weird I didn't get an email notification for the review!

re point 2. Yeah I think it makes sense to add to Channels too.
re point 3. Weird, not sure why I added the initialize in an if statement. Reverted and added just in the loop.

domw30: re point 1. How about adding/using the off method before the on method in the `initializeRoomEventHandlers` to safe guard this? We'd ensure that each room will only have one instance of the event handler attached, regardless of how many times getConversations is called. I've pushed this proposed change, let me know what you think @dalefukami

src/lib/chat/matrix-client.ts
```diff
@@ -450,10 +450,15 @@ export class MatrixClient implements IChatClient {
...
-  private initializeRoomEventHandlers(room) {
-    room.on(RoomEvent.UnreadNotifications, (unreadNotifications) => {
+  private initializeRoomEventHandlers(room: Room) {
+    const handleUnreadNotifications = (unreadNotifications) => {
       this.events.receiveUnreadCount(room.roomId, unreadNotifications?.total || 0);
-    });
+    };
+
+    // Removing the event handler if it already exists to prevent duplicates.
+    room.off(RoomEvent.UnreadNotifications, handleUnreadNotifications);
```
dalefukami, line 459: Close but this won't actually work because you're creating a brand new function every time this is called so they won't be detected as the same thing and the `off` won't do anything.

domw30: Thanks for the insight. I'm thinking of mapping event handlers to room IDs to reuse them instead of creating new ones each time. On disconnect, we'd clean up by removing the handlers from the map. Uncertain if this would really work for cleaning them up on disconnect though.

```
  private async initializeRoomEventHandlers(room: Room) {
    console.log(this.eventHandlers);
    const roomId = room.roomId;

    let handleUnreadNotifications = this.eventHandlers.get(roomId);

    if (!handleUnreadNotifications) {
      handleUnreadNotifications = (unreadNotifications) => {
        this.events.receiveUnreadCount(roomId, unreadNotifications?.total || 0);
      };

      this.eventHandlers.set(roomId, handleUnreadNotifications);
    }
   console.log('handleUnreadNotifications', handleUnreadNotifications)
    if (!room.listenerCount(RoomEvent.UnreadNotifications)) {
      room.on(RoomEvent.UnreadNotifications, handleUnreadNotifications);
    }
  }
  ```

dalefukami: In the end, I just checked the cache to see if it already existed: https://github.com/zer0-os/zOS/pull/1139/commits/249c9b81ee2f921550c3b5ff3f2ee757407f701e

PR #1144, opened 2023-10-27
Batch the processing of new message received events

dalefukami:
### What does this do?

Batches the saga processing of new message events

When loading messages from Matrix we get them as individual "decrypted" events. This means we publish the "new message" event multiple times. If you load a room with 100s of messages this causes multiple invocations of the saga which was causing lots of rerenders.

Batching the processing allows for fewer rerenders. Simplified timing results when loading and rendering a conversation with around 130 messages:
Before - 9 to 10 seconds
After - 800ms

### How do I test this?

Ensure you can load messages in a conversation/channel and verify that receiving real time messges still works.

### DIFF
--- a/src/store/messages/saga.ts
+++ b/src/store/messages/saga.ts
@@ -440,6 +440,10 @@ export function* receiveNewMessage(action) {
   // Clone and empty so follow up events can debounce again
   const batchedPayloads = [...savedMessages];
   savedMessages = [];
+  return yield call(batchedReceiveNewMessage, batchedPayloads);
+}
+
+export function* batchedReceiveNewMessage(batchedPayloads) {
   const byChannelId = {};
   batchedPayloads.forEach((m) => {
     byChannelId[m.channelId] = byChannelId[m.channelId] || [];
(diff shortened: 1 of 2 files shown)


src/store/messages/saga.ts
```diff
@@ -429,38 +429,67 @@ export function* receiveDelete(action) {
...
-  yield spawn(sendBrowserNotification, channelId, message);
-
-  if (yield select(_isActive(channelId))) {
-    const isChannel = yield select(_isChannel(channelId));
-    const markAllAsReadAction = isChannel ? markChannelAsRead : markConversationAsRead;
-    yield spawn(markAllAsReadAction, channelId);
-  }
+  // Since the conversation load happens via these events now, this can't just happen. We need to figure out if this is a bullk load or a
+  // "fresh" message
+  // yield spawn(sendBrowserNotification, channelId, message);
```
dalefukami, line 492: Left this here since I know we'll be looking at browser notifications soon and wanted to leave a breadcrumb for whoever steps into it.

PR #1147, opened 2023-10-31
feat: conversation creation with inviter on registration completion

domw30:
### What does this do?
- fetches the `inviter` data from the registration and creates a conversation between the two users.

- following the behaviour when using the sendbird client, the changes are to automatically connect new users with their inviters by creating a conversation when completing registration and logging into the app.

### How do I test this?
- get an invite code from your current user > go to the `/get-access` route and enter the code > complete registration > when logged in, check a conversation has been created and appears in your conversation list with the inviter.

#### Things to consider:
  1. Does this change any APIs?

Join Demo

https://github.com/zer0-os/zOS/assets/39112648/a5c5392f-fa96-4aa7-9959-7ca79fd5de41


src/store/channels-list/saga.ts
```diff
@@ -146,7 +146,13 @@ export function userSelector(state, userIds) {
   return userIds.map((id) => (state.normalized.users || {})[id]);
 }
 
-export function* createConversation(userIds: string[], name: string = null, image: File = null) {
+export function* createConversation(
+  userIds: string[],
+  name: string = null,
+  image: File = null,
+  adminMessageType?: AdminMessageType,
```
domw30, line 153: We can make this prop required in the follow up PR

src/lib/chat/index.ts
```diff
@@ -43,7 +43,9 @@ export interface IChatClient {
     users: User[],
     name: string,
     image: File,
-    optimisticId: string
+    optimisticId: string,
+    adminMessageType?: AdminMessageType,
```
domw30, line 47: We can make these props required in the follow up PR

src/lib/chat/index.ts
```diff
@@ -97,8 +99,15 @@ export class Chat {
...
   }
 
-  async createConversation(users: User[], name: string, image: File, optimisticId: string) {
-    return this.client.createConversation(users, name, image, optimisticId);
+  async createConversation(
+    users: User[],
+    name: string,
+    image: File,
+    optimisticId: string,
+    adminMessageType?: AdminMessageType,
```
domw30, line 107: We can make this prop required in the follow up PR

src/lib/chat/matrix-client.ts
```diff
@@ -238,12 +248,36 @@ export class MatrixClient implements IChatClient {
...
+  ) {
+    let content: any = {
+      type,
+      body: 'You have joined a conversation.',
+    };
+
+    if (type === AdminMessageType.JOINED_ZERO) {
+      content.inviteeId = currentUserId;
+      content.inviterId = inviterId;
+    } else if (type === AdminMessageType.CONVERSATION_STARTED) {
```
dalefukami, line 274: We shouldn't be creating this type of event. We can already get this one from the standard Matrix events.

src/lib/chat/matrix/types.ts
```diff
@@ -15,3 +15,7 @@ export enum MatrixConstants {
   NEW_CONTENT = 'm.new_content',
   REPLACE = 'm.replace',
 }
+
+export enum CustomEventType {
+  SEND_ADMIN_MESSAGE = 'm.custom.send_admin_message',
```
dalefukami, line 20: Is the `m.custom` a standard prefix?

`admin_message` is probably a better event name. "send" sounds like an action that's being performed. Eg: The matrix event is "EventType.RoomMessage" not "EventType.SendRoomMessage"

src/lib/chat/matrix-client.ts
```diff
@@ -238,12 +248,36 @@ export class MatrixClient implements IChatClient {
     }
 
     const result = await this.matrix.createRoom(options);
+
     // Any room is only set as a DM based on a single user. We'll use the first one.
     await setAsDM(this.matrix, result.room_id, users[0].matrixId);
 
+    await this.sendAdminMessage(result.room_id, users[0].userId, currentUserId, adminMessageType);
```
dalefukami, line 255: I don't really think this should be tied up in the matrix-client. Sending this welcome message is business logic and should be handled in the sagas

src/lib/chat/matrix/types.ts
domw30: This was just something I made up, not a standard prefix. Updated to `admin_message`.

src/store/registration/saga.ts
```diff
@@ -169,6 +172,30 @@ export function* updateProfile(action) {
...
+        return {
+          userId: userData.id,
+          firstName: userData.profileSummary.firstName,
+          profileImage: userData.profileSummary.profileImage,
+          matrixId: userData.matrixId,
+        };
+      };
+
+      if (userData && userData.length > 0) {
+        yield put(receive({ users: { [inviterUserId]: normalizeUserData(userData[0]) } }));
```
dalefukami, line 190: This feels a little off. We shouldn't generally be forcefully "receiving" directly from the normalized data. Pretty sure you want to import `receive` from the `users` store and just receive the user:
`yield put(receive(<user>));`

At least, that's my hunch.

src/store/registration/saga.ts
```diff
@@ -169,6 +172,30 @@ export function* updateProfile(action) {
       yield put(setFirstTimeLogin(true));
       yield call(completeUserLogin);
       yield put(setStage(RegistrationStage.Done));
+
+      const inviterMatrixId = response.response.inviter.matrixId;
+      const inviterUserId = response.response.inviter.id;
+      const userData = yield call(getZEROUsersAPI, [inviterMatrixId]);
+
+      const normalizeUserData = (userData) => {
```
dalefukami, line 180: Does this need to be a function? Can we just create the  user struct right here? We only use it in one place.

Side note: do we truly need this here? Doesn't creating the conversation already go through paths that would fetch the rest of the user data anyway?

src/store/registration/saga.ts
```diff
@@ -169,6 +172,30 @@ export function* updateProfile(action) {
...
+          profileImage: userData.profileSummary.profileImage,
+          matrixId: userData.matrixId,
+        };
+      };
+
+      if (userData && userData.length > 0) {
+        yield put(receive({ users: { [inviterUserId]: normalizeUserData(userData[0]) } }));
+      }
+
+      if (inviterUserId) {
```
dalefukami, line 193: Should this guard clause be higher? If it's empty then we just tried to fetch a user that doesn't exist and put it into the state, I think.

dalefukami (approved):
We'll want to deal with these comments, I think. But if this works for now I'm ok with merging and handling as follow ups.

src/store/registration/saga.ts
domw30: Makes sense - Updated

src/store/registration/saga.ts
domw30: Updated to use a struct rather than function. I believe we do need this here, if removed, there is no user data.

src/store/registration/saga.ts
domw30: Updated this, thanks!

src/store/registration/saga.ts
dalefukami: What happens when a we normally create a new conversation with a user we've never loaded previously? Do we not have code that ends up fetching the user data we need to render the user?

domw30: The user data is available from the store when searching for the user you are about to start a conversation with. Until you search, the user data is empty. That was why I added this here, but maybe there's a better solution?

dalefukami: Sure, but if a different user creates a conversation with you and you receive the real time event we have to fetch the info. I just assumed the conversation handling went through the same mapping. Perhaps it doesn't.

PR #1162, opened 2023-11-08
feat: handle user joined inviter on zero conversation admin message

domw30:
### What does this do?
- handles admin message for the conversation where a user has joined the inviter on zero.

- to add an admin message to the newly created conversation between the current user and the inviter

### How do I test this?
- complete the registration flow using an invite code > in conversation list there should be the newly created conversation with the inviter containing an admin message stating you have joined xyz on zero.

#### Things to consider:
  1. Does this change any APIs?

Screenshot 2023-11-08 at 18 07 50


src/lib/chat/matrix/chat-message.ts
```diff
@@ -46,9 +48,35 @@ export async function mapMatrixMessage(matrixMessage, sdkMatrixClient: SDKMatrix
...
     isAdmin: false,
     optimisticId: content.optimisticId,
-    ...{ mentionedUsers: [], hidePreview: false, media: null, image: null, admin: {} },
+    ...{
+      admin: {},
+      mentionedUsers: [],
+      hidePreview: false,
+      media: null,
+      image: null,
+    },
```
dalefukami, line 57: I don't understand why this is "splatting" an object that we're just creating. Why don't we just put these properties on the object directly?

src/lib/chat/matrix/chat-message.ts
```diff
@@ -46,9 +48,35 @@ export async function mapMatrixMessage(matrixMessage, sdkMatrixClient: SDKMatrix
...
+      ? AdminMessageType.JOINED_ZERO
+      : AdminMessageType.CONVERSATION_STARTED;
+
+  return {
+    id: event_id,
+    message: 'Conversation was started',
+    createdAt: origin_server_ts,
+    isAdmin: true,
+    sender: {
+      userId: senderId,
```
dalefukami, line 78: Do we have to put a sender here? I don't think anyone is the sender of these "messages".

dalefukami (approved):
Definitely need testing around much of this.

src/lib/chat/matrix/chat-message.ts
domw30: This was existing but I will update in the follow up PRs, no reason that I can see for the 'splatting'

src/lib/chat/matrix/chat-message.ts
domw30: Correct yeah no one is the sender but if I do remove this 

Screenshot 2023-11-13 at 17 09 31

Suggesting this would need to be handled in mapChannelMessages. Shall take a look in the follow ups

PR #1179, opened 2023-11-13
feat: handle create room event admin message

domw30:
### What does this do?
- listens for `RoomCreate` events when processing raw events to messages.
- maps `RoomCreate` events to admin messages.
- removes `RoomCreate` admin messages when an admin message with the type `AdminMessageType.JOINED_ZERO` exists. 
- does the necessary admin message filtering when rendering messages.

- to ensure an admin message is mapped when there is a `RoomCreate` event. So when a user starts a conversation with another user, a system message will be rendered.
- to ensure only one system message is rendered depending on the events for that room.

### How do I test this?
- create a conversation with another user > once conversation has been created an admin message should be displayed stating either : `you started a conversation` or `xyz started the conversation`

#### Things to consider:
  1. Does this change any APIs?

Screenshot 2023-11-13 at 17 00 29

https://github.com/zer0-os/zOS/assets/39112648/f5492073-6bf2-47cc-b99b-477e45a4fcd9


dalefukami (approved):

domw30:
@dalefukami - updated to handle the filtering when rendering messages

src/components/chat-view-container/chat-view.tsx
```diff
@@ -201,10 +201,16 @@ export class ChatView extends React.Component<Properties, State> {
 
   renderMessages() {
     const messagesByDay = this.getMessagesByDay();
-    const groupDays = Object.keys(messagesByDay).sort((a, b) => (a > b ? 1 : -1));
+    const filteredMessagesByDay = filterAdminMessages(messagesByDay);
```
dalefukami, line 204: I think it might make more sense to filter this at a higher level component but this is fine for now.

dalefukami (approved):

src/components/chat-view-container/chat-view.tsx
domw30: Yeah if we want to handle at a higher level, the `chat-view-container`, `get messages()` getter may be the place to filter. Shall note to move that. Thanks Dale.

PR #1180, opened 2023-11-13
Open the conversation with a user (if there is one) while searching for users in create-conversation flow

ratik21:
### What does this do?

Before:

https://github.com/zer0-os/zOS/assets/33264364/c761a95f-bd92-43f4-8d51-fa8d59e41913

After:

https://github.com/zer0-os/zOS/assets/33264364/85c119e2-1cd2-416c-b0f1-d86c4d4cc460


src/components/messenger/list/index.tsx
```diff
@@ -155,7 +155,16 @@ export class Container extends React.Component<Properties, State> {
     return users.map((user) => ({ ...user, image: user.profileImage }));
   };
 
+  isUserAlreadyInConversation = (userId: string) => {
+    return this.props.conversations.find((c) => c.otherMembers[0].userId === userId);
```
dalefukami, line 159: Does this make sense? What if the user is in multiple group conversations and just happens to be the first one listed?

ratik21: yes just thought about that before seeing your comment 😅 , i'll filter for only "one2one" convo's

ratik21: updated

src/components/messenger/list/index.tsx
```diff
@@ -156,7 +156,9 @@ export class Container extends React.Component<Properties, State> {
   };
 
   isUserAlreadyInConversation = (userId: string) => {
-    return this.props.conversations.find((c) => c.otherMembers[0].userId === userId);
+    return this.props.conversations
+      .filter((conversation) => conversation.otherMembers.length === 1)
```
dalefukami, line 160: Should use the `isOneOnOne` here. Otherwise a group conversation where everyone who has left except these 2 people could possible be opened and that's not really what the user would want.

ratik21: done

src/components/messenger/list/index.tsx
```diff
@@ -155,7 +155,18 @@ export class Container extends React.Component<Properties, State> {
     return users.map((user) => ({ ...user, image: user.profileImage }));
   };
 
+  isUserAlreadyInConversation = (userId: string) => {
+    return this.props.conversations
+      .filter((conversation) => conversation.isOneOnOne === true)
```
dalefukami, line 160:
```suggestion
      .filter((c) => c.isOneOnOne)
```

`isOneOnOne` is already a boolean value.

dalefukami (approved):

PR #1181, opened 2023-11-14
feat: enhance event timeline processing to include create room event

domw30:
### What does this do?
- This update modifies the getAllMessagesFromRoom method to ensure that room creation events are included when fetching room events. The method now explicitly checks for the existence of a room creation event and adds it to the timeline events.

- Previously, room creation events were sometimes omitted in the event timeline, leading to incomplete room histories when processing raw events to messages. By including room creation events, this change ensures that admin messages will render in the conversation item preview when fetching conversations. (Before and After below).

### How do I test this?
- with another user, create a room with your current user > when the conversation renders in the conversation list panel, you should see the admin message in the preview > when refreshing this should persist. 

#### Things to consider:
  1. Does this change any APIs?

BEFORE:

https://github.com/zer0-os/zOS/assets/39112648/e29065c6-d288-43f4-9ed3-d99b783c9e9f

AFTER:

https://github.com/zer0-os/zOS/assets/39112648/69ad691c-ee7a-48ee-8058-abc45309dcf3


domw30:
Proposed changes to ensure create room events are processed to raw messages for admin messages.

src/lib/chat/matrix-client.ts
```diff
@@ -742,11 +742,44 @@ export class MatrixClient implements IChatClient {
   }
 
   private async getAllMessagesFromRoom(room: Room): Promise<any[]> {
-    const events = room
+    const timelineEvents = this.getTimelineEvents(room);
+    const roomCreationEvent = this.getRoomCreationEvent(room);
```
dalefukami, line 746: I don't think we want to do this. This means that no matter how far up the conversation we are we'll always see this message, right? It should only show at the top which would only be in the timeline if we've loaded all the way to the beginning of it, right?

domw30: I see, yes you are right, we'll always see the message even if we haven't loaded to the beginning of the conversation, which now thinking of it, isn't the desired behaviour as you've mentioned. 

In the before demo above, you see one conversation does not have the admin message, `Matty Test started the conversation`, in the preview when refreshing the page. If I log the live timeline events for the room, `room.create` is not part of the list, because it is a state event. However, when the conversation is selected and active, the `room.create` event is then found in the live timeline events, meaning the event gets processed and mapped to an admin message. This was why I created this PR but as you mentioned, it doesn't quite resolve the issue with the expected behaviour.

domw30:
Outdated PR

PR #1190, opened 2023-11-20
fix: channel list saga type error failure

domw30:
### What does this do?
- fixes a type error in the channel list saga.

- fixing the type error in the channel list saga prevents the saga failing and therefore the names of the user are correctly rendered in the conversation list items. 
- this fix also appears to prevent users from creating more than one conversation with another user.

### How do I test this?
- create a conversation between two users > check name in conversation list item is rendered as expected > check that you can not create another conversation with the same user as they should not appear in the search results.

#### Things to consider:
  1. Does this change any APIs?

BEFORE DEMOS:

UserId in place of User Name when creating a conversation & creating more than 1 conversation with user.

https://github.com/zer0-os/zOS/assets/39112648/3bc92578-dea0-4c66-bcae-667636ded33c

Audio with this demo explaining what we see. Blank conversation names & Creating more than 1 conversation with user.

https://github.com/zer0-os/zOS/assets/39112648/43625e61-fa3a-446f-928c-a804640a3af1

AFTER DEMO:

Audio with this demo explaining what we see.

https://github.com/zer0-os/zOS/assets/39112648/50ee4229-ffcf-417f-9f43-b006a3edadfc

### DIFF
--- a/src/store/channels-list/saga.ts
+++ b/src/store/channels-list/saga.ts
@@ -355,8 +355,7 @@ function* listenForUserLogout() {
   }
 }
 
-export function* currentUserAddedToChannel(action) {
-  console.log('OR HERE', action);
+export function* currentUserAddedToChannel(_action) {
   // For now, just force a fetch of conversations to refresh the list.
   yield fetchConversations();
 }
@@ -407,7 +406,6 @@ export function* saga() {
 }
 
 function* userJoinedChannelAction({ payload }) {
-  console.log('USER JOINED CHANNEL', payload);
   yield addChannel(payload.channel);
 }
 
@@ -420,7 +418,6 @@ function* roomAvatarChangedAction(action) {
 }
 
 function* otherUserJoinedChannelAction({ payload }) {
-  console.log('PAYLOAD', payload);
   yield otherUserJoinedChannel(payload.channelId, payload.user);
 }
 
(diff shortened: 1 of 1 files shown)


src/store/channels-list/saga.ts
```diff
@@ -453,12 +453,12 @@ export function* otherUserJoinedChannel(roomId: string, user: User) {
...
 
-  if (!channel?.otherMembers.includes(user.userId)) {
-    const otherMembers = [...channel.otherMembers, user];
+  if (!channel?.otherMembers?.includes(user.userId)) {
+    const otherMembers = [...(channel?.otherMembers || []), user];
     yield put(
       receiveChannel({
         id: channel.id,
-        isOneOnOne: channel.isChannel === true ? false : otherMembers.length === 1,
+        isOneOnOne: channel?.isChannel === true ? false : otherMembers.length === 1,
```
domw30, line 461: This should be changed to `otherMembers.length === 1` in a follow up.

domw30: Ended up updating this now.

src/store/channels-list/saga.ts
```diff
@@ -453,12 +453,12 @@ export function* otherUserJoinedChannel(roomId: string, user: User) {
     user = yield call(getUserByMatrixId, user.matrixId) || user;
   }
 
-  if (!channel?.otherMembers.includes(user.userId)) {
-    const otherMembers = [...channel.otherMembers, user];
+  if (!channel?.otherMembers?.includes(user.userId)) {
+    const otherMembers = [...(channel?.otherMembers || []), user];
```
dalefukami, line 457:
```suggestion
  const otherMembers = [...(channel?.otherMembers || [])]
  if (!otherMembers.includes(user.userId)) {
    otherMembers.push(user);
```

dalefukami (approved):

PR #1202, opened 2023-11-24
feat: add wallet icons

domw30:
### What does this do?
- Temporary fix to render wallet icons in the wallet select modal.

- Requests for icons from cloudinary are returning a 404. 

### How do I test this?
- Visit the login via web3 wallet page and check icons render correctly.

#### Things to consider:
  1. Does this change any APIs?

Before:

Screenshot 2023-11-24 at 13 36 04

After:

Screenshot 2023-11-24 at 13 36 56


dalefukami (commented):
We should just move this icons to the production cloudinary account rather than the repo.

src/components/wallet-select/wallets.ts
```diff
@@ -16,35 +22,28 @@ const wallets: { [walletType: string]: Wallet } = {
...
     type: WalletType.Portis,
     name: 'Portis',
-    imageSource: 'portis.svg',
+    imageSource: portisIcon,
   },
 };
 
-for (const wallet in wallets) {
-  wallets[wallet].imageSource = [
-    'https://res.cloudinary.com/fact0ry-dev/image/upload/v1637005920/zero-assets/wallet-providers',
```
ratik21, line 45: yes, i think we can just update the url here with the new one (provided we have uploaded the icons there).

domw30: Yup we need to upload the icons. Requested this in the Jira task earlier this week, just waiting for Joel/Dale to have some free time to do so, think other work is higher priority at the mo, but keeping tabs on this.

PR #1203, opened 2023-11-27
fix: handle processing undecryptable message edits

domw30:
### What does this do?
- handles the processing of undecryptable message edits by replacing the message with `Message edit cannot be decrypted.`.

- We currently ignore message edits that are undecryptable. The changes ensure that we display something else rather than show the original message / empty message.

### How do I test this?
- locate a message that has been edited and the `msgtype` is `m.bad.encrypted` > check the message displays: `Message edit cannot be decrypted`

#### Things to consider:
  1. Does this change any APIs?

BEFORE:
Screenshot 2023-11-27 at 12 34 25

Screenshot 2023-11-27 at 12 35 06

AFTER:

Screenshot 2023-11-27 at 12 33 10

Screenshot 2023-11-27 at 12 35 39


src/lib/chat/matrix-client.ts
```diff
@@ -231,13 +231,22 @@ export class MatrixClient implements IChatClient {
       const relatedEventId = this.getRelatedEventId(editEvent);
       const messageIndex = messages.findIndex((msg) => msg.id === relatedEventId);
       if (messageIndex > -1) {
-        const newContent = this.getNewContent(editEvent);
-        if (newContent) {
+        if (editEvent.content.msgtype === MatrixConstants.BAD_ENCRYPTED_MSGTYPE) {
```
dalefukami, line 234: I feel like this whole function is getting out of hand. Let's refactor this before adding new functionalty.

domw30: https://github.com/zer0-os/zOS/pull/1210. I will update this branch with those changes.

PR #1205, opened 2023-11-28
feat: remove url paths (replace zns route connect)

domw30:
### What does this do?
- removes url paths
- replaces zns route connect
- updates routing

- to remove url paths i.e. so the url does not contain `/0.znsRoute/app`

### How do I test this?
- load up messenger and check url > no paths should be added to the url unless the path is for `/login`, `/get-access`, `/reset-password`.

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 
 import ReactDOM from 'react-dom';
-import { ZnsRouteConnect } from './zns-route-connect';
+import { MessengerMain } from './messenger-main';
 import { store, runSagas } from './store';
 import { Provider } from 'react-redux';
 import { EscapeManagerProvider } from '@zer0-os/zos-component-library';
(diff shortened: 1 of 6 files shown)


src/index.tsx
```diff
@@ -50,8 +49,8 @@ ReactDOM.render(
                 <Route path='/get-access' exact component={Invite} />
                 <Route path='/login' exact component={LoginPage} />
                 <Route path='/reset-password' exact component={ResetPassword} />
-                <Route path='/:znsRoute?/' exact render={redirectToDefaults} />
-                <Route path='/:znsRoute/:app' component={ZnsRouteConnect} />
+                <Route path='/' exact component={MessengerMain} />
+                <Route path='/:znsRoute' render={redirectToDefaults} />
```
dalefukami, line 53: Is the `ZnsRouteConnect` component used any more?

domw30: Not used anymore, we don't need the complexity for the setting routes. The component and test file now removed :) If all looks good I'll merge this one in.

src/messenger-main.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
```
ratik21, line 1: This seems like "main.tsx" to me. What's different b/w `main.tsx` & `messenger-main.tsx`? 
Assuming all zOS is, is messenger right now 🤷‍♂️

domw30: Wouldn't `main.tsx` would clash with the existing `Main.tsx` ?

dalefukami (approved):

PR #1206, opened 2023-11-28
Add `admin: <>` key while mapping a conversation in Matrix

ratik21:


src/store/channels/index.ts
```diff
@@ -53,6 +53,7 @@ export interface Channel {
   hasLoadedMessages: boolean;
   conversationStatus: ConversationStatus;
   messagesFetchStatus: MessagesFetchState;
+  admin: string;
```
dalefukami, line 56: Probably could have made this an array: `adminIds: string[]` since we know that ALL admins can do the things we're implementing.

dalefukami: We only have 1 for now. The intent is that there will be many.

PR #1211, opened 2023-11-29
Set power level from `100` -> `0` while loading conversations if *this* user is not the creator/admin

ratik21:
### What does this do?

Lowers the power level of the current user to 0, for all conversation + groups if the current user is NOT the admin/creator.

Currently all users in an "already exisiting" conversation group has a power_level of 100. This is not ideal.


src/lib/chat/matrix-client.ts
```diff
@@ -104,6 +104,27 @@ export class MatrixClient implements IChatClient {
...
+  // and we create a new room with the appropriate power_levels now.
+  private async setPowerLevels(rooms: Room[]) {
+    for (const room of rooms) {
+      const powerLevels = this.getLatestEvent(room, EventType.RoomPowerLevels);
+      if (!powerLevels || this.userId === room.getCreator()) {
+        continue;
+      }
+
+      const powerLevelsByUser = powerLevels.getContent()?.users || {};
+      for (const userId of Object.keys(powerLevelsByUser)) {
```
dalefukami, line 118: Since we're only doing the current user can we just get their record rather than loop?
`powerLevelsByUser[this.userId]`?

src/lib/chat/matrix-client.ts
```diff
@@ -104,6 +104,27 @@ export class MatrixClient implements IChatClient {
     return [];
   }
 
+  // NOTE: This can be removed in the next release cycle,
```
dalefukami, line 107: Probably not the "next" release cycle but definitely after a couple of weeks.

dalefukami (approved):

src/lib/chat/matrix-client.ts
ratik21: oh yes you're right. thanks for spotting!

PR #1212, opened 2023-11-30
feat: enable add member menu option exclusively for room admins

domw30:
### What does this do?
- Adds ‘Add Member’ as an option list to the triple dot menu, such that ONLY ROOM ADMINS can see it.

- So room admins are able to start the add member flow by selecting `Add Member` from the group management dropdown menu.

### How do I test this?
- `enableGroupManagementMenu` > `enableAddMemberToGroup` >  Create a room > Click three dot menu > Check if `Add Member` menu item renders > Click `Add Member` > Check console log for `group management: starting add group member`.

#### Things to consider:
  1. Does this change any APIs?

Add Group Member Menu Item Exclusively For Room Admin:

https://github.com/zer0-os/zOS/assets/39112648/d421a6e5-e93a-452a-90f3-493b5c998c1b

### DIFF
--- a/src/components/messenger/chat/index.tsx
+++ b/src/components/messenger/chat/index.tsx
@@ -11,11 +11,11 @@ import { getProvider } from '../../../lib/cloudinary/provider';
 import { otherMembersToString } from '../../../platform-apps/channels/util';
 import { GroupManagementMenu } from '../../group-management-menu';
 import { featureFlags } from '../../../lib/feature-flags';
-
-import './styles.scss';
 import { enterFullScreenMessenger, exitFullScreenMessenger } from '../../../store/layout';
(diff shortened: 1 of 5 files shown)


src/components/messenger/chat/index.tsx
```diff
@@ -37,16 +38,19 @@ export class Container extends React.Component<Properties, State> {
 
   static mapState(state: RootState): Partial<Properties> {
     const {
+      authentication,
```
dalefukami, line 41: I think we should start using the existing selectors for this:
`const currentUser = currentUserSelector(state);`

I know most of the other places reach all the way into the authentication object but that makes things very inflexible.

dalefukami (approved):

PR #1218, opened 2023-12-01
[Matrix] Add `Leave Group` functionality by a non admin member

ratik21:
### What does this do?

Adds functionality for a member to leave a group.

So that users that leave the group they're in, if he/she is not an admin.

https://github.com/zer0-os/zOS/assets/33264364/68f77457-5958-4623-b502-59c72e80ca79


src/components/group-management/leave-group-dialog/index.test.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+
+import { shallow } from 'enzyme';
+
+import { LeaveGroupDialog, Properties } from '.';
+
+describe('InviteDialog', () => {
```
dalefukami, line 7:
```suggestion
describe(LeaveGroupDialog, () => {
```

src/components/group-management/leave-group-dialog/index.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+      status: 0,
+      onClose: () => {},
+      onLeave: () => {},
+      ...props,
+    };
+
+    return shallow(<LeaveGroupDialog {...allProps} />);
+  };
+
+  it('renders the name of group', function () {
```
dalefukami, line 20: What if the group isn't named?

src/components/group-management/leave-group-dialog/index.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+    wrapper.find('IconButton').simulate('click');
+
+    expect(onClose).toHaveBeenCalled();
+  });
+
+  it('publishes onClose event when "cancel" button is clicked', () => {
+    const onClose = jest.fn();
+    const wrapper = subject({ onClose });
+
+    wrapper.find('Button[variant="text"]').simulate('press');
```
dalefukami, line 39:
```suggestion
    buttonLabelled('Cancel').simulate('press');
```

src/components/group-management/leave-group-dialog/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+          <IconButton {...cn('close')} size='large' Icon={IconXClose} onClick={this.props.onClose} />
+        </div>
+
+        <div {...cn('body')}>
+          Are you sure you want to leave {this.props.name}? You will lose access to the conversation and its history.
+        </div>
+
+        <div {...cn('footer')}>
+          <Button {...cn('text-button')} variant='text' onPress={this.props.onClose}>
+            <div {...cn('text-button-text')}>Cancel</div>
```
dalefukami, line 38: I'm not sure why we need an extra `div` here. Button's are supposed to "just work" if we use the right variant, aren't they?

src/store/channels-list/saga.ts
```diff
@@ -366,6 +366,22 @@ export function* userLeftChannel(channelId, userId) {
...
+// TODO: we can remove this function and simply use the "lastMessage.createdAt" property
+// look into https://github.com/zer0-os/zOS/pull/1063
+const conversationWithLatestMessage = (conversations) =>
+  conversations.reduce((latest, current) => {
+    if (current.messages && current.messages.length > 0) {
+      const latestMessage = current.messages.reduce((prevMsg, currentMsg) => {
+        return currentMsg.createdAt > prevMsg.createdAt ? currentMsg : prevMsg;
+      });
+      if (!latest || latestMessage.createdAt > latest.messages[0].createdAt) {
+        current.latestMessage = latestMessage;
```
dalefukami, line 378: Do we have to set this attribute at all? We're returning the conversation.

src/store/group-management/saga.ts
```diff
@@ -10,13 +12,16 @@ export function* saga() {
   yield fork(authWatcher);
 
   while (true) {
-    const { startEvent } = yield race({
+    const { startEvent, leaveGroupEvent } = yield race({
```
dalefukami, line 15: I don't think the leave group needs to be part of the panel management that the add member is part of since it doesn't require rendering anything in the side panel. Can we just have a `takeLatest` above this? If not, then maybe we should pull this panel management bit out to its own thing so we can fork it.

dalefukami (approved):
Nothing critical.

src/store/group-management/saga.ts
ratik21: makes sense, updated to use `takeLatest`

src/components/group-management-menu/index.tsx
```diff
@@ -34,15 +35,24 @@ export class GroupManagementMenu extends React.Component<Properties, State> {
...
     if (featureFlags.enableAddMemberToGroup && this.props.isRoomAdmin) {
       menuItems.push({
         id: 'add-member',
         label: this.renderMenuItem(<IconPlus />, 'Add Member'),
-        onSelect: this.startAddMember,
+        onSelect: this.startAddMember as any,
+      });
+    }
+
+    if (!this.props.isRoomAdmin) {
```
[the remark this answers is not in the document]
ratik21: understood.

src/components/group-management/leave-group-dialog/index.test.tsx
ratik21: oh I didn't consider that. For now, I am just displaying "Are you sure you want to leave this group?" (instead of the name). Will also ask Ethan and will update in followup if any changes required.

src/components/group-management/leave-group-dialog/index.tsx
ratik21: I added an extra `div` to update the CSS of the "text". Actually we're supposed to update the zUI buttons to add "size" variant (small, medium, large). But since we don't have that yet, i have added this div to apply the styling for "large text". 

I have added a TODO
image

src/store/channels-list/saga.ts
ratik21: updated but i still needed this for sorting. Created a `deepCopy` and did it on that.

src/components/group-management/leave-group-dialog/index.test.tsx
ratik21: Thanks for pointing! Didn't knew this existed 😄

src/components/group-management-menu/index.tsx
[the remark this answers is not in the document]
dalefukami: @ratik21 speaking of which... I just remembered, I'm pretty sure we decided that you're not allowed to leave a group if there are only 2 people left in the conversation. Is that stated in the Jira task? Do we need to add logic here?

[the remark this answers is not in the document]
ratik21: NO it wasn't. I will add. Also, on a separate note, we should "hide" the three dot menu completely for one on one conversations, no?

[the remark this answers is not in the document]
dalefukami: I think we should hide it if there are no available options. There's no reason you can't add people to a one on one conversation, for example.

PR #1222, opened 2023-12-04
feat: submit selected members and add to room

domw30:
### What does this do?
- submits the selected members and adds member/s to active conversation.

- to ensure state and client update with newly added members to room/conversation.

### How do I test this?
- Open messenger > window.FEATURE_FLAGS.enableGroupManagementMenu = true; > window.FEATURE_FLAGS.enableAddMemberToGroup = true; > Click three dot menu > Select Add Member > Add Member panel should be displayed in the sidekick (messenger list) > Add new member > Check conversation / room contains new member/s.

#### Things to consider:
  1. Does this change any APIs?

Add Members

https://github.com/zer0-os/zOS/assets/39112648/135c243f-6e97-4186-bb48-0c243817645c

### DIFF
--- a/src/lib/chat/index.ts
+++ b/src/lib/chat/index.ts
@@ -205,7 +205,7 @@ export class Chat {
   }
 
   async addMembersToRoom(roomId: string, users: User[]): Promise<void> {
-    return this.client.addMembersToRoom(roomId, users);
+    return this.matrix.addMembersToRoom(roomId, users);
   }
 
   async discardOlmSession(roomId: string) {
--- a/src/lib/chat/index.ts
+++ b/src/lib/chat/index.ts
@@ -83,6 +83,7 @@ export interface IChatClient {
   generateSecureBackup: () => Promise<any>;
   saveSecureBackup: (MatrixKeyBackupInfo) => Promise<void>;
   restoreSecureBackup: (recoveryKey: string) => Promise<void>;
+  addMembersToRoom: (roomId: string, users: User[]) => Promise<void>;
 }
 
 export class Chat {
@@ -201,6 +202,10 @@ export class Chat {
 
   async cancelAndResendKeyRequests() {
     return this.matrix.cancelAndResendKeyRequests();
+  }
+
+  async addMembers(roomId: string, users: User[]): Promise<void> {
+    return this.client.addMembersToRoom(roomId, users);
   }
 
   async discardOlmSession(roomId: string) {
--- a/src/lib/chat/index.ts
+++ b/src/lib/chat/index.ts
@@ -204,7 +204,7 @@ export class Chat {
     return this.matrix.cancelAndResendKeyRequests();
   }
 
-  async addMembers(roomId: string, users: User[]): Promise<void> {
+  async addMembersToRoom(roomId: string, users: User[]): Promise<void> {
     return this.client.addMembersToRoom(roomId, users);
   }
 
--- a/src/store/group-management/index.ts
+++ b/src/store/group-management/index.ts
@@ -23,10 +23,12 @@ export enum Stage {
 
 type GroupManagementState = {
   stage: Stage;
+  isAddingMembers: boolean;
 };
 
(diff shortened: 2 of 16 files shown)


src/store/group-management/saga.ts
```diff
@@ -1,22 +1,47 @@
...
+}
+
 export function* saga() {
   yield fork(authWatcher);
 
   while (true) {
-    const { startEvent } = yield race({
+    const { startEvent, addSelectedMembersEvent } = yield race({
       startEvent: take(SagaActionTypes.StartAddMember),
+      addSelectedMembersEvent: take(SagaActionTypes.AddSelectedMembers),
```
dalefukami, line 37: I'm confused why we're taking this event here AND on line 86. We should only ever get this event after we've started the process, right? So, we shouldn't need to listen for it here, I don't think.

src/store/group-management/index.ts
```diff
@@ -16,10 +23,16 @@ export enum Stage {
 
 type GroupManagementState = {
   stage: Stage;
+  roomMembers: {
```
dalefukami, line 26: I'm not sure what this is for. We set it when we add the users to the conversation but we never reference it?

src/lib/chat/index.ts
```diff
@@ -203,6 +204,10 @@ export class Chat {
     return this.matrix.cancelAndResendKeyRequests();
   }
 
+  async addMembersToRoom(roomId: string, users: User[]): Promise<void> {
+    return this.client.addMembersToRoom(roomId, users);
```
dalefukami, line 208: Simpler to just use `this.matrix.addMembersToRoom...` and just not add it to the common interface. We need to remove the interface anyway.

src/store/group-management/saga.ts
```diff
@@ -1,22 +1,47 @@
...
+
+    const userIds = selectedMembers.map((user) => user.value);
+    const users = yield select((state) => denormalizeUsers(userIds, state));
+
+    const chatClient: Chat = yield call(chat.get);
+    yield call([chatClient, chatClient.addMembersToRoom], roomId, users);
+    yield put(setRoomMembers(selectedMembers));
+
+    return Stage.None;
+  } catch (error) {}
```
dalefukami, line 28: So, if there's an error, the user just stays on the screen they're on? Should there be feedback or are we just going to assume the user will continue to click the button?

Side note: I am also a little worried that we don't disable the button while processing this. It opens us up to all the various "double click" bugs.

src/store/group-management/saga.ts
domw30: Yes correct, don't need to listen for it here. Thanks, updated.

src/lib/chat/index.ts
domw30: Updated

src/store/group-management/index.ts
domw30: Good spot, there is no reason for this state, I think I added this during the earlier stages of developing it and never removed. Thanks!

src/store/group-management/saga.ts
domw30: I've asked Eth how we want to handle the scenario where this would error. For now I've added a console log, with a comment to handle once we have UI for it/decided how we want to handle it.

re side note: I've added loading to this now which will disable the button while processing the adding of members: `isAddingMembers`

src/lib/chat/index.ts
dalefukami: You forgot to remove the method from the `IChatClient` interface and the SendbirdClient. That would be the reason to not use `this.client` here...so we don't have to maintain useless empty functionality.

src/store/group-management/saga.ts
dalefukami: Nice

dalefukami (approved):
Minor comments. Can be handled in a follow up.

dalefukami:
Huh, I must have done something weird... my comments showed up in a weird spot.

PR #1227, opened 2023-12-05
Use `canLeaveRoom` prop, don't show the "Leave Group" options if members in a conversation are <=2

ratik21:


dalefukami (approved):
Nice!

PR #1228, opened 2023-12-05
use `admins: string[]` instead of `admin: string` for room admins

ratik21:


src/lib/chat/matrix-client.ts
```diff
@@ -113,6 +113,18 @@ export class MatrixClient implements IChatClient {
...
+  private getRoomAdmins(room: Room): string[] {
+    const powerLevels = this.getLatestEvent(room, EventType.RoomPowerLevels);
+    if (!powerLevels) {
+      // possible if you've just _created_ the conversation, in which case we don't
+      // have the power levels events yet
+      return [room.getCreator()];
+    }
+    const powerLevelsByUser = powerLevels.getContent()?.users || {};
+    const admins = Object.keys(powerLevelsByUser).filter((userId) => powerLevelsByUser[userId] === PowerLevels.Owner);
+    return admins;
```
dalefukami, line 125: lol, didn't really need to go this far but if it works then great.

src/store/channels/index.ts
```diff
@@ -53,7 +53,7 @@ export interface Channel {
   hasLoadedMessages: boolean;
   conversationStatus: ConversationStatus;
   messagesFetchStatus: MessagesFetchState;
-  admin: string;
+  admins: string[];
```
dalefukami, line 56: At a glance you'd never know if this was actual objects of some type or not. `adminIds` or, probably better, `adminMatrixIds` since it's not obvious whether it's a user id or a matrix id.

dalefukami (approved):

src/store/channels/index.ts
ratik21: ok

PR #1237, opened 2023-12-05
refactor: minor group management improvements 

domw30:
### What does this do?
- makes `activeConversationId` required.
- removes `addMembersToRoom` from `IChatClient` and sendbird client.
- adds `finally` to `roomMembersSelected` saga function.

- minor improvements

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/store/group-management/saga.ts
+++ b/src/store/group-management/saga.ts
@@ -115,11 +115,11 @@ export function* roomMembersSelected(action) {
     const chatClient: Chat = yield call(chat.get);
     yield call([chatClient, chatClient.addMembersToRoom], roomId, users);
 
-    yield put(setIsAddingMembers(false));
     return Stage.None;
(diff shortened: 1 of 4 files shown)


src/store/group-management/saga.ts
```diff
@@ -72,22 +72,24 @@ export function* leaveGroup(action) {
 export function* roomMembersSelected(action) {
   const { users: selectedMembers, roomId } = action.payload;
 
+  yield put(setIsAddingMembers(true));
+
   try {
     if (!roomId || !selectedMembers) {
       return;
     }
```
dalefukami, line 80:
```suggestion
    if (!roomId || !selectedMembers) {
      return;
    }
    
  yield put(setIsAddingMembers(true));

  try {
```

dalefukami (approved):

PR #1239, opened 2023-12-05
Edit Conversation: List the users

dalefukami:
### What does this do?

Lists the users currently in the conversation on the Edit Conversation panel

### DIFF
--- a/src/components/citizen-list-item/index.tsx
+++ b/src/components/citizen-list-item/index.tsx
@@ -1,0 +1,41 @@
+import * as React from 'react';
+
+import { User } from '../../store/channels';
+import { bemClassName } from '../../lib/bem';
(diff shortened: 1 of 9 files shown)


src/components/citizen-list-item/index.tsx
```diff
@@ -0,0 +1,41 @@
+import * as React from 'react';
```
ratik21, line 1: why do we call it "citizen", instead of "member" (which we use in other places)?

ratik21 (approved):

src/components/citizen-list-item/index.tsx
dalefukami: The global app name (from the user perspective) is Citizen. We just haven't ever picked it up in our code. But, in this particular case, I used this to match the component name used in Figma.

PR #1245, opened 2023-12-06
Edit Conversation (Name, Icon): Saga logic

ratik21:
### What does this do?

Adds the saga logic to actually make the API call to matrix to update the icon/url.

(normal case)

https://github.com/zer0-os/zOS/assets/33264364/0592151d-93b2-4c4f-88c0-0f4e95a64b66

(case when setting a name to "empty", and vice versa)

https://github.com/zer0-os/zOS/assets/33264364/b7335aa5-0a89-41e8-9ccb-2557d267c1eb

### DIFF
--- a/src/store/group-management/index.ts
+++ b/src/store/group-management/index.ts
@@ -1,9 +1,15 @@
 import { PayloadAction, createAction, createSlice } from '@reduxjs/toolkit';
-import { GroupManagementErrors } from './types';
+import { EditConversationState, GroupManagementErrors } from './types';
 
 export interface MembersSelectedPayload {
   roomId: string;
   users: any[];
+}
+
+export interface EditConversationPayload {
+  name: string;
+  image: File | null;
+  roomId: string;
 }
 
 export enum SagaActionTypes {
@@ -14,6 +20,7 @@ export enum SagaActionTypes {
   Cancel = 'group-management/cancel',
   AddSelectedMembers = 'group-management/add-selected-members',
   RemoveMember = 'group-management/remove-member',
+  EditConversationNameAndIcon = 'group-management/edit-conversation-name-and-icon',
 }
 
 export enum Stage {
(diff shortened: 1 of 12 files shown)


src/store/group-management/index.ts
```diff
@@ -63,11 +75,30 @@ const slice = createSlice({
     setAddMemberError: (state, action: PayloadAction<GroupManagementState['addMemberError']>) => {
       state.addMemberError = action.payload;
     },
+    setEditConversationErrors: (
```
dalefukami, line 78: This seems like it's doing more than `set` the errors. It's adding errors?

Is there actually a reason to maintain the old errors? If we're setting them when something fails that probably means the previous error is no longer valid?

src/store/group-management/saga.ts
```diff
@@ -93,3 +100,30 @@ export function* roomMembersSelected(action) {
...
+        imageUrl = uploadResult.url;
+      } catch (error) {
+        yield put(setEditConversationErrors({ image: 'Failed to upload image, please try again...' }));
+        return;
+      }
+    }
+
+    const chatClient: Chat = yield call(chat.get);
+    yield call([chatClient, chatClient.editRoomNameAndIcon], roomId, name, imageUrl);
+    yield put(setEditConversationState(EditConversationState.SUCCESS));
```
dalefukami, line 122: Should we also clear the errors on success?

dalefukami (approved):

src/store/group-management/index.ts
ratik21: I..was keeping the errors as separate `{ image: <error_msg>, general: <error_msg>}`. This is what we're adding/setting in the sagas. NO need to maintain these errors separately, but..i thought that would make more sense.. (to make errors look more verbose)? 

Why do i feel i actually did the opposite of that here 😅

src/store/group-management/saga.ts
ratik21: Yes we should. Will add. (doubt) But i guess..we'll never reach this stage If there is ANY error (image/general)?

dalefukami: Right but, if I tried to save and there was an error...and then I try to save again and it succeeds, won't the errors still be there? Maybe I'm mistaken.

PR #1248, opened 2023-12-07
feat: add member left admin message when member leaves room

domw30:
### What does this do?
- admin message for when a member leaves a room

- to highlight to other users that a member has left the room

### How do I test this?
- with two test users on different browser profiles, initiate the leave room flow and leave the room with one of your test users. once you have left the room, check on the other test users conversation that the admin message clearly states which user has left.

#### Things to consider:
  1. Does this change any APIs?

Screenshot 2023-12-07 at 14 20 31

DEMO

https://github.com/zer0-os/zOS/assets/39112648/cb20bd68-cbfb-468e-995b-bfa0eaf97af7

### DIFF
--- a/src/lib/chat/chat-message.test.ts
+++ b/src/lib/chat/chat-message.test.ts
@@ -234,13 +234,13 @@ describe(adminMessageText, () => {
   });
 
   describe(AdminMessageType.CONVERSATION_STARTED, () => {
-    it('returns default message if creator not found', () => {
+    it('returns default message if admin user id not found', () => {
       const state = getState('current-user', {});
       const adminText = adminMessageText(
         {
           message: 'some message',
           isAdmin: true,
-          admin: { type: AdminMessageType.CONVERSATION_STARTED, creatorId: 'unknown-user-id' },
+          admin: { type: AdminMessageType.CONVERSATION_STARTED, userId: 'unknown-user-id' },
         } as Message,
         state
       );
@@ -248,12 +248,12 @@ describe(adminMessageText, () => {
       expect(adminText).toEqual('some message');
     });
 
-    it('translates message if current user was not the creator', () => {
-      const state = getState('current-user', { 'creator-id': { id: 'creator-id', firstName: 'Courtney' } });
+    it('translates message if current user id is not the admin user id', () => {
+      const state = getState('current-user', { 'admin-user-id': { id: 'admin-user-id', firstName: 'Courtney' } });
       const message = {
         message: 'some message',
         isAdmin: true,
-        admin: { type: AdminMessageType.CONVERSATION_STARTED, creatorId: 'creator-id' },
+        admin: { type: AdminMessageType.CONVERSATION_STARTED, userId: 'admin-user-id' },
       } as Message;
 
       const adminText = adminMessageText(message, state);
@@ -261,12 +261,12 @@ describe(adminMessageText, () => {
       expect(adminText).toEqual('Courtney started the conversation');
     });
 
-    it('translates message if current user was the creator', () => {
-      const state = getState('creator-id', { 'creator-id': { id: 'creator-id', firstName: 'Julie' } });
+    it('translates message if current user id is the admin user id', () => {
+      const state = getState('admin-user-id', { 'admin-user-id': { id: 'admin-user-id', firstName: 'Julie' } });
       const message = {
         message: 'some message',
         isAdmin: true,
-        admin: { type: AdminMessageType.CONVERSATION_STARTED, creatorId: 'creator-id' },
+        admin: { type: AdminMessageType.CONVERSATION_STARTED, userId: 'admin-user-id' },
       } as Message;
 
       const adminText = adminMessageText(message, state);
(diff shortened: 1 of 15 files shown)


src/lib/chat/matrix-client.ts
```diff
@@ -297,9 +297,18 @@ export class MatrixClient implements IChatClient {
     switch (event.type) {
       case EventType.RoomMessage:
         return mapMatrixMessage(event, this.matrix);
+
       case CustomEventType.USER_JOINED_INVITER_ON_ZERO:
       case EventType.RoomCreate:
         return mapEventToAdminMessage(event);
+
+      case EventType.RoomMember:
+        if (event.content.membership === 'leave') {
```
domw30, line 306: In the next admin message PR I will change `leave` to use the `MembershipStateType.Leave`

src/lib/chat/matrix/chat-message.ts
```diff
@@ -65,23 +65,38 @@ export async function mapMatrixMessage(matrixMessage, sdkMatrixClient: SDKMatrix
...
+      return getRoomMemberAdminData(content, senderId);
+    case EventType.RoomCreate:
+      return { type: AdminMessageType.CONVERSATION_STARTED, creatorId: senderId };
+    default:
+      return {};
+  }
+}
+
+function getRoomMemberAdminData(content, senderId) {
+  if (content.membership === 'leave') {
```
domw30, line 94: Same here.

src/lib/chat/chat-message.test.ts
```diff
@@ -274,6 +274,35 @@ describe(adminMessageText, () => {
...
+          isAdmin: true,
+          admin: { type: AdminMessageType.MEMBER_LEFT_CONVERSATION, creatorId: 'unknown-user-id' },
+        } as Message,
+        state
+      );
+
+      expect(adminText).toEqual('some message');
+    });
+
+    it('translates message if current user was not the creator', () => {
```
domw30, line 293: I'll also replace the content of `it` here with `translates message if creator is found` as we do not check for current user for this admin message type.

src/store/channels-list/saga.ts
```diff
@@ -139,6 +158,13 @@ export function* fetchConversations() {
     chatClient,
     chatClient.getConversations,
   ]);
+
+  const matrixIds = extractMatrixIdsFromConversations(conversations);
+  if (matrixIds.length > 0) {
+    yield call(fetchMissingUsersData, matrixIds);
+  }
+
+  yield call(updateMessageSenders, conversations);
```
dalefukami, line 167: We already have a `mapToZeroUsers` method. Seems like it makes sense to do this there rather than doing this mapping in 2 separate steps.

But, thinking more high level, the issue here is that we usually base our fetches on the `otherMembers` and some members may not be in the conversation any more, right? Maybe we just need a new field on the conversation that holds everyone who was ever invited to a conversation? Then we wouldn't need to sift through every single message.

src/lib/chat/chat-message.test.ts
```diff
@@ -274,6 +274,35 @@ describe(adminMessageText, () => {
...
   });
+
+  describe(AdminMessageType.MEMBER_LEFT_CONVERSATION, () => {
+    it('returns default message if creator not found', () => {
+      const state = getState('current-user', {});
+      const adminText = adminMessageText(
+        {
+          message: 'some message',
+          isAdmin: true,
+          admin: { type: AdminMessageType.MEMBER_LEFT_CONVERSATION, creatorId: 'unknown-user-id' },
```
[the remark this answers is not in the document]
domw30: I'll need to update or add `userId` in Message. Is there a generic name you could think of in place of `creatorId`.

```
  admin?: {
    type: AdminMessageType;
    inviterId?: string;
    inviteeId?: string;
    creatorId?: string;
  };
```

src/lib/chat/matrix-client.ts
```diff
@@ -868,6 +881,7 @@ export class MatrixClient implements IChatClient {
 
   private mapConversation = async (room: Room): Promise<Partial<Channel>> => {
     const otherMembers = this.getOtherMembersFromRoom(room).map((userId) => this.mapUser(userId));
+    const memberHistory = this.getMemberHistoryFromRoom(room).map((userId) => this.mapMemberForHistory(userId));
```
dalefukami, line 884: As we discussed, let's modify the user presence fetching to be more asynchronous (not delay the conversation rendering) and then we can use the `mapUser` here.

Hacked possible pseudo code
```
  // const chatClient = yield call(chat.get);
  //   for (let user of users) {
  //     const matrixId = user?.matrixId;
  //     if (!matrixId) continue;
  //     const presenceData = yield call([chatClient, chatClient.getUserPresence], matrixId);
  //     if (!presenceData) continue;
  //     const { lastSeenAt, isOnline } = presenceData;
        //  yield put(receiveUser({userId: ..., lastSeenAt, isOnline}))
  //   }
```

src/store/channels/index.ts
```diff
@@ -40,6 +40,7 @@ export interface Channel {
   name: string;
   messages: Message[];
   otherMembers: User[];
+  memberHistory: Partial<User>[];
```
dalefukami, line 43: Maybe just a whole `User` now?

src/lib/chat/chat-message.test.ts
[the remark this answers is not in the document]
domw30: Updated to `userId`

src/store/channels/index.ts
domw30: Update to `User`

dalefukami (approved):

PR #1271, opened 2023-12-12
Remove `isSideKickOpen` global var

ratik21:
### What does this do?

removes code related to `isSideKickOpen`.

### DIFF
--- a/src/Main.tsx
+++ b/src/Main.tsx
@@ -19,7 +19,6 @@ import { featureFlags } from './lib/feature-flags';
 export interface Properties {
   hasContextPanel: boolean;
   isContextPanelOpen: boolean;
-  isSidekickOpen: boolean;
   isMessengerFullScreen: boolean;
   context: {
     isAuthenticated: boolean;
@@ -34,7 +33,6 @@ export class Container extends React.Component<Properties> {
       return {
         hasContextPanel: false,
         isContextPanelOpen: false,
-        isSidekickOpen: true,
         isMessengerFullScreen: true,
       };
     }
@@ -42,7 +40,6 @@ export class Container extends React.Component<Properties> {
     return {
       hasContextPanel: layout.hasContextPanel,
       isContextPanelOpen: layout.isContextPanelOpen,
-      isSidekickOpen: layout.isSidekickOpen,
       isMessengerFullScreen: false,
     };
   }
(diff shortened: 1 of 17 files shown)


src/Main.tsx
```diff
@@ -54,7 +51,7 @@ export class Container extends React.Component<Properties> {
   render() {
     const mainClassName = classNames('main', {
       'context-panel-open': this.props.isContextPanelOpen,
-      'sidekick-panel-open': this.props.isSidekickOpen && this.props.context.isAuthenticated,
+      'sidekick-panel-open': this.props.context.isAuthenticated,
```
dalefukami, line 54: As a note: `public ZOS` is also not a thing any more so we shouldn't even need to add this class. It should just be that way all the time. `has-context-panel` can also be removed because its state should also be constant now.

Doesn't mean you have to do it. Just noting more potential cleanup.

src/components/user-actions/index.tsx
```diff
@@ -62,7 +61,7 @@ export class UserActions extends React.Component<Properties, State> {
       <>
         <div className='user-actions'>
           <button className='user-actions__icon-button'>
-            <IconMessageSquare2 isFilled={this.props.isConversationListOpen} />
+            <IconMessageSquare2 isFilled />
```
dalefukami, line 64: Better solution here could have been to just remove this component as it's never rendered any more.

src/lib/chat/matrix-client.ts
```diff
@@ -86,28 +86,29 @@ export class MatrixClient implements IChatClient {
...
 
-  async getUserPresence(userId: string) {
+  async getUserPresence(_userId: string) {
     await this.waitForConnection();
 
-    try {
-      const userPresenceData = await this.matrix.getPresence(userId);
+    return { lastSeenAt: null, isOnline: false };
+    // try {
+    //   const userPresenceData = await this.matrix.getPresence(userId);
```
dalefukami, line 94: Was it intentional to comment out all this code?

dalefukami (commented):
Suggestion for future changes like this would be to do it in steps. Find one component/area that's impacted by that flag and remove the conditions on it. Once that's been deployed for a while with no impact then move on to another spot until nothing is referencing the state any more...then you can remove the state.

src/Main.tsx
ratik21: got it

PR #1272, opened 2023-12-13
Remove all code related to `isMessengerFullScreen`, `isFullScreen`

ratik21:
### What does this do?

It removes all code related to "checking" whether the messenger is full screen or not. Since it is always fullscreen now. 

NOTE: As i started removing code, ..it got a little complex, because there's a lot of code here which was dependent directly/indirectly on that `var` (`isFullScreen`). I have noted some followups which can be done in later PR's.


ratik21:
closing this, will do it in smaller PR's

PR #1274, opened 2023-12-13
refactor: remove rewards from messenger list

domw30:
### What does this do?
- removes rewards from messenger list

Before
Screenshot 2023-12-13 at 11 23 10

After
Screenshot 2023-12-13 at 11 09 08

### DIFF
--- a/src/components/messenger/list/index.tsx
+++ b/src/components/messenger/list/index.tsx
@@ -28,11 +28,8 @@ import { getMessagePreview, previewDisplayDate } from '../../../lib/chat/chat-me
 import { enterFullScreenMessenger } from '../../../store/layout';
 import { Modal, ToastNotification } from '@zero-tech/zui/components';
 import { InviteDialogContainer } from '../../invite-dialog/container';
-import { fetch as fetchRewards, rewardsPopupClosed, rewardsTooltipClosed } from '../../../store/rewards';
-import { RewardsContainer } from '../../rewards-container';
 import { receiveSearchResults } from '../../../store/users';
 import { SettingsMenu } from '../../settings-menu';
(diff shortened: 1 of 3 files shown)


src/components/messenger/list/index.tsx
```diff
@@ -227,22 +205,6 @@ export class Container extends React.Component<Properties, State> {
...
-              meowPreviousDay={this.props.meowPreviousDay}
-              isRewardsLoading={this.props.isRewardsLoading}
-              isMessengerFullScreen={this.props.isMessengerFullScreen}
-              showRewardsInTooltip={this.props.showRewardsInTooltip}
-              showRewardsInPopup={this.props.showRewardsInPopup}
-              isFirstTimeLogin={this.props.isFirstTimeLogin}
-              onRewardsPopupClose={this.props.rewardsPopupClosed}
-              onRewardsTooltipClose={this.props.rewardsTooltipClosed}
-              hasLoadedConversation={this.props?.conversations[0]?.hasLoadedMessages}
-            />
```
ratik21, line 243: so we're not gonna implement rewards in the future? Or will it 'not' be a part of messenger list?

domw30: Yeah, I thought best to remove the RewardsContainer from here as the Rewards will be a subcomponent of a new panel (profile panel). 

https://www.figma.com/file/btwltL4Os8hAy3eS3pxaa6/ZERO-Messenger-Web-%7C-Glass?type=design&node-id=12361-20661&mode=design&t=8qNSkEFnhIoF6nTR-4

Similar thinking to how we've added `<GroupManagement />`. Something like `<UserProfile />` would contain the rewards container....Have a look at <GroupManagement />` in messenger list, let me know if that makes sense

PR #1275, opened 2023-12-13
feat: add users name to user account container

domw30:
### What does this do?
- adds user name to user account container

Before
Screenshot 2023-12-13 at 11 59 48

After
Screenshot 2023-12-13 at 11 59 18

### DIFF
--- a/src/components/messenger/list/styles.scss
+++ b/src/components/messenger/list/styles.scss
@@ -17,17 +17,6 @@ $side-padding: 16px;
 }
 
 .messenger-list {
-  &__header {
-    height: 32px;
-    background-color: theme.$color-primary-3;
-    border-radius: 8px 8px 0px 0px;
-    display: flex;
-    justify-content: flex-end;
-    gap: 2px;
-    align-items: center;
-    padding-right: 13px;
-  }
-
   &__user-account-container {
     display: flex;
     align-items: center;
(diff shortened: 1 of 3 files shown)


src/components/messenger/list/styles.scss
```diff
@@ -20,14 +20,17 @@ $side-padding: 16px;
...
-    display: flex;
-    padding-left: 16px;
-    width: 100%;
-    height: 64px;
-    background-color: theme.$color-primary-3;
+  &__user-name {
+    font-size: $font-size-small;
+    font-weight: 600;
+    line-height: 15px;
+    color: rgba(253, 253, 255, 0.93);
```
domw30, line 33: Hmm I've directly added the rgba for the glass colour here, did we discuss how we are adding these colours, such as a colour library?

src/components/messenger/list/styles.scss
```diff
@@ -20,13 +21,17 @@ $side-padding: 16px;
...
-  &__settings-menu-container {
-    display: flex;
-    padding-left: 16px;
-    width: 100%;
-    height: 64px;
+  &__user-name {
+    @include glass-text-primary-color;
+
+    font-size: $font-size-small;
+    font-weight: 600;
```
dalefukami, line 33: Did we find out if there are "categories" of text rather than just one color?

dalefukami (approved):

PR #1277, opened 2023-12-13
refactor: extract user account container to user header component

domw30:
### What does this do?
-  extracts user account container and adds a user header component

- this aligns with the naming of components/containers on figma (`UserHeader`).
- keeps the user header and children components separate from messenger list.
- extracts tests not required in messenger list

NOTE: As with previous PR, settings menu will be removed at some point. When this happens we just need to extract the avatar component from the settings menu dropdown and add into the `UserHeader` component.

### DIFF
--- a/src/components/messenger/list/user-header/index.tsx
+++ b/src/components/messenger/list/user-header/index.tsx
@@ -1,0 +1,41 @@
+import * as React from 'react';
+
+import { SettingsMenu } from '../../../settings-menu';
+
+import { bemClassName } from '../../../../lib/bem';
+import './styles.scss';
+
+const cn = bemClassName('user-header');
+
+export interface Properties {
+  userName: string;
+  userHandle: string;
+  userAvatarUrl: string;
+  userIsOnline: boolean;
+
+  includeUserSettings?: boolean;
+  onLogout?: () => void;
+}
+
+export class UserHeader extends React.Component<Properties> {
+  get userStatus(): 'active' | 'offline' {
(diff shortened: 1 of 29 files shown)


src/components/messenger/list/user-header/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export class UserHeader extends React.Component<Properties> {
+  get userStatus(): 'active' | 'offline' {
+    return this.props.userIsOnline ? 'active' : 'offline';
+  }
+
+  render() {
+    return (
+      <div {...cn('')}>
+        {this.props.includeUserSettings && (
```
ratik21, line 28: while we're doing this, maybe remove this flag now? It is coming from `isMessengerFullScreen` global state, which is always true now

domw30: Yes good shout! I'll remove that in a follow up PR

src/components/messenger/list/user-header/styles.scss
```diff
@@ -0,0 +1,18 @@
...
+  padding: 16px;
+  height: 32px;
+  gap: 8px;
+  background-color: theme.$color-primary-3;
+
+  &__user-name {
+    font-size: $font-size-small;
+    font-weight: 600;
+    line-height: 15px;
+    color: rgba(253, 253, 255, 0.93);
```
dalefukami, line 16: Mixin here?

dalefukami (approved):

PR #1280, opened 2023-12-13
refactor: replace three dot icon with icon button

domw30:
### What does this do?
- replaces group management three dot icon with the `IconButton` component that contains the three dot icon. This component 

- the styles of the icon button align with figma, i.e. correct padding and size.
- the icon button will have a hover.

Screenshot 2023-12-13 at 14 53 21
Screenshot 2023-12-13 at 14 50 13
Screenshot 2023-12-13 at 14 50 24

### DIFF
--- a/src/components/group-management-menu/index.tsx
+++ b/src/components/group-management-menu/index.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 
 import { IconDotsHorizontal, IconEdit5, IconPlus, IconUserRight1 } from '@zero-tech/zui/icons';
-import { DropdownMenu } from '@zero-tech/zui/components';
+import { DropdownMenu, IconButton } from '@zero-tech/zui/components';
 
 import './styles.scss';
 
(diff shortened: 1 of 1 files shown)


src/components/group-management-menu/index.tsx
```diff
@@ -81,7 +81,7 @@ export class GroupManagementMenu extends React.Component<Properties, State> {
         side='bottom'
         alignMenu='end'
         onOpenChange={this.handleOpenChange}
-        trigger={<IconDotsHorizontal size={24} isFilled />}
+        trigger={<IconButton Icon={IconDotsHorizontal} size={32} isFilled onClick={() => {}} />}
```
domw30, line 84: `onClick` doesn't need to do anything here as `trigger` handles the click functionality.

domw30: Alternatively we could add a wrapper around `<IconDotsHorizontal size={24} isFilled />` and set the hover styles with another className, but adding `IconButton` instead won't require us to add any other styles.

dalefukami (approved):

PR #1282, opened 2023-12-13
Remove background color/fill from conversationList header

ratik21:
### What does this do?

Before:
image

After:
image


ratik21:
> Are there any other changes to the header that you are making? One of my PRs removes the '+' icon from the panel and adds to the header so you won't need to make that change. Wondering as my changes today are all header related.

No i am just adding the glass related style on the conversation list now

PR #1283, opened 2023-12-13
refactor: updateUserPresence to accept users directly and dispatch actions

domw30:
### What does this do?
- updates `updateUserPresence` to accept users directly and dispatch actions (users presence in state).

- streamline/enhance fetching users presence.

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/store/channels-list/saga.ts
+++ b/src/store/channels-list/saga.ts
@@ -5,6 +5,7 @@ import uniqBy from 'lodash.uniqby';
 import { takeLatest, put, call, take, race, all, select, spawn } from 'redux-saga/effects';
 import { SagaActionTypes, receive, denormalizeConversations } from '.';
 import { chat } from '../../lib/chat';
+import { receive as receiveUser } from '../users';
 
 import { AsyncListStatus } from '../normalized';
 import {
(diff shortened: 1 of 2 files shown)


src/store/channels-list/saga.ts
```diff
@@ -110,8 +104,10 @@ export function* fetchConversations() {
     chatClient,
     chatClient.getConversations,
   ]);
+  const otherMembersOfConversations = conversations.map((c) => c.otherMembers);
```
dalefukami, line 107: Are we sure this works? This would create an array of arrays.... doesn't seem like the user presence function handles that.

src/store/channels-list/saga.ts
```diff
@@ -110,8 +104,10 @@ export function* fetchConversations() {
...
+
   yield call(mapToZeroUsers, conversations);
-  yield call(updateUserPresence, conversations);
+  yield call(updateUserPresence, otherMembersOfConversations);
```
dalefukami, line 110: Use `fork` so this doesn't delay the rest of the conversation loading.

src/store/channels-list/saga.ts
```diff
@@ -408,7 +404,7 @@ export function* addChannel(channel) {
   const conversationsList = yield select(rawConversationsList());
   const channelsList = yield select(rawChannelsList());
   yield call(mapToZeroUsers, [channel]);
-  yield call(updateUserPresence, [channel]);
+  yield call(updateUserPresence, [channel.otherMembers]);
```
dalefukami, line 407: `fork`

src/store/channels-list/saga.ts
dalefukami: I'm also worried that at this point the member objects haven't been mapped to Zero users yet? So, perhaps this should happen after that call.

src/store/channels-list/saga.ts
domw30: updated, thanks!

src/store/channels-list/saga.ts
domw30: updated, thanks!

src/store/channels-list/saga.ts
domw30: oops yes I was supposed to flatten! Updated the the order as well. Thanks!

dalefukami (approved):

PR #1286, opened 2023-12-14
ZOS Redesign (glass style): Conversations Panel

ratik21:
### What does this do?

Applies "glass" styles (stroke, fill, box-shadow, and blur), to the conversation list panel.

Before:
image

After:
image

### DIFF
--- a/src/_glass.scss
+++ b/src/_glass.scss
@@ -1,6 +1,17 @@
 // "glass" style mixin
 @mixin glass-outer {
-  border: 1px solid rgba(255, 255, 255, 0.25); // stroke should be applied to the outer div
+  // border-width: 1px;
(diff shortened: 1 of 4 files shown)


src/_glass.scss
```diff
@@ -0,0 +1,13 @@
+// "glass" style mixin
+@mixin glass-outer {
+  border: 1px solid rgba(255, 255, 255, 0.25); // stroke should be applied to the outer div
```
dalefukami, line 3: This doesn't look like the right gradient. I've been playing with it and it doesn't seem like it's possible to do what the design asks for. I'm contacting Eth to see what he'd like to do.

dalefukami (approved):
Removed the border for now.

PR #1287, opened 2023-12-14
ZOS Redesign: Apply flat/thick style to conversation header

ratik21:
Before:

image

After:
image


src/_glass.scss
```diff
@@ -0,0 +1,5 @@
+// "flat" style mixin
+@mixin flat-thick {
+  background: var(--d-glass-materials-flat-thick, rgba(10, 10, 10, 0.75));
```
dalefukami, line 3: Not really worth using a variable that we know we don't even have. Just use the color directly.

dalefukami (approved):

PR #1310, opened 2023-12-17
Add default and active states for message input (composer)

ratik21:
### What does this do?

Previously there was only a single state for message input background. The new design adds 2 - one for default, and another for active (i.e focused)

Default
image

Active (when typing something)
image

### DIFF
--- a/src/_glass.scss
+++ b/src/_glass.scss
@@ -69,11 +69,13 @@
   backdrop-filter: blur(64px);
 }
 
-@mixin text-input {
-  background: var(--D-Glass-Ui-State-Active, rgba(163, 162, 163, 0.1));
-
-  /* D/Glass/Inset/Shadows */
+@mixin text-input-default {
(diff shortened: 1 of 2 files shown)


domw30 (commented):
Looks good ! Minor comment on naming, `text-input-default` and `text-input-focus` should be more generic as they are not specific to the `text` input, from what I've seen at a quick glance at figma. 

This is how I've set the names in zUI 
https://github.com/zer0-os/zUI/blob/f7b991ff390253691ee338107bac7611d015bccf/src/styles/_glass.scss#L3

ratik21:
Hey @domw30 thanks for review. Weird, i my figma i see names like `--D-Glass-Ui-State-Active` & `--D-Glass-Ui-State-default`, instead of `glass-inset-input..`. 

But we should use the names in zOS as in zUI. will update.

PR #1315, opened 2023-12-18
ZOS Redesign: Update conversation thread divider color text to primary

ratik21:
### What does this do?

image

NOTE: Not sure if this should also be applies to the colors of admin messages as well. Will ask Ethan about it.


domw30 (approved):
Nice 👍 

Yeah, looks like the admin message should be updated as well.

https://www.figma.com/file/btwltL4Os8hAy3eS3pxaa6/ZERO-Messenger-Web-%7C-Glass?type=design&node-id=12384-202059&mode=design&t=kYgNjnjKs9WL0pTD-4

PR #1318, opened 2023-12-18
feat: unread count redesign

domw30:
### What does this do?
- implements the redesign for the unread message count

- as per figma designs

Screenshot 2023-12-18 at 18 33 03
Screenshot 2023-12-18 at 18 32 59

### DIFF
--- a/src/_glass.scss
+++ b/src/_glass.scss
@@ -13,9 +13,19 @@
   color: rgba(238, 236, 255, 0.42);
 }
 
+// D/Glass/Ui State/Default
+@mixin glass-state-default-color {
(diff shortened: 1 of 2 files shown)


src/_glass.scss
```diff
@@ -85,14 +95,15 @@
 }
 
 // D/Glass/Materials/Raised
-@mixin glass-materials-raised {
+@mixin glass-materials-raised($startColor: rgba(167, 163, 163, 0.05), $endColor: rgba(167, 163, 163, 0.05)) {
```
dalefukami, line 98: I'm a little confused as to why we had to add parameters here. Nothing is using these.

PR #1329, opened 2023-12-20
(Bug)  `+` icon not responding after you open a conversation with someone already on your list

ratik21:
### What does this do?

The issue was with the race condition during create conversation's `startEvent` saga. It sets up a few race conditions, but if we "don't create a conversation", & just open an already existing one, the race condition isn't resolved and hence the `+` icon does not respond afterwards.

Before:

https://github.com/zer0-os/zOS/assets/33264364/5823c8e7-1233-40bd-aeb7-b4c59e039bd5

After:

https://github.com/zer0-os/zOS/assets/33264364/07580419-aa02-4d75-958d-e3913457d0fc

### DIFF
--- a/src/store/channels/saga.ts
+++ b/src/store/channels/saga.ts
@@ -1,7 +1,7 @@
 import getDeepProperty from 'lodash.get';
 import { takeLatest, put, call, select, spawn } from 'redux-saga/effects';
 import { SagaActionTypes, receive, schema, removeAll } from '.';
-
+import { SagaActionTypes as CreateConversationSagaActionTypes } from '../create-conversation';
 import { joinChannel as joinChannelAPI } from './api';
 import { takeEveryFromBus } from '../../lib/saga';
 import { Events as ChatEvents, getChatBus } from '../chat/bus';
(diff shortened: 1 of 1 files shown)


src/store/channels/saga.ts
```diff
@@ -81,6 +81,8 @@ export function* openConversation(conversationId) {
   }
 
   yield call(reset);
+  // cancel start of create conversation flow if it's in progress (resolves the race condition)
+  yield put({ type: CreateConversationSagaActionTypes.Cancel });
```
ratik21, line 85: This cancels the "startConversation", which resolves the race condition.

image

Not sure if this is the correct place, since this will also be called whenever user just clicks on a conversation from the conversation list panel. But the otherway would be to call it from the frontend i guess (from `createOneOnOneConversation` in `messenger/list/index` file..)

dalefukami: Yeah, I think the problem is that we ended up putting business logic in the component (https://github.com/zer0-os/zOS/blob/main/src/components/messenger/list/index.tsx#L143)

That should probably just call up to the sagas where we can handle bailing out of the flow properly rather than trying to determine that a conversation exists within the component.

PR #1341, opened 2024-01-04
fix: resolve group management menu nested button console error

domw30:
### What does this do?
- removes the use of `IconButton`.
- adds `IconDotsHorizontal` in place of the button.
- makes use of the `className` attribute on the dropdown. This is the className for the trigger.
- added trigger styles to stylesheet.

- prevent the console error.

#### Things to consider:
  1. Does this change any APIs?

Screenshot 2024-01-04 at 09 31 22

### DIFF
--- a/src/components/group-management-menu/index.tsx
+++ b/src/components/group-management-menu/index.tsx
@@ -82,11 +82,7 @@ export class GroupManagementMenu extends React.Component<Properties, State> {
         alignMenu='end'
         onOpenChange={this.handleOpenChange}
         className={'group-management-trigger'}
-        trigger={
-          <div>
-            <IconDotsHorizontal size={24} isFilled />
-          </div>
-        }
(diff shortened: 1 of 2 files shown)


src/components/group-management-menu/index.tsx
```diff
@@ -81,7 +81,8 @@ export class GroupManagementMenu extends React.Component<Properties, State> {
         side='bottom'
         alignMenu='end'
         onOpenChange={this.handleOpenChange}
-        trigger={<IconButton Icon={IconDotsHorizontal} size={32} isFilled onClick={() => {}} />}
+        className={'group-management-trigger'}
+        trigger={<IconDotsHorizontal size={24} isFilled />}
```
dalefukami, line 85: Hmm... is this going to cause us problems if we ever change the icon button styles?

dalefukami: Though, I can't really think of a fix. lol

PR #1351, opened 2024-01-05
feat: clear invite when closing invite dialog

domw30:
### What does this do?
- dispatch a `clearInvite` saga action to clear the invite when the invite dialog is closed.

Issue:
The InviteDialog component's button flickers between disabled and enabled states on subsequent openings. Initially, on first opening, the inviteCode is not present, so the button is correctly disabled. However, on closing and reopening the dialog, the button quickly toggles from enabled (as inviteCode is initially present) to disabled (as inviteCode is momentarily cleared due to a reset() in the fetch saga) and back to enabled (after inviteCode is refetched and set).

Root Cause:
The fetchInvite saga always resets the inviteCode state on each call, causing the inviteCode to be temporarily undefined even if it was previously fetched. This reset, followed by a refetch, leads to flickering behaviour of the button tied to the inviteCode's availability.

### How do I test this?
Open invite dialog on first load > check disabled state is as expected > close dialog and reopen > check disabled state of button again.

#### Things to consider:
  1. Does this change any APIs?

Before:

https://github.com/zer0-os/zOS/assets/39112648/118728b5-fe2d-420f-a024-c7bf5295ce61

After:

https://github.com/zer0-os/zOS/assets/39112648/c643d043-be0c-499f-9dea-6d1ff417711c

### DIFF
--- a/src/store/create-invitation/saga.ts
+++ b/src/store/create-invitation/saga.ts
@@ -35,7 +35,12 @@ function* clearOnLogout() {
   yield put(reset());
 }
 
+export function* clearInvite() {
+  yield put(reset());
+}
+
 export function* saga() {
   yield takeLeading(SagaActionTypes.GetCode, fetchInvite);
   yield takeEveryFromBus(yield call(getAuthChannel), AuthEvents.UserLogout, clearOnLogout);
+  yield takeLeading(SagaActionTypes.Cancel, clearInvite);
 }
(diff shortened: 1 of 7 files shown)


src/store/create-invitation/saga.ts
```diff
@@ -35,7 +35,12 @@ function* clearOnLogout() {
...
 }
 
+export function* clearInvite() {
+  yield put(reset());
+}
+
 export function* saga() {
   yield takeLeading(SagaActionTypes.GetCode, fetchInvite);
   yield takeEveryFromBus(yield call(getAuthChannel), AuthEvents.UserLogout, clearOnLogout);
+  yield takeLeading(SagaActionTypes.Cancel, clearInvite);
```
domw30, line 45: The saga test file already has test coverage for `Cancel` action.

dalefukami (approved):

PR #1356, opened 2024-01-05
Remove the Channels App

dalefukami:
### What does this do?

Removes the Channels App component and some related, unreachable, business logic


src/components/messenger/chat/index.tsx
```diff
@@ -27,7 +27,7 @@ import { MessageInput } from '../../message-input/container';
 import { searchMentionableUsersForChannel } from '../../../platform-apps/channels/util/api';
 import { Media } from '../../message-input/utils';
 
-export interface PublicProperties { }
```
dalefukami, line 30: Not quite sure why prettier started complaining about these.

PR #1364, opened 2024-01-08
feat: add error dialog component

domw30:
### What does this do?
- Adds an `ErrorDialog` component

- To be used as a 'pop-up' to notify the user that they are not a member of the conversation. 

This is a very basic iteration of an `ErrorDialog`. The component can be made more generic/customisable in the future if needed.

#### Things to consider:
  1. Does this change any APIs?

Screenshot 2024-01-08 at 15 03 19


src/components/error-dialog/container.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react';
```
dalefukami, line 1: Why do we need a container if it's not providing any functionality?

src/components/error-dialog/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export class ErrorDialog extends React.Component<Properties> {
+  render() {
+    return (
+      <div {...cn('')}>
+        <IconButton {...cn('close')} size={40} Icon={IconXClose} onClick={this.props.onClose} />
+
+        <div {...cn('heading-container')}>
+          <div {...cn('header')}>Access Denied</div>
+          <div {...cn('divider')} />
```
dalefukami, line 24: Do we need a "divider"? Why can't the `header` or the `heading-container` just have the border?

dalefukami (approved):

src/components/error-dialog/container.tsx
domw30: I was thinking the same, but wondered if at any point in the future we would be pulling in state or actions if this dialog was to be used in other scenarios. Maybe thinking a little too far ahead for that. I can do a follow up to this PR and remove the container

src/components/error-dialog/index.tsx
domw30: Yeah I can update one of the header borders instead - admittedly this was just a copy and paste from one of the other modals.

dalefukami: We definitely need to generalize this modal a bit. We shouldn't need to rebuild this whole thing every time. lol

PR #1366, opened 2024-01-08
feat: implement validate active conversation saga function

domw30:
### What does this do?
- adds functionality to be able to validate the active conversation

- to be able to handle the display of the error dialog when validating a conversation id (good/bad url)

#### Things to consider:
  1. Does this change any APIs?


domw30:
Or...Do we want to handle the state of the Modal (open/closed) in the redux state/sagas?

src/store/channels-list/saga.ts
```diff
@@ -91,6 +92,9 @@ export function* fetchConversations() {
 
   const channel = yield call(getConversationsBus);
   yield put(channel, { type: ConversationEvents.ConversationsLoaded });
+
+  // Validate the active conversation after conversations are loaded
+  yield call(performValidateActiveConversation);
```
domw30, line 97: The follow up PR for displaying the error dialog will handle performing this validation check when the url/room id changes.

src/store/chat/saga.ts
```diff
@@ -82,6 +85,25 @@ export function* setActiveConversation(id: string) {
   history.push({ pathname: `/conversation/${id}` });
 }
 
+export function* performValidateActiveConversation() {
+  yield put(setIsErrorDialogOpen(false));
```
dalefukami, line 89: Probably don't need this. It should already be closed if we're just now performing the validation.

src/store/chat/saga.ts
```diff
@@ -82,6 +85,25 @@ export function* setActiveConversation(id: string) {
...
+    return;
+  }
+
+  if (activeConversationId && !isMemberOfActiveConversation) {
+    yield put(setIsErrorDialogOpen(true));
+  }
+}
+
+export function* closeErrorDialog() {
+  yield call(openFirstConversation);
```
dalefukami, line 104: We should also `setIsErrorDialogOpen` to false, right? That's what closes the dialog.

src/store/chat/saga.ts
```diff
@@ -70,10 +73,45 @@ export function* setActiveConversation(id: string) {
...
+
+  if (!activeConversationId || conversationList.length === 0) {
+    return;
+  }
+
+  const isMemberOfActiveConversation = conversationList.includes(activeConversationId);
+  yield put(setIsConversationErrorDialogOpen(!isMemberOfActiveConversation));
+}
+
+function* watchForConversationChange() {
```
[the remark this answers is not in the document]
ratik21: I guess the case when the user isn't a part of the conversation, and he still tries to access it would be when he has an unknown url, and tries to load it, in which case we can just move this check to `setActiveConversationId`? (i.e validate the conversation before setting the active conversationId ?)

dalefukami (approved):

PR #1368, opened 2024-01-10
feat: display error dialog if active conversation is not valid

domw30:
### What does this do?
- wires up displaying the error dialog if the active conversation fails the validation check.

- to display an error to the user if the room url (active conversation id) is invalid.

### How do I test this?
- visit a url that is either `bad` or does not contain a room id that the user is not a member of > check the error dialog is displayed > when closing the dialog check the first conversation in the list opens.

#### Things to consider:
  1. Does this change any APIs?

https://github.com/zer0-os/zOS/assets/39112648/6f3e7870-9b11-4c9e-b1f1-6f3443299a74


src/messenger-main.tsx
```diff
@@ -34,6 +36,7 @@ export class Container extends React.Component<Properties> {
   componentDidUpdate(prevProps: Properties): void {
     if (this.idChanged(prevProps)) {
       this.props.setActiveConversationId(this.conversationId);
+      this.props.validateActiveConversation();
```
dalefukami, line 39: This should happen in a saga. There should be no business logic in the components.

ratik21 (approved):

PR #1390, opened 2024-01-11
feat: add sort members and is user admin util functions to messenger/list/utils

domw30:
### What does this do?
- adds helper functions (utils) to sort members for view group info panel and edit group info panel.

In the `sortMembers` function, the order is:

1. The current user is sorted to the top.
2. Admins are sorted next.
3. Users are then sorted based on their online status, with online users appearing before offline users.
4. Lastly, users are sorted alphabetically by their first name.

### DIFF
--- a/src/components/messenger/list/utils/utils.tsx
+++ b/src/components/messenger/list/utils/utils.tsx
@@ -20,18 +20,28 @@ export function isCustomIcon(url?: string) {
   return !url.startsWith('https://static.sendbird.com/sample');
 }
 
-export function isUserAdmin(user: User, adminIds: string[]) {
-  return adminIds.includes(user.matrixId);
+export function isUserAdmin(user: Partial<User>, adminIds: string[]) {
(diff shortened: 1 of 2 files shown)


src/components/messenger/list/utils/utils.tsx
```diff
@@ -18,3 +19,29 @@ export function isCustomIcon(url?: string) {
...
   return !url.startsWith('https://static.sendbird.com/sample');
 }
+
+export function isUserAdmin(user: User, adminIds: string[]) {
+  return adminIds.includes(user.matrixId);
+}
+
+export function sortMembers(members: User[], adminIds?: string[], currentUserId?: string) {
+  return members.sort((a, b) => {
+    const aIsAdmin = adminIds ? isUserAdmin(a, adminIds) : false;
```
dalefukami, line 29: This is over complicated. Just make the adminIds required.

PR #1396, opened 2024-01-12
feat: navigate to edit group flow from view group information panel

domw30:
### What does this do?
- adds edit button to view group information panel to navigate to the edit group panel.

- as per design - so the user can access the edit group panel from the view group information panel.

### How do I test this?
- if you are admin of a group, navigate to the group information panel > click `Edit` button > check edit group panel is displayed.

#### Things to consider:
  1. Does this change any APIs?

Navigate to Edit Group Panel from View Group Information Panel.
Note :: we may choose to improve/adjust back navigation at a later date.

https://github.com/zer0-os/zOS/assets/39112648/8337f1c9-c9a9-4d97-bb1e-c056a2aa300e

### DIFF
--- a/src/components/messenger/list/view-group-information-panel/index.test.tsx
+++ b/src/components/messenger/list/view-group-information-panel/index.test.tsx
@@ -4,7 +4,7 @@ import { ViewGroupInformationPanel, Properties } from '.';
 import { CitizenListItem } from '../../../citizen-list-item';
 import { User } from '../../../../store/channels';
 import { PanelHeader } from '../panel-header';
-import { Image } from '@zero-tech/zui/components';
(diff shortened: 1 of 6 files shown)


src/components/messenger/list/view-group-information-panel/index.test.tsx
```diff
@@ -51,6 +62,11 @@ describe(ViewGroupInformationPanel, () => {
     expect(wrapper).toHaveElement(IconUsers1);
   });
 
+  it('renders edit button when current user is admin', () => {
```
dalefukami, line 65: Probably should have the inverse test too to ensure people who can't edit don't see it.

PR #1399, opened 2024-01-12
feat: open leave group dialog from view group information panel

domw30:
### What does this do?
- opens leave group dialog from view group information panel if current user is not admin.

- as per designs.
- 
### How do I test this?
- navigate to a group conversation you are not an admin of > click group management three dot dropdown menu > select group info menu item > click Leave Group buttons.

Note: the leave group button is not exact to figma designs as we will be updating the button component in the near future.

https://github.com/zer0-os/zOS/assets/39112648/59e40da2-9e11-482d-a486-f2b75ce1a9aa

### DIFF
--- a/src/components/messenger/list/view-group-information-panel/index.tsx
+++ b/src/components/messenger/list/view-group-information-panel/index.tsx
@@ -1,12 +1,13 @@
 import * as React from 'react';
 
 import { Button, IconButton, Image } from '@zero-tech/zui/components';
-import { IconArrowNarrowLeft, IconPlus, IconUsers1 } from '@zero-tech/zui/icons';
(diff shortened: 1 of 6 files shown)


src/components/messenger/list/view-group-information-panel/index.tsx
```diff
@@ -113,14 +119,31 @@ export class ViewGroupInformationPanel extends React.Component<Properties> {
...
+    const showLeaveGroupButton = !isCurrentUserRoomAdmin && otherMembers.length > 1;
+
     return (
       <div {...cn()}>
         {this.renderBanner()}
         {this.renderBackIcon()}
         <div {...cn('body')}>
           {this.renderDetails()}
           {this.renderMembers()}
+          {showLeaveGroupButton && this.renderLeaveGroupButton()}
```
dalefukami, line 146: Use a `canLeaveRoom` property like the group-management-menu does.

PR #1403, opened 2024-01-15
refactor: move group permissions logic into group management mapState

domw30:
### What does this do?
- uses `canAddMembers`, `canEditGroup`, `canLeaveGroup` props in place of `isCurrentUserRoomAdmin`.
- moves permissions logic into group management mapState.

- improve code quality.

#### Things to consider:
  1. Does this change any APIs?


domw30:
We're using a range of terminology for `rooms` within Group Management, i.e. `EditConversationPanel`, `canLeaveGroup`, `isCurrentUserRoomAdmin`, as well as classNames and test descriptions. As previously discussed, we made an error with the naming of `GroupManagement`, this really should have been `Room Management`. We should discuss when a good time would be to put some time into refactoring the naming involved with group management. For now I have used `group` just for consistency with classNames and other prop names.

src/components/messenger/list/view-group-information-panel/index.test.tsx
```diff
@@ -88,14 +90,19 @@ describe(ViewGroupInformationPanel, () => {
...
     expect(wrapper).toHaveElement(Button);
   });
 
-  it('renders leave group button when current user is not admin and more than 1 other member', () => {
+  it('does not render edit button when user cannot edit group', () => {
+    const wrapper = subject({ canEditGroup: false });
+    expect(wrapper).not.toHaveElement(Button);
+  });
+
+  it('renders leave group button when current user can leave group and more than 1 other member', () => {
```
ratik21, line 103:
```suggestion
  it('renders leave group button when current user can leave group and the group has more than 1 other member', () => {
```

ratik21 (approved):

dalefukami (approved):

PR #1406, opened 2024-01-16
Validate active conversation id (room alias, create/join a room via API)

ratik21:
### What does this do?

Adds validation logic while setting/loading a new conversation

- a room can be referenced by an alias, upon which we get the room_id (by alias). If it doesn't exist, then we call the backend API to create one.
- if the id is not an alias, but the user isn't a member of the conversation by that id, then we call the API to join the room.

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -618,12 +618,8 @@ export class MatrixClient implements IChatClient {
     return await this.matrix
       .getRoomIdForAlias(alias) // { room_id, servers[] }
       .catch((err) => {
-        if (err.errcode === 'M_NOT_FOUND') {
+        if (err.errcode === 'M_NOT_FOUND' || err.errcode === 'M_INVALID_PARAM') {
           Promise.resolve(undefined);
-        }
-
-        if (err.errcode === 'M_INVALID_PARAM') {
-          // handle invalid alias error
         }
       })
       .then((response) => response && response.room_id);


src/store/chat/saga.ts
```diff
@@ -91,8 +96,28 @@ export function* performValidateActiveConversation() {
...
+    if (!featureFlags.allowJoinRoom) {
+      return;
+    }
+
+    conversationId = yield call(
+      [chatClient, chatClient.joinRoom],
+      conversationId ? conversationId : activeConversationId
+    );
+  }
+  yield put(setActiveConversationId(conversationId));
```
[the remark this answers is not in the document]
dalefukami: Ah, right, I forgot that your changes will cause the url and the activeId to diverge.

src/lib/chat/index.ts
```diff
@@ -80,6 +80,8 @@ export interface IChatClient {
   generateSecureBackup: () => Promise<any>;
   saveSecureBackup: (MatrixKeyBackupInfo) => Promise<void>;
   restoreSecureBackup: (recoveryKey: string) => Promise<void>;
+  getRoomIdForAlias: (alias: string) => Promise<string | undefined>;
+  apiJoinRoom: (aliasOrId: string) => Promise<{ success: boolean; response: any; message: string }>;
```
dalefukami, line 84: We really don't need to keep adding to this interface since we only have a single client now. Ok to leave in but for the future.

src/lib/chat/matrix-client.ts
```diff
@@ -613,6 +613,44 @@ export class MatrixClient implements IChatClient {
...
+    await this.waitForConnection();
+    return await this.matrix
+      .getRoomIdForAlias(alias) // { room_id, servers[] }
+      .catch((err) => {
+        if (err.errcode === 'M_NOT_FOUND') {
+          Promise.resolve(undefined);
+        }
+
+        if (err.errcode === 'M_INVALID_PARAM') {
+          // handle invalid alias error
```
dalefukami, line 626: for us this should just be the same as if the room does not exist. There's no reason to tell the user anything else, I don't think.

src/lib/chat/matrix-client.ts
```diff
@@ -613,6 +613,44 @@ export class MatrixClient implements IChatClient {
     await this.matrix.redactEvent(roomId, messageId);
   }
 
+  async getRoomIdForAlias(alias: string): Promise<string | undefined> {
+    await this.waitForConnection();
+    return await this.matrix
+      .getRoomIdForAlias(alias) // { room_id, servers[] }
+      .catch((err) => {
```
dalefukami, line 620: Is there a particular reason you didn't use try/catch here instead of using the promise functions?

src/lib/chat/matrix-client.ts
```diff
@@ -613,6 +613,44 @@ export class MatrixClient implements IChatClient {
...
+        }
+
+        if (err.errcode === 'M_INVALID_PARAM') {
+          // handle invalid alias error
+        }
+      })
+      .then((response) => response && response.room_id);
+  }
+
+  async apiJoinRoom(aliasOrId: string): Promise<{ success: boolean; response: any; message: string }> {
```
dalefukami, line 632: I don't think this should be on the matrix-client. Ok to leave for now but it can probably just live as an independent function in the chat lib.

src/messenger-main.tsx
```diff
@@ -5,13 +5,14 @@ import { connectContainer } from './store/redux-container';
...
 import { ZUIProvider } from '@zero-tech/zui/ZUIProvider';
 import { Provider as AuthenticationContextProvider } from './components/authentication/context';
-import { setActiveConversationId } from './store/chat';
+import { setActiveConversationId, validateAndSetActiveConversationId } from './store/chat';
 
 export interface Properties {
   isAuthenticated: boolean;
 
   match: { params: { conversationId: string } };
   setActiveConversationId: (id: string) => void;
```
dalefukami, line 14: We can remove this now?

src/store/chat/saga.ts
```diff
@@ -56,6 +57,7 @@ function* activateWhenConversationsLoaded(activate) {
 
   if (conversationsLoaded) {
     activate();
+    console.log('activate');
```
dalefukami, line 60:
```suggestion
```

src/store/chat/index.ts
```diff
@@ -12,9 +12,13 @@ const initialState: ChatState = {
 
 export enum SagaActionTypes {
   CloseConversationErrorDialog = 'chat/saga/closeConversationErrorDialog',
+  ValidateAndSetActiveConversationId = 'channelsList/saga/ValidateAndSetActiveConversationId',
 }
 
 const closeConversationErrorDialog = createAction(SagaActionTypes.CloseConversationErrorDialog);
+export const validateAndSetActiveConversationId = createAction<{ id: string }>(
```
dalefukami, line 19: This feels like an awkward name. The caller shouldn't really know what's involved in specifying the active conversation. But I'm struggling to come up with something that's not already too overloaded. `setActiveConversationId` is already how we do the actual data set.... `activateConversation`, perhaps?.... :thinking:

dalefukami: Maybe the redux setter could change to `rawSetActiveConversationId`? Does that make that one clearly different and then we can use `setActiveConversationId` here?... just throwing out ideas.

dalefukami (approved):

src/lib/chat/index.ts
ratik21: got it.

src/lib/chat/matrix-client.ts
ratik21: weirdly, `try catch` wasn't catching the error here. So i had to use `.catch()` & `.then()` :/

src/lib/chat/matrix-client.ts
ratik21: i'll note it as a follow up

src/messenger-main.tsx
ratik21: it's being used with `componentDidUpdate`

src/store/chat/index.ts
ratik21: `rawSetActiveConversationId` & `setActiveConversationId` makes more sense

src/lib/chat/matrix-client.ts
ratik21: 👍

src/messenger-main.tsx
dalefukami: Pretty sure we'd want the same validation when the component updates. Otherwise if we use a linkable channel in the chat somehow (think "#zos-core") then someone could click a conversation they aren't a part of.

PR #1409, opened 2024-01-17
feat: add util to translate join room api error codes to error dialog content

domw30:
### What does this do?
- adds function to translate join room api request error codes to error dialog content.

- to determine what error dialog content to display when, for whatever reason, a user does not have access to a room / non existing room.

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/store/chat/utils.ts
+++ b/src/store/chat/utils.ts
@@ -1,0 +1,24 @@
+export enum JoinRoomApiErrorCode {
+  ROOM_NOT_FOUND = 'ROOM_NOT_FOUND',
+  ACCESS_TOKEN_REQUIRED = 'ACCESS_TOKEN_REQUIRED',
+  GENERAL_ERROR = 'GENERAL_ERROR',
+}
+
+const ERROR_DIALOG_CONTENT = {
+  [JoinRoomApiErrorCode.ROOM_NOT_FOUND]: {
+    header: 'Conversation Not Found',
+    body: 'This conversation does not exist.',
+  },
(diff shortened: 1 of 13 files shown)


src/store/chat/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+  ROOM_NOT_FOUND = 'ROOM_NOT_FOUND',
+  ACCESS_TOKEN_REQUIRED = 'ACCESS_TOKEN_REQUIRED',
+  GENERAL_ERROR = 'GENERAL_ERROR',
+}
+
+const ERROR_DIALOG_CONTENT = {
+  [JoinRoomApiErrorCode.ROOM_NOT_FOUND]: {
+    header: 'Conversation Not Found',
+    body: 'This conversation does not exist.',
+  },
```
ratik21, line 11: Isn't the API supposed to create a room, if it does not exist? How can we get this error code from the API? 🤔

src/store/chat/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+    header: 'World Members Only',
+    body: 'You cannot join this conversation as you do not own a domain in this World.',
+  },
+  [JoinRoomApiErrorCode.GENERAL_ERROR]: {
+    header: 'Error',
+    body: 'An error occurred while verifying the conversation.',
+  },
+};
+
+export function translateJoinRoomApiError(errorCode: JoinRoomApiErrorCode) {
```
ratik21, line 22:
```suggestion
export function parseJoinRoomApiError(errorCode: JoinRoomApiErrorCode) {
```

dalefukami (approved):

PR #1410, opened 2024-01-17
feat: add set join room error content state

domw30:
### What does this do?
- adds `setJoinRoomErrorContent` state to chat saga.
- sets the join room error content when joinRoomApi request errors.
- sets the join room error content when user is not a member of the active conversation.

- to display the correct error message in the error dialog when the user is denied access to a room url for whatever reason.

#### Things to consider:
  1. Does this change any APIs?


src/components/error-dialog/index.tsx
```diff
@@ -9,6 +9,11 @@ import './styles.scss';
 
 const cn = bemClassName('error-dialog');
 
+export type ErrorDialogContent = {
```
domw30, line 12: These will be used in the error dialog in a follow up PR. Keeping this PR separate from those changes.

src/store/chat/saga.ts
```diff
@@ -115,6 +122,14 @@ export function* performValidateActiveConversation(activeConversationId: string)
...
     const isUserMemberOfActiveConversation = yield call(isMemberOfActiveConversation, activeConversationId);
     yield put(setIsConversationErrorDialogOpen(!isUserMemberOfActiveConversation));
+    if (!isUserMemberOfActiveConversation) {
+      yield put(
+        setJoinRoomErrorContent({
+          header: 'Access Denied',
+          body: 'You do not have permission to join this conversation.',
+        })
+      );
+    }
```
ratik21, line 132: it could also be the case that the room does not exist?

src/store/chat/saga.test.ts
```diff
@@ -140,11 +178,47 @@ describe(closeErrorDialog, () => {
...
 });
 
+describe(joinRoom, () => {
+  describe('error scenarios', () => {
+    const roomIdOrAlias = 'some-room-id-or-alias';
+    it.each([
+      [JoinRoomApiErrorCode.ROOM_NOT_FOUND, ERROR_DIALOG_CONTENT[JoinRoomApiErrorCode.ROOM_NOT_FOUND]],
+      [JoinRoomApiErrorCode.ACCESS_TOKEN_REQUIRED, ERROR_DIALOG_CONTENT[JoinRoomApiErrorCode.ACCESS_TOKEN_REQUIRED]],
+      [JoinRoomApiErrorCode.GENERAL_ERROR, ERROR_DIALOG_CONTENT[JoinRoomApiErrorCode.GENERAL_ERROR]],
+    ])('handles error "%s"', async (errorCode, expectedErrorContent) => {
```
ratik21, line 205: This seems a little complex, could you please explain this?

domw30: `it.each` is used to simplify writing multiple, similar tests. I've now changed this and added an individual test for each error case instead so its a little easier to follow. 

` const expectedErrorContent = ERROR_DIALOG_CONTENT[JoinRoomApiErrorCode.ROOM_NOT_FOUND];` 

is:

```
 [JoinRoomApiErrorCode.ROOM_NOT_FOUND]: {
    header: 'Conversation Not Found',
    body: 'This conversation does not exist.',
  }
```

Hopefully the changes will help make more sense!

src/store/chat/saga.ts
ratik21: > isMemberOfActiveConversation is false, this doesn't necessarily mean that the room does not exist.

Correct, the room can exist, but if the user isn't part of it, then we call the API where the user can request to join that room. But that's featureFlagged, and i see this error is thrown if the feature flag is not enabled.

I guess we can just update this error message to a more generic one - maybe `either the room doesn't exist, or you don't have permission to join it`? But fine for now, since this will be removed anyway.

src/store/chat/saga.test.ts
ratik21: thanks

src/store/chat/saga.ts
dalefukami: Generally speaking if you don't have permission then we should tell the user it doesn't exist. Much like if you try to view a github private repo but you're not logged in. This prevents people from just spamming urls until they find ids that exist.

dalefukami: Also, given that the feature flag is only there while we wait for the api end to be completed, it's pretty meaningless what happens when it's disabled.

src/components/messenger/list/index.tsx
```diff
@@ -162,7 +162,7 @@ export class Container extends React.Component<Properties, State> {
   }
 
   get isErrorDialogOpen(): boolean {
-    return this.props.isConversationErrorDialogOpen;
+    return this.props.joinRoomErrorContent !== null;
```
dalefukami, line 165:
```suggestion
    return !!this.props.joinRoomErrorContent;
```
Might be better because it's possible it accidentally gets set to undefined?

dalefukami (approved):

PR #1411, opened 2024-01-17
feat: render appropriate error when user is unable to join room

domw30:
### What does this do?
- wires up the rendering of the appropriate error when `joinRoomApi` response provides an error code.
- wires up the rendering of the appropriate error when user is not a member of the active conversation.
- updates the `ErrorDialog` to handle the different errors.

- render appropriate error when user is unable to join room.

#### Things to consider:
  1. Does this change any APIs?


domw30:
A follow up to this PR will be to add an additional button to the Error Dialog, i.e `Buy A Domain`

src/components/error-dialog/index.test.tsx
```diff
@@ -4,9 +4,14 @@ import { shallow } from 'enzyme';
...
 import { Button, IconButton } from '@zero-tech/zui/components';
 
+import { bem } from '../../lib/bem';
+
+const c = bem('.error-dialog');
+
 describe('ErrorDialog', () => {
   const subject = (props: Partial<Properties>) => {
     const allProps: Properties = {
+      content: { header: '', body: '' },
```
domw30, line 14: lift props up

src/components/messenger/list/index.tsx
```diff
@@ -165,6 +166,10 @@ export class Container extends React.Component<Properties, State> {
     return this.props.joinRoomErrorContent !== null;
   }
 
+  get errorDialogContent(): ErrorDialogContent {
```
dalefukami, line 169: Is this used?

dalefukami (approved):

PR #1416, opened 2024-01-18
Remove `openFirstConversationAfterChannelsLoaded` from `store/login/saga.ts`

ratik21:
### What does this do?

Since we're validating the active conversation each time the components mounts/updates (after channels load), we can remove a previous function which was opening the first conversation after loading the channels. We can just open the first conversation in `performValidateActiveConversation` only.

Earlier this was causing an issue since two events were fired at the same time: `openFirstConversation` (by `openFirstConversationAfterChannelsLoaded` saga event), and the `validateActiveConversation` saga event.


dalefukami (approved):
Nice find

PR #1417, opened 2024-01-19
fix: handle conversation header avatar

domw30:
### What does this do?
- correctly handles the conversation header avatar for group conversations, one on one conversations and whether there is a custom icon or not.

- to fix how the avatar is displayed in the conversation header depending on the conditions mentioned above.

### How do I test this?
- open messenger and check the avatar of the different types of conversations and set icons.

#### Things to consider:
  1. Does this change any APIs?

Before: 
Screenshot 2024-01-19 at 12 29 03
Screenshot 2024-01-19 at 12 29 19
Screenshot 2024-01-19 at 12 29 21

After:
Screenshot 2024-01-19 at 12 26 47
Screenshot 2024-01-19 at 12 26 44
Screenshot 2024-01-19 at 12 26 41

### DIFF
--- a/src/components/messenger/chat/index.test.tsx
+++ b/src/components/messenger/chat/index.test.tsx
@@ -179,7 +179,6 @@ describe('messenger-chat', () => {
       const headerAvatar = wrapper.find('.direct-message-chat__header-avatar');
 
       expect(headerAvatar).toHaveProp('style', { backgroundImage: 'url(avatar-url)' });
-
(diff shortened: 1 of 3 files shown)


domw30:
I've made a note to refactor and improve the `messenger/chat` test file here whenever next would be best to do so. Lots of improvements can be made to the tests. In addition, we should really be using the Avatar component for the conversation header. For now, I've made the change without needing to spend too much time on this.

PR #1421, opened 2024-01-22
feat: handle navigation link data for join room error dialog content

domw30:
### What does this do?
- Adds `linkPath` and `linkText` to the relevant error dialog content (`ACCESS_TOKEN_REQUIRED`).
- When error code is `ACCESS_TOKEN_REQUIRED`, we need to add a `linkPath` which should include the room alias. 

- This will ensure that the user will be able to navigate to the correct url in the zns explorer.

### How do I test this?
- You can modify the current code that is run whilst we have a feature flag on `joinRoom` to be set with the `ACCESS_TOKEN_REQUIRED` error code and pass in a room alias, for example, in the demo below I have used the `free` room alias.

#### Things to consider:
  1. Does this change any APIs?

E.G. If room alias was `free`

https://github.com/zer0-os/zOS/assets/39112648/f187265a-d7ba-4130-ba33-61dfddcaff4b

### DIFF
--- a/src/store/chat/saga.ts
+++ b/src/store/chat/saga.ts
@@ -119,8 +119,8 @@ export function* performValidateActiveConversation(activeConversationId: string)
     } else {
       yield put(
         setJoinRoomErrorContent({
-          header: 'Access Denied',
-          body: 'You do not have permission to join this conversation.',
+          header: 'There’s no one here...',
+          body: 'This conversation does not exist or you are not a member.',
         })
       );
     }
--- a/src/store/chat/utils.ts
+++ b/src/store/chat/utils.ts
@@ -7,12 +7,14 @@ export enum JoinRoomApiErrorCode {
 
 export const ERROR_DIALOG_CONTENT = {
   [JoinRoomApiErrorCode.ROOM_NOT_FOUND]: {
-    header: 'Conversation Not Found',
-    body: 'This conversation does not exist.',
+    header: 'There’s no one here...',
+    body: 'This conversation does not exist or you are not a member.',
   },
   [JoinRoomApiErrorCode.ACCESS_TOKEN_REQUIRED]: {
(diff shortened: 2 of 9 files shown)


src/store/chat/saga.ts
```diff
@@ -119,8 +119,8 @@ export function* performValidateActiveConversation(activeConversationId: string)
     } else {
       yield put(
         setJoinRoomErrorContent({
-          header: 'Access Denied',
-          body: 'You do not have permission to join this conversation.',
+          header: 'There’s no one here...',
```
domw30, line 122: Updated these as per figma design ['here'](https://www.figma.com/file/btwltL4Os8hAy3eS3pxaa6/ZERO-Messenger-Web-%7C-Glass?type=design&node-id=12825-41059&mode=design&t=dlAwjivfKOymGjtB-4)

src/store/chat/utils.ts
```diff
@@ -7,12 +7,14 @@ export enum JoinRoomApiErrorCode {
...
   [JoinRoomApiErrorCode.ROOM_NOT_FOUND]: {
-    header: 'Conversation Not Found',
-    body: 'This conversation does not exist.',
+    header: 'There’s no one here...',
+    body: 'This conversation does not exist or you are not a member.',
   },
   [JoinRoomApiErrorCode.ACCESS_TOKEN_REQUIRED]: {
     header: 'World Members Only',
-    body: 'You cannot join this conversation as you do not own a domain in this World.',
+    body: 'You cannot join this conversation as your wallet does not hold a domain in this world. Buy a domain or switch to a wallet that holds one.',
```
domw30, line 15: As mentioned above, updated these as per figma designs

dalefukami (approved):
Yeah, probably good enough for now.

PR #1423, opened 2024-01-22
Wait for `UserJoinedRoom` Event before setting the `rawActiveConversationId`

ratik21:
### What does this do?

After calling the API (`/matrix/room/join`), we should wait for the "real time" event from Matrix before setting the conversation Id. 

Previously we were setting the conversationId before even receiving the event, which was causing the UI to not behave properly (eg. messages fetch failed, UI crashing for some users etc).

### DIFF
--- a/src/store/chat/saga.ts
+++ b/src/store/chat/saga.ts
@@ -2,7 +2,7 @@ import { put, select, call, take, takeEvery, spawn, race, takeLatest } from 'red
 import { takeEveryFromBus } from '../../lib/saga';
 
 import { rawSetActiveConversationId, SagaActionTypes, setJoinRoomErrorContent, clearJoinRoomErrorContent } from '.';
-import { createChatConnection, getChatBus } from './bus';
+import { Events as ChatBusEvents, createChatConnection, getChatBus } from './bus';
 import { getAuthChannel, Events as AuthEvents } from '../authentication/channels';
 import { getSSOToken } from '../authentication/api';
 import { currentUserSelector } from '../authentication/saga';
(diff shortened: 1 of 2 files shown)


src/store/chat/saga.ts
```diff
@@ -105,6 +104,19 @@ export function* isMemberOfActiveConversation(activeConversationId) {
   return conversationList.includes(activeConversationId);
 }
 
+// NOTE: we're waiting for the event to be fired, but if it doesn't..then
+// we should just keep showing the loading spinner?
+export function* setWhenUserJoinedRoom(conversationId: string) {
+  const { userJoined } = yield race({
```
dalefukami, line 110: What happens if we've already joined the room?

dalefukami (approved):
One scenario noted that we should handle.

src/store/chat/saga.ts
ratik21: > What happens if we've already joined the room?

This function won't be called. We call it after calling the `joinRoom` (API), which is only called when user isn't in a room, or the room does not exist.

ratik21: > Also, what if it's a different room event?

not sure what you mean here? For both creation & join we get this event (since the user is invited to the room either way)

dalefukami: But what if the event came in _before_ we got the result from the server?

dalefukami: > > Also, what if it's a different room event?
> 
> not sure what you mean here? For both creation & join we get this event (since the user is invited to the room either way)

Sorry, I meant, what if we get a join event for a different room? Then this would still cause problems, right?

ratik21: ohh..right. we need to ensure that we get the event for "this" room, and if we get another `joinRoom` event, then we should still keep waiting. 🤔

ratik21: updated ^

src/store/chat/saga.ts
```diff
@@ -93,10 +93,9 @@ export function* joinRoom(roomIdOrAlias: string) {
   if (!success) {
     const error = translateJoinRoomApiError(response);
     yield put(setJoinRoomErrorContent(error));
-    return undefined;
   } else {
     yield put(clearJoinRoomErrorContent());
-    return response.roomId;
+    yield call(setWhenUserJoinedRoom, response.roomId); // wait for user to join room
```
dalefukami, line 98:
```suggestion
    yield call(setWhenUserJoinedRoom, response.roomId);
```

dalefukami (approved):

PR #1424, opened 2024-01-23
feat: return primaryZID when fetching user data

domw30:
### What does this do?
- returns primaryZID when fetching user data.

- to use/display the users primaryZID in the UI, i.e. display users primaryZID with user details/search by primaryZID.

### How do I test this?
- run api locally, set primaryZID in database table and check data is returned when fetching user data by logging the primaryZID.

#### Things to consider:
  1. Does this change any APIs?

Screenshot 2024-01-23 at 10 33 16

### DIFF
--- a/src/store/channels-list/utils.ts
+++ b/src/store/channels-list/utils.ts
@@ -114,6 +114,6 @@ export function rawUserToDomainUser(u): User {
     lastName: u.profileSummary?.lastName,
     profileImage: u.profileSummary?.profileImage,
     lastSeenAt: u.lastActiveAt,
-    primaryZID: u?.primaryZID,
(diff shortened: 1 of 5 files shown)


ratik21 (approved):

src/store/channels-list/utils.ts
```diff
@@ -114,5 +114,6 @@ export function rawUserToDomainUser(u): User {
     lastName: u.profileSummary?.lastName,
     profileImage: u.profileSummary?.profileImage,
     lastSeenAt: u.lastActiveAt,
+    primaryZID: u?.primaryZID,
```
dalefukami, line 117:
```suggestion
    primaryZID: u.primaryZID,
```

PR #1426, opened 2024-01-23
feat: add get user handle util

domw30:
### What does this do?
- adds util to handle getting the `userHandle` i.e. zid, wallet address, nothing.

- to set the `userHandle` in state and pass to components.

### How do I test this?
- best way to test this would be run the util tests added for the `getUserHandle` util function.

### DIFF
--- a/src/components/messenger/list/utils/utils.test.ts
+++ b/src/components/messenger/list/utils/utils.test.ts
@@ -1,4 +1,4 @@
-import { isUserAdmin, sortMembers } from './utils';
+import { getUserHandle, isUserAdmin, sortMembers } from './utils';
 import { User } from '../../../../store/channels';
 
 describe('sortMembers', () => {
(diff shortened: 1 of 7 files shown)


src/components/messenger/list/utils/utils.tsx
```diff
@@ -41,3 +42,17 @@ export function sortMembers(members: User[], adminIds: string[]) {
     return a.firstName!.localeCompare(b.firstName);
   });
 }
+
+export function getUserHandle(primaryZID: string, wallets: Wallet[]) {
+  if (primaryZID) {
+    return primaryZID;
+  }
```
ratik21, line 49: if `primaryZID` is an optional parameter, when we should pass it after the `wallets: Wallet[]` in the function call

src/components/messenger/list/utils/utils.test.ts
```diff
@@ -44,3 +44,20 @@ describe('isUserAdmin', () => {
...
+  });
+
+  it('returns truncated publicAddress from the first wallet when primaryZID is absent', () => {
+    const user = { primaryZID: null, wallets: [{ id: 'wallet-id-1', publicAddress: '123456789' }] };
+    expect(getUserHandle(user.primaryZID, user.wallets)).toEqual('1234...6789');
+  });
+
+  it('returns empty string when both primaryZID and wallets are absent', () => {
+    const user = { primaryZID: null, wallets: [] };
+    expect(getUserHandle(user.primaryZID, user.wallets)).toEqual('');
```
ratik21, line 61: a `nit` but i think `getUserHandle` should just take in the `currentUser` & return the `handle` 🤔

ratik21 (approved):

src/components/messenger/list/utils/utils.test.ts
domw30: We should probably only pass in the necessary data rather than the whole `currentUser`

src/components/messenger/list/utils/utils.tsx
domw30: `primaryZID` isn't an optional parameter for `getUserHandle`. The `if` statement handles this.

ratik21: i am saying you should do `getUserHandle(wallets: [], undefined)` rather than `getUserHandle(undefined, wallets: [])`

PR #1427, opened 2024-01-23
feat: set userHandle in messenger list state using getUserHandle util

domw30:
### What does this do?
- set `userHandle` in messenger list state using `getUserHandle` util

- to set `userHandle` to either `primaryZID`, `publicAddress` or `''` in messenger list state to pass to components i.e. `user-header`.

### How do I test this?
- for now you will only be able to render a users wallet address. login with a wallet and console log the `userHeader`.

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/messenger/list/index.tsx
+++ b/src/components/messenger/list/index.tsx
@@ -34,6 +34,7 @@ import { receiveSearchResults } from '../../../store/users';
 import { Stage as GroupManagementSagaStage } from '../../../store/group-management';
 import { GroupManagementContainer } from './group-management/container';
 import { UserHeader } from './user-header';
+import { getUserHandle } from './utils/utils';
 
 import { bemClassName } from '../../../lib/bem';
 import './styles.scss';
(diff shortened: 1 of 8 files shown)


src/components/messenger/list/index.tsx
```diff
@@ -95,7 +96,7 @@ export class Container extends React.Component<Properties, State> {
       isFirstTimeLogin: registration.isFirstTimeLogin,
       isInviteNotificationOpen: registration.isInviteToastOpen,
       userName: user?.data?.profileSummary?.firstName || '',
-      userHandle: (hasWallet ? user?.data?.wallets[0]?.publicAddress : user?.data?.profileSummary?.primaryEmail) || '',
```
ratik21, line 98: shouldn't this be featureFlagged (via `allowEmailRegistration`)? We're assuming everyone logs in via wallet but i think it is still a feature flagged feature

domw30: `userHandle` no longer uses `primaryEmail` here so we shouldn't need to feature flag this. If the user has a `primaryZID`, we'll render that, if not, we'll render the wallet address, if that's not available for whatever reason then we won't render the `userHandle`

ratik21: We still have a featureFlag `allowEmailRegistration`. What happens when i enable that & register via email & I don't have a ZID? That's what i was asking.

PR #1434, opened 2024-01-24
feat: add verify id nav link to user header

domw30:
### What does this do?
- adds the 'verify id' nav link to user header if user handle is a wallet address.

- as per figma design.
- to direct the user to the explorer to purchase a zid.

### How do I test this?
- log in with a wallet and check the user handle as shown below.

#### Things to consider:
  1. Does this change any APIs?

Screenshot 2024-01-24 at 10 15 54

### DIFF
--- a/src/components/messenger/list/user-header/index.tsx
+++ b/src/components/messenger/list/user-header/index.tsx
@@ -25,6 +25,33 @@ export class UserHeader extends React.Component<Properties> {
     return this.props.userIsOnline ? 'active' : 'offline';
   }
 
+  get isWalletAddress() {
+    return this.props.userHandle.startsWith('0x');
+  }
+
+  renderLink() {
+    return (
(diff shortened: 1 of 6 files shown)


src/components/messenger/list/user-header/index.tsx
```diff
@@ -25,6 +27,33 @@ export class UserHeader extends React.Component<Properties> {
     return this.props.userIsOnline ? 'active' : 'offline';
   }
 
+  get isWalletAddress() {
+    return this.props.userHandle.startsWith('0x');
+  }
+
+  renderLink() {
+    return (
+      <a {...cn('link')} href={'https://explorer.zero.tech/'} target='_blank' rel='noopener noreferrer'>
```
dalefukami, line 36: I think we need to put the actual url into an environment variable.

PR #1435, opened 2024-01-24
Render primary ZID in conversation header

ratik21:
### What does this do?

Renders primary ZID OR wallet_address (or empty string if none exist) in conversation header.

Before:
image

After:
image

NOTE: For group conversations we still show online/offline status
image

### DIFF
--- a/src/components/messenger/chat/styles.scss
+++ b/src/components/messenger/chat/styles.scss
@@ -190,9 +190,11 @@ $recent-indicator-size: 8px;
     text-overflow: ellipsis;
     display: inline-block;
     font-weight: 400;
-    font-size: 12px;
-    line-height: 15px;
-    color: theme.$color-greyscale-11;
+    font-size: 10px;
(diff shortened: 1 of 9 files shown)


src/components/messenger/chat/styles.scss
```diff
@@ -190,9 +190,11 @@ $recent-indicator-size: 8px;
     text-overflow: ellipsis;
     display: inline-block;
     font-weight: 400;
-    font-size: 12px;
-    line-height: 15px;
-    color: theme.$color-greyscale-11;
+    font-size: 10px;
+    line-height: normal;
+    font-family: 'Roboto Mono';
```
domw30, line 195: Hey Ratik, this probably doesn't set the font-family to 'Roboto Mono' due to the underlying font settings. We'll need to update so we can use fonts like `Roboto Mono`

domw30 (approved):

PR #1437, opened 2024-01-24
Refactor matrix events, publish custom event when current users joins channel

ratik21:


src/lib/chat/matrix-client.ts
```diff
@@ -745,7 +747,7 @@ export class MatrixClient implements IChatClient {
 
     this.matrix.on(ClientEvent.Event, this.publishUserPresenceChange);
     this.matrix.on(RoomEvent.Name, this.publishRoomNameChange);
-    this.matrix.on(RoomStateEvent.Members, this.publishMembershipChange);
+    //this.matrix.on(RoomStateEvent.Members, this.publishMembershipChange);
```
dalefukami, line 750: Remove?

PR #1438, opened 2024-01-24
Remove `onUserReceivedInvitation` event

ratik21:
### What does this do?

We can remove this event (triggerred via `MemberShip.Invite`), since we're handling `currentUserAddedToChannel` via `MemberShip.Join` event


dalefukami (commented):
Yay!! So glad to have that full channel refetch removed!

PR #1440, opened 2024-01-24
Render PrimaryZID in group-info & group-edit panels

ratik21:
### What does this do?

Group Info Panel:
image

Group Edit panel:
image

### DIFF
--- a/src/components/citizen-list-item/styles.scss
+++ b/src/components/citizen-list-item/styles.scss
@@ -27,6 +27,15 @@
     white-space: nowrap;
   }
 
+  &__primary-zid {
+    font-size: 10px;
+    font-style: normal;
+    font-weight: 400;
+    display: flex;
+
+    @include glass-text-tertiary-color;
+  }
+
   &__remove {
     display: none;
   }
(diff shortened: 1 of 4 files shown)


src/components/citizen-list-item/styles.scss
```diff
@@ -25,6 +24,17 @@
     text-overflow: ellipsis;
     overflow: hidden;
     white-space: nowrap;
+
+    @include glass-text-primary-color;
+  }
+
+  &__primary-zid {
+    font-size: 10px;
+    font-style: normal;
```
domw30, line 33: Couple nits: Shouldn't need to set `font-style: normal`. `font-weight: 400` is default so don't need to explicitly set this either.

src/components/messenger/list/group-management/container.tsx
```diff
@@ -78,6 +79,7 @@ export class Container extends React.Component<Properties> {
         profileImage: currentUser?.profileSummary.profileImage,
         matrixId: currentUser?.matrixId,
         isOnline: currentUser?.isOnline,
+        primaryZID: getUserHandle(currentUser?.primaryZID, currentUser?.wallets),
```
domw30, line 82: We probably should have named this `userHandle` or something similar as it could be the wallet address. Maybe something to discuss when we're all on a call

domw30 (approved):

src/components/messenger/list/group-management/container.tsx
ratik21: I think it should just be `getPrimaryZID(user: User)`/ `getDisplayZID(user)`. 

Something similar to (in `src/lib/user`):
image

PR #1441, opened 2024-01-24
Render Primary ZID while mentioning users

ratik21:
### What does this do?

Renders primary ZID of user while mentioning users in a chat.

image

image

image

NOTE: I didn't update the UI much, as there is a task dedicated to mentions redesign in the backlog.

### DIFF
--- a/src/components/message-input/index.tsx
+++ b/src/components/message-input/index.tsx
@@ -231,7 +231,12 @@ export class MessageInput extends React.Component<Properties, State> {
         renderSuggestion={(suggestion) => (
           <>
             <Avatar size={'small'} type={'circle'} imageURL={suggestion.profileImage} />
-            <div className='message-input__mentions-text-area-wrap__suggestions__item-name'>{suggestion.display}</div>
+            <div>
+              <div className='message-input__mentions-text-area-wrap__suggestions__item-name'>{suggestion.display}</div>
+              <div className='message-input__mentions-text-area-wrap__suggestions__item-zid'>
+                {suggestion.primaryZID}
+              </div>
+            </div>
           </>
         )}
       />,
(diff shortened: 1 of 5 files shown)


src/components/message-input/index.tsx
```diff
@@ -231,7 +231,12 @@ export class MessageInput extends React.Component<Properties, State> {
         renderSuggestion={(suggestion) => (
           <>
             <Avatar size={'small'} type={'circle'} imageURL={suggestion.profileImage} />
-            <div className='message-input__mentions-text-area-wrap__suggestions__item-name'>{suggestion.display}</div>
+            <div>
+              <div className='message-input__mentions-text-area-wrap__suggestions__item-name'>{suggestion.display}</div>
+              <div className='message-input__mentions-text-area-wrap__suggestions__item-zid'>
+                {suggestion.primaryZID}
```
dalefukami, line 237: Is this supposed to be the handle? As in, if the zid does not exist should it be the wallet?

dalefukami (approved):

src/components/message-input/index.tsx
ratik21: Yes, this is the handle. Computed [here](https://github.com/zer0-os/zOS/blob/c9a794a17ca66eb3e580be34b42131914ae036b4/src/store/channels-list/utils.ts#L106). That basically sets the primaryZID for all "otherMembers" in a conversation (fetched OR gotten real-time).

If that's confusing, or doesn't make much sense then we can use `primaryZIDHandle` maybe? OR just pass on the `primaryZID` & `wallets` from the backend while mapping ZERO User, and the frontend component _compute_ the "handle". I thought about these two but then i got confused so just used `primaryZID` 😛

dalefukami: mmm... I see. Yeah, the field should either be `displayHandle` or it should be both fields and calculated on the fly. I'd prefer the first option. We shouldn't name fields the same as a concept that already exists but with a different meaning.

ratik21: okay, yes, makes sense. Will update in a follow up.

PR #1442, opened 2024-01-25
[Fix] Non Admin member (non-creator) should NOT be able to add new users to group

ratik21:
### What does this do?

- Updates `createConversation` flow to first create a room, and then invite the appropriate users. 
- Sets the default power levels for all users (including new users which add to room)

Earlier, if you add a new member to room, then that member can still add other users to the group (this is disabled via UI, but you can still call the API directly and it will add new members). This is fixed now.

If you try to add a new member now, and you're not an admin, you'll get this error:
image
 
 Similar code has been added on the API side as well.

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -396,7 +396,7 @@ export class MatrixClient implements IChatClient {
     const options: ICreateRoomOpts = {
       preset: Preset.TrustedPrivateChat,
       visibility: Visibility.Private,
-      invite: users.map((u) => u.matrixId),
+      invite: [],
       is_direct: true,
       initial_state,
       power_level_content_override: {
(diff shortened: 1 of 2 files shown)


ratik21:
I am not sure what to do about "existing conversations". Since we don't have the `power_level` events for the _newly added_ users, and we didn't had the defaults set in before..so they can still add members to the group..

src/lib/chat/matrix-client.ts
```diff
@@ -396,14 +396,20 @@ export class MatrixClient implements IChatClient {
     const options: ICreateRoomOpts = {
       preset: Preset.TrustedPrivateChat,
       visibility: Visibility.Private,
-      invite: users.map((u) => u.matrixId),
+      invite: [],
       is_direct: true,
       initial_state,
       power_level_content_override: {
         users: {
           ...users.reduce((acc, u) => ({ ...acc, [u.matrixId]: PowerLevels.Viewer }), {}),
```
dalefukami, line 404: Can remove this now?

dalefukami (approved):

dalefukami:
> I am not sure what to do about "existing conversations". Since we don't have the `power_level` events for the _newly added_ users, and we didn't had the defaults set in before..so they can still add members to the group..

Yeah, we might need to add another login bit that finds convos that you're an admin of and then modifies the room setting?

src/lib/chat/matrix-client.ts
ratik21: i left it in there as a "just in case"/"good to have". But yeah we can remove now. will do.

ratik21:
> Yeah, we might need to add another login bit that finds convos that you're an admin of and then modifies the room setting?

Yeah, something along those lines 🤔

PR #1445, opened 2024-01-27
Removing `decodeURIComponent` (to decode `%23` from the conversation URL if the `conversationId` is an alias)

ratik21:
### What does this do?

Removes having to use `%23` in the conversation url (which was needed to parse the alias before). Now, you don't need to use any encoded char before the alias string. 

The logic is update to check if the id starts with `!`, then it's a valid (unique) ID, otherwise it's an alias.

### Usage (for eg. alias = `#hite:zos-dev.zer0.io`)

Before you'll hit this url:
`http://localhost:3002/conversation/%23hite:zos-dev.zer0.io`

Now the url updates to:
`http://localhost:3002/conversation/hite:zos-dev.zer0.io`

Screenshot
image

### DIFF
--- a/src/store/chat/saga.ts
+++ b/src/store/chat/saga.ts
@@ -14,7 +14,7 @@ import { getHistory } from '../../lib/browser';
 import { openFirstConversation } from '../channels/saga';
 import { rawConversationsList, waitForChannelListLoad } from '../channels-list/saga';
 import { featureFlags } from '../../lib/feature-flags';
-import { translateJoinRoomApiError } from './utils';
+import { translateJoinRoomApiError, parseAlias, isAlias } from './utils';
 import { joinRoom as apiJoinRoom } from './api';
 
 function* initChat(userId, chatAccessToken) {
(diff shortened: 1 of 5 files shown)


src/store/chat/saga.ts
```diff
@@ -153,6 +148,7 @@ export function* performValidateActiveConversation(activeConversationId: string)
 
   let conversationId = activeConversationId;
   if (isAlias(activeConversationId)) {
+    activeConversationId = parseAlias(activeConversationId);
```
ratik21, line 151: this will just prefix the id with `#` before proceeding, so everything else (save to store, calling api etc) remains the same

src/store/chat/utils.ts
```diff
@@ -37,3 +37,24 @@ export function translateJoinRoomApiError(errorCode: JoinRoomApiErrorCode | stri
...
+
+// conversation can be referenced by an id or an alias
+export function isAlias(id: string) {
+  const isAlias = !id.startsWith('!');
+  if (isAlias) {
+    id = '#' + id;
+    return true;
+  }
+
+  return false;
```
dalefukami, line 49:
```suggestion
  return !id.startsWith('!');
```

Pretty sure the statement setting the variable isn't doing anything (and shouldn't because this is just a boolean check function)

dalefukami (approved):

src/store/chat/utils.ts
ratik21: 🤦‍♂️ i forgot to update this. Had added `parseAlias`. Thanks.

PR #1446, opened 2024-01-29
feat: handle search result filtering to include primary zid direct and indirect matches

domw30:
### What does this do?
- handles primary zid when searching and filtering conversations.

- to ensure a users `primaryZID` is considered when searching and filtering in the conversation list.

### How do I test this?
- you will need to have test users that have a `primaryZID` which will allow you to search using a ZID in the search input and render results.

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/messenger/list/conversation-list-panel/index.tsx
+++ b/src/components/messenger/list/conversation-list-panel/index.tsx
@@ -68,6 +68,29 @@ export class ConversationListPanel extends React.Component<Properties, State> {
     this.setState({ userSearchResults: filteredItems?.map(itemToOption) });
   };
 
+  // Checks if the conversation's name matches the search filter.
+  isFilterNameMatch(conversation, searchRegEx) {
+    return searchRegEx.test(conversation.name ?? '');
+  }
+
+  // Determines if there's a direct match in a one-on-one conversation based on name or primaryZID.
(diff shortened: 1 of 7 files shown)


src/components/messenger/list/conversation-list-panel/index.tsx
```diff
@@ -68,32 +68,46 @@ export class ConversationListPanel extends React.Component<Properties, State> {
     this.setState({ userSearchResults: filteredItems?.map(itemToOption) });
   };
 
+  // Checks if the conversation's name matches the search filter.
```
dalefukami, line 71: I feel like maybe there's some improvement that could be made here. Generally speaking, if you have to describe what the function is doing with a comment then that means the code isn't clear enough (not a global rule...but a reasonable guideline). Can we come up with some better names or more refactoring to clear this up?

dalefukami (approved):

src/components/messenger/list/conversation-list-panel/index.tsx
domw30: Hi Dale. Had a go at some refactoring here. Decided to move helpers into a utils file to keep the conversation list panel code a bit cleaner. Also updated the names of the helpers to improve readability and thus remove the need for comments. Let me know what you think of the updates. Thanks!

PR #1452, opened 2024-01-31
feat: add joining conversation loading state to chat saga

domw30:
### What does this do?
- adds loading state when validating joining a conversation.

- to communicate to the user the following : there will be a delay between loading the conversation list and displaying the conversation a user has been linked to. During this delay, the app is checking whether the URL is valid (does conversation exist) and if the user is a member (whether you are already added to the group or, if it is a TGC, whether you have the required token). 

### How do I test this?
- pass state to component level and log the boolean | alternatively add a basic ui component - then navigate to a conversation url.

#### Things to consider:
  1. Does this change any APIs?

When passed to component level:

https://github.com/zer0-os/zOS/assets/39112648/e8f5face-b3c4-43c3-a43e-6b2536f6e6d0

### DIFF
--- a/src/store/chat/saga.ts
+++ b/src/store/chat/saga.ts
@@ -135,8 +135,6 @@ export function* setWhenUserJoinedRoom(conversationId: string) {
 }
 
 export function* performValidateActiveConversation(activeConversationId: string) {
-  yield put(setIsJoiningConversation(true));
-
   if (!activeConversationId) {
     yield put(clearJoinRoomErrorContent());
     yield call(openFirstConversation);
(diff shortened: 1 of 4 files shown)


src/store/chat/saga.ts
```diff
@@ -125,6 +135,8 @@ export function* setWhenUserJoinedRoom(conversationId: string) {
 }
 
 export function* performValidateActiveConversation(activeConversationId: string) {
+  yield put(setIsJoiningConversation(true));
```
dalefukami, line 138: Why do we need to set these again in this saga? Nothing else should be calling this aside from the `validateActiveConversation` saga that's controlling the primary flow, right?

dalefukami (approved):

PR #1453, opened 2024-01-31
feat: add joining conversation dialog component

domw30:
### What does this do?
- adds a joining conversation dialog component.

- to import and use the component in `MessengerChat`.

### How do I test this?
- this is simply adding a component to the code base. no manual testing required here.

#### Things to consider:
  1. Does this change any APIs?

Screenshot 2024-01-31 at 16 56 57

### DIFF
--- a/src/components/joining-conversation-dialog/index.tsx
+++ b/src/components/joining-conversation-dialog/index.tsx
@@ -1,0 +1,21 @@
+import * as React from 'react';
+
+import { IconLock1 } from '@zero-tech/zui/icons';
+
+import { bemClassName } from '../../lib/bem';
+
+import './styles.scss';
+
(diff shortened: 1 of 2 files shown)


src/components/joining-conversation-dialog/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import { IconLock1 } from '@zero-tech/zui/icons';
+
+import { bemClassName } from '../../lib/bem';
+
+import './styles.scss';
+
+const cn = bemClassName('joining-conversation-dialog');
+
+export class JoiningConversationDialog extends React.Component {
```
dalefukami, line 11: "Dialog" seems like a weird name for this but I don't have a better proposal.

PR #1457, opened 2024-01-31
Fetch a list of owned ZID's while editing a user's profile

ratik21:
### What does this do?

Adds saga logic for fetching ownedZID's when the user opens the edit profile modal.

### DIFF
--- a/src/store/edit-profile/saga.test.ts
+++ b/src/store/edit-profile/saga.test.ts
@@ -1,13 +1,14 @@
 import { expectSaga } from 'redux-saga-test-plan';
 import { call } from 'redux-saga/effects';
-import { editProfile as editProfileSaga, updateUserProfile } from './saga';
-import { editUserProfile as apiEditUserProfile } from './api';
+import { editProfile as editProfileSaga, updateUserProfile, fetchOwnedZIDs } from './saga';
+import { editUserProfile as apiEditUserProfile, fetchOwnedZIDs as apiFetchOwnedZIDs } from './api';
 import { uploadImage } from '../registration/api';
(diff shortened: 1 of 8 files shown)


src/store/edit-profile/saga.test.ts
```diff
@@ -140,6 +141,27 @@ describe('updateUserProfile', () => {
...
+
+    const {
+      storeState: { editProfile },
+    } = await expectSaga(fetchOwnedZIDs)
+      .provide([
+        [
+          call(apiFetchOwnedZIDs),
+          ownedZIDs,
+        ],
+      ])
```
dalefukami, line 156: FYI, if you remove the trailing commas the formatter will pull this onto a single line. It reads much better.

src/store/registration/index.ts
```diff
@@ -55,6 +55,7 @@ export enum ProfileDetailsErrors {
   UNKNOWN_ERROR = 'UNKNOWN_ERROR',
   NAME_REQUIRED = 'NAME_REQUIRED',
   FILE_UPLOAD_ERROR = 'FILE_UPLOAD_ERROR',
+  FetchOwned_ZIDs_ERROR = 'FetchOwned_ZIDs_ERROR',
```
dalefukami, line 58: Any reason you deviated from the convention of the rest of the enum here?

dalefukami (approved):

src/store/edit-profile/saga.test.ts
ratik21: got it, thanks

src/store/registration/index.ts
ratik21: 😭 updated

PR #1458, opened 2024-02-01
fix: workflow environment variable naming to match with github (zns explorer url)

domw30:
### What does this do?
- updates zns explorer url environment variable naming to match that on github env variables.

- should fix issue where the environment variable is not being set (due to mismatch of naming).

### How do I test this?
- on development or production we expect the use of this variable to direct you to the explorer url. this will need to be deployed to test.

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -44,7 +44,7 @@ jobs:
       REACT_APP_ASSETS_PATH: ${{vars.ASSETS_PATH}}
       REACT_APP_ANDROID_STORE_PATH: ${{vars.ANDROID_STORE_PATH}}
       REACT_APP_MATRIX_HOME_SERVER_URL: ${{vars.MATRIX_HOME_SERVER_URL}}
-      REACT_APP_ZNS_EXPLORER_URL: ${{vars.ZNS_EXPLORER_URL}}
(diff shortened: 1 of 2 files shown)


.github/workflows/development.yml
```diff
@@ -44,7 +44,7 @@ jobs:
       REACT_APP_ASSETS_PATH: ${{vars.ASSETS_PATH}}
       REACT_APP_ANDROID_STORE_PATH: ${{vars.ANDROID_STORE_PATH}}
       REACT_APP_MATRIX_HOME_SERVER_URL: ${{vars.MATRIX_HOME_SERVER_URL}}
-      REACT_APP_ZNS_EXPLORER_URL: ${{vars.ZNS_EXPLORER_URL}}
+      REACT_APP_ZNS_EXPLORER_URL: ${{vars.REACT_APP_ZNS_EXPLORER_URL}}
```
dalefukami, line 47: Would have been better to fix the github var to be consistent with the pattern of the other variables.

domw30: Yeah agreed, I also need to remove leading '/' on the sepolia explorer path so I'll update these this afternoon as well

PR #1463, opened 2024-02-02
Add Dropdown to select primary_zid in `edit profile` screen

ratik21:
### What does this do?

image

image

### DIFF
--- a/src/components/edit-profile/index.test.tsx
+++ b/src/components/edit-profile/index.test.tsx
@@ -94,11 +94,30 @@ describe('EditProfile', () => {
     expect(saveButton(wrapper).prop('isDisabled')).toEqual(false);
   });
 
+  it('renders dropdown with ownedZIDs with the primaryZID as the first option', () => {
+    featureFlags.allowEditPrimaryZID = true;
+
+    const wrapper = subject({
+      currentPrimaryZID: '0://jane:smith',
+      ownedZIDs: ['0://john:doe', '0://jane:smith', '0://jijitsu:kaezen'],
+    });
+
+    const dropdown = wrapper.find('SelectInput[label=""]');
+    const items: any = dropdown.prop('items');
+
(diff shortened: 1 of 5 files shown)


ratik21:
Update: Added scrollbar
image

src/components/edit-profile/index.test.tsx
```diff
@@ -94,11 +94,30 @@ describe('EditProfile', () => {
...
 
+  it('renders dropdown with ownedZIDs with the primaryZID as the first option', () => {
+    featureFlags.allowEditPrimaryZID = true;
+
+    const wrapper = subject({
+      currentPrimaryZID: '0://jane:smith',
+      ownedZIDs: ['0://john:doe', '0://jane:smith', '0://jijitsu:kaezen'],
+    });
+
+    const dropdown = wrapper.find('SelectInput[label=""]');
```
dalefukami, line 105: Is there more than one SelectInput at a time on this screen? I.E., do we need the `label` filter?

src/index.scss
```diff
@@ -107,5 +107,5 @@ hr.glow:before {
 // and add a permanent node at the root level and set it as the popup container.
 // Using this method for now due to time constraints.
 [data-radix-popper-content-wrapper] {
-  z-index: 9999 !important;
+  z-index: 10001 !important;
```
dalefukami, line 110: Is something else a higher z-index on our page?! ugggg

ratik21: yeaaahhh i need this to display the "dropdown" on top the "modal" (which has 9999 z-index set if i remember correctly) 😢

src/components/edit-profile/index.test.tsx
ratik21: we don't here, but i had to add it on the component side (`label = ''`), as it is a required field in the zUI component

PR #1470, opened 2024-02-05
deps: upgrade to node v20.11.0 (LTS)

thewarman:
### What does this do?
This is PR 1/3.
Update the toolchain references to node v20.11 (LTS) release.
Note that webpack v4 utilizes a deprecated crypto api (for generating asset hashes) and will cause node to error with a `ERR_OSSL_EVP_UNSUPPORTED` error. The workaround is to specify `NODE_OPTIONS='--openssl-legacy-provider'`.

In preparation for required toolchain dependencies for Electron desktop app.

### How do I test this?
Run development / tests as usual.

#### Things to consider:
  1. Does this change any APIs?


dalefukami (changes requested):
Failing CI.

thewarman:
Might have to run #1471 instead as that includes the required updates to GHA.

PR #1471, opened 2024-02-05
deps: upgrade to node v20.11.0 (LTS) + update GHA v4

thewarman:
### What does this do?
Update the toolchain references to node v20.11 (LTS) release.

Note that webpack v4 utilizes a deprecated crypto api (for generating asset hashes) and will cause node to error with a ERR_OSSL_EVP_UNSUPPORTED error. The workaround is to specify NODE_OPTIONS='--openssl-legacy-provider'.

Follow up commit updates GHA workflows to use node v20 and actions v4.
 
To support node v20 update.

### How do I test this?
Run GHA deploys / workflows as usual. 

#### Things to consider:
  1. Does this change any APIs?


dalefukami:
Looks ok. Just waiting for confirmation from our admin team that it's ok for me to accept submissions and then I'll merge this.

PR #1472, opened 2024-02-05
deps: webpack v5 tooling + polyfill and import fixes 

thewarman:
### What does this do?
Follow up to #1471.

Upgrades `react-scripts` and `craco` to webpack v5 based versions. v5 webpack no longer polyfills node.js dependencies so they have to be manually specified.

Relevant docs on polyfill / resolve.fallback:
- https://webpack.js.org/configuration/resolve/#resolvefallback
- https://web3auth.io/docs/troubleshooting/webpack-issues#react-create-react-app
- https://docs.cloud.coinbase.com/wallet-sdk/docs/web3-react

`react-scripts` now also generates noisy warnings about source maps in node_modules so we silence that warning.
- https://github.com/facebook/create-react-app/discussions/11767

Other changes include updating `web3-utils` as previous version did not correctly export a default object. This small change is here in this [commit](https://github.com/zer0-os/zOS/pull/1466/commits/131e7a72cba2d6bb97b21c3b820a74a1a3333852).

In preparation for required toolchain dependencies for Electron desktop app.

### How do I test this?
Run development / tests as usual.

#### Things to consider:
  1. Does this change any APIs?


craco.config.js
```diff
@@ -1,15 +1,34 @@
...
+        // https://webpack.js.org/configuration/resolve/#resolvefallback
+        // https://web3auth.io/docs/troubleshooting/webpack-issues#react-create-react-app
+        // https://docs.cloud.coinbase.com/wallet-sdk/docs/web3-react
+        fallback: {
+          crypto: false,
+          fs: false,
+          buffer: require.resolve('buffer'),
+          stream: require.resolve('stream-browserify'),
+          util: require.resolve('util/'),
+          assert: require.resolve('assert/'),
```
dalefukami, line 18: Is there a reason to have these fallbacks? I tried a build with these all set to `false` and I haven't found any issues yet. What issues were you running into?

thewarman: The current web3 login needs them to function correctly. The app builds without it but when you try to open the wallet logins (specifically wallet connect or coinbase options), they error out.

Ideally libraries shouldn't need these fallbacks as it is considered bad practice and instead should use browser native functionality. I'd recommend later updating or changing the web3 login libraries and should be able to remove the fallbacks.

dalefukami: So the other ones work without the polyfills?

Why don't we need this too? It seems to be pointed out in multiple places but we only added `Buffer`.

```
new webpack.ProvidePlugin({
	// Make a global `process` variable that points to the `process` package,
	// because the `util` package expects there to be a global variable named `process`.
       // Thanks to https://stackoverflow.com/a/65018686/14239942
	process: 'process/browser'
   })
```

dalefukami:
image

Is there an argument for not cleaning this warning up as it suggests?

craco.config.js
```diff
@@ -20,12 +39,15 @@ module.exports = {
...
-      ignored: '**/node_modules/**',
+    plugins: {
+      add: [
+        new webpack.ProvidePlugin({
+          Buffer: ['buffer', 'Buffer'],
+        }),
+      ],
     },
   },
+  devServer: {},
```
dalefukami, line 50: Why remove this? Is the option no longer available?

thewarman: The option is different now and from what I can tell in the guts of create-react-app, it ignores node_modules by default.

craco.config.js
thewarman: It appears to work without it - from my testing the `ProvidePlugin` is needed for WalletConnect but not Coinbase.

The other wallet integrations seem to be working fine but they error out for other configuration errors that seem unrelated (can try current staging environment to compare). The polyfills fundamentally are needed for `web3-react` in general so think it applies to all connectors.

thewarman:
> image
> 
> Is there an argument for not cleaning this warning up as it suggests?

I misread this error but yeah should be an easy fix to just add this missing dependency.

craco.config.js
dalefukami: Just curious where you saw that it was different now. This page seems to describe the option exactly as it was set: https://webpack.js.org/configuration/watch/#watchoptionsignored

thewarman: It throws an error about an invalid configuration if you try to set it. It looks like it moved to be a general webpack option while the above setting is for the `devServer` specifically. There is `devServer` watchFiles but no ignore.

`craco` can be confusing - not sure where you would set `watch` options, under the `configure` object maybe?

craco.config.js
dalefukami: Alright, we ended up removing all options aside from `Metamask` and `Coinbase` so that limits our test maintenance area (considering most of the others didn't work at all).

dalefukami:
Closing until #1471 is merged and this can be updated.

dalefukami:
@thewarman #1471 has been merged. This PR can now be rebased and cleaned up and then we'll take a look into it again.

PR #1473, opened 2024-02-05
fix: handle token gated join room error navigation

domw30:
### What does this do?
- appends the extracted domain to the explorer url if activeConversationId is an alias.

- this ensures the user will be directed to the explorer - to a specific world.

### How do I test this?
- attempt to visit a room url (alias not id) > click 'Buy A Domain' link in error dialog > check explorer url (as shown below)

#### Things to consider:
  1. Does this change any APIs?

DEMO:

https://github.com/zer0-os/zOS/assets/39112648/e7bd9480-6d76-45ba-805e-6877f93b8975

### DIFF
--- a/src/store/chat/utils.ts
+++ b/src/store/chat/utils.ts
@@ -31,9 +31,10 @@ export const ERROR_DIALOG_CONTENT = {
 export function translateJoinRoomApiError(errorCode: JoinRoomApiErrorCode | string, domain?: string) {
   const content = ERROR_DIALOG_CONTENT[errorCode] || ERROR_DIALOG_CONTENT[JoinRoomApiErrorCode.UNKNOWN_ERROR];
 
-  if (errorCode === JoinRoomApiErrorCode.ACCESS_TOKEN_REQUIRED) {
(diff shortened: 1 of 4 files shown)


src/store/chat/utils.ts
```diff
@@ -28,11 +28,11 @@ export const ERROR_DIALOG_CONTENT = {
   },
 };
 
-export function translateJoinRoomApiError(errorCode: JoinRoomApiErrorCode | string, roomAlias?: string) {
+export function translateJoinRoomApiError(errorCode: JoinRoomApiErrorCode | string, domain?: string) {
   const content = ERROR_DIALOG_CONTENT[errorCode] || ERROR_DIALOG_CONTENT[JoinRoomApiErrorCode.UNKNOWN_ERROR];
 
-  if (content.linkPath && content.linkPath.includes('{roomAlias}')) {
-    content.linkPath = content.linkPath.replace('{roomAlias}', roomAlias);
+  if (errorCode === JoinRoomApiErrorCode.ACCESS_TOKEN_REQUIRED && content.linkPath?.includes('{domain}')) {
```
dalefukami, line 34: Just as an FYI, you don't really need the `includes` check because replace will just do nothing if it doesn't find the string.

PR #1476, opened 2024-02-06
refactor: move get user handle to messenger lib utils folder

domw30:
### What does this do?
- moves method higher up to libs folder in messenger 

- we'll be using this util outside of the `list` folder, therefore moving higher up in the `messenger` tree.

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/messenger/lib/utils.tsx
+++ b/src/components/messenger/lib/utils.tsx
@@ -1,5 +1,6 @@
 import { Item, Option } from './types';
 import { Channel, User } from '../../../store/channels';
+import { Wallet } from '../../../store/authentication/types';
 
 export const itemToOption = (item: Item): Option => {
   return {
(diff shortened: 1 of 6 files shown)


ratik21 (approved):

src/components/messenger/lib/utils.tsx
```diff
@@ -34,3 +35,17 @@ export const highlightFilter = (text, filter) => {
 
   return text;
 };
+
+export function getUserHandle(primaryZID: string, wallets: Wallet[]) {
```
dalefukami, line 39: At some point it'd be good to change this to `getUserSubHandle` since we use their name as their handle.

src/store/channels-list/utils.ts
```diff
@@ -4,7 +4,7 @@ import { denormalize } from './../channels/index';
 import getDeepProperty from 'lodash.get';
 import { select } from 'redux-saga/effects';
 import { currentUserSelector } from '../authentication/selectors';
-import { getUserHandle } from '../../components/messenger/list/utils/utils';
+import { getUserHandle } from '../../components/messenger/lib/utils';
```
dalefukami, line 7: Redux stores shouldn't reference component code. Seems like we moved this to the wrong place. If it's not component specific code it should be in a more general lib folder.

PR #1477, opened 2024-02-06
feat: add subLabel, primaryZID and wallets props to option and item type

domw30:
### What does this do?
- adds `subLabel` prop to `Option` type.
- adds `primaryZID` prop to `Item` type.
- adds `wallets` prop to `Item` type.

- adding these props enables us to set either the `primaryZID` or users `publicWalletAddress` to the subLabel prop.
- display the `subLabel` on user search result items.

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/messenger/lib/types.ts
+++ b/src/components/messenger/lib/types.ts
@@ -1,11 +1,16 @@
+import { Wallet } from '../../../store/authentication/types';
+
 export interface Item {
   id: string;
   name: string;
   image?: string;
+  primaryZID?: string;
+  wallets?: Wallet[];
 }
 
 export interface Option {
   value: string;
   label: string;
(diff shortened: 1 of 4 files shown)


src/components/messenger/lib/types.ts
```diff
@@ -1,11 +1,16 @@
+import { Wallet } from '../../../store/authentication/types';
+
 export interface Item {
   id: string;
   name: string;
   image?: string;
+  primaryZID?: string;
+  wallets?: Wallet[];
```
domw30, line 8: NOTE: In order to display the users public wallet address, some follow up changes will be required in the backend to return the users wallets in the search results.

ratik21 (approved):

PR #1478, opened 2024-02-06
feat(user-search-results-component): expose sub-label to search result items

domw30:
### What does this do?
- exposes `subLabel` to search result items for the `UserSearchResults` component.

- to display the users `primaryZID` (or, after some follow up changes, the users `publicWalletAddress`).

### How do I test this?
- on the conversation list panel, search users that have a primaryZID set and check it is displayed as per screenshot below.

#### Things to consider:
  1. Does this change any APIs?

Sub-Label:

Screenshot 2024-02-06 at 11 54 51

### DIFF
--- a/src/components/messenger/list/user-search-results/index.test.tsx
+++ b/src/components/messenger/list/user-search-results/index.test.tsx
@@ -23,9 +23,9 @@ describe('UserSearchResults', () => {
 
   it('renders user search results', function () {
     const userResults = [
-      { value: 'user-1', label: 'jack', image: 'image-1' },
-      { value: 'user-2', label: 'bob', image: 'image-2' },
-      { value: 'user-3', label: 'jacklyn', image: 'image-3' },
+      { value: 'user-1', label: 'jack', image: 'image-1', subLabel: '0://jack.test' },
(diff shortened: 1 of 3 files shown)


src/components/messenger/list/user-search-results/index.test.tsx
```diff
@@ -18,35 +23,54 @@ describe('UserSearchResults', () => {
   it('renders the title', function () {
     const wrapper = subject({});
 
-    expect(wrapper.find('.user-search-results__title').text()).toEqual('Start a new conversation:');
+    expect(wrapper.find(c('title')).text()).toEqual('Start a new conversation:');
```
ratik21, line 26: 🌟

ratik21 (approved):

src/components/messenger/list/user-search-results/index.test.tsx
dalefukami: bonus would be `toHaveText` :)

domw30: > bonus would be `toHaveText` :)

https://github.com/zer0-os/zOS/pull/1478/commits/13934accd2b52b1aa819146d266df10f118e234b

😉

PR #1480, opened 2024-02-06
fix(citizen-list-item): details, tag and remove icon positioning

domw30:
### What does this do?
- fixes styles of the `CitizenListItem` that appear to have broken/been changed at some point recently.
- some refactoring to tests.

- as per figma designs. before and after below.

### How do I test this?
- check group info panel or edit group panel.

#### Things to consider:
  1. Does this change any APIs?

Before:
Screenshot 2024-02-06 at 13 10 20

After:
Screenshot 2024-02-06 at 13 10 35

### DIFF
--- a/src/components/citizen-list-item/index.test.tsx
+++ b/src/components/citizen-list-item/index.test.tsx
@@ -19,19 +19,25 @@ describe(CitizenListItem, () => {
   it('renders the user display name', function () {
     const wrapper = subject({ user: { firstName: 'John', lastName: 'Doe' } as any });
 
-    expect(wrapper.find(c('name')).text()).toEqual('John Doe');
+    const name = wrapper.find(c('name'));
+
+    expect(name).toHaveText('John Doe');
   });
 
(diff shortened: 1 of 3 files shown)


src/components/citizen-list-item/index.test.tsx
```diff
@@ -19,19 +19,25 @@ describe(CitizenListItem, () => {
   it('renders the user display name', function () {
     const wrapper = subject({ user: { firstName: 'John', lastName: 'Doe' } as any });
 
-    expect(wrapper.find(c('name')).text()).toEqual('John Doe');
+    const name = wrapper.find(c('name'));
+
+    expect(name).toHaveText('John Doe');
```
dalefukami, line 24: My personal preference would be to inline this variable. "Finding the node" isn't really part of the test so having it as a separate step is slightly confusing. Not worth changing just pointing out my thoughts.

```
expect(wrapper.find(c('name'))).toHaveText('John Doe');
```

PR #1486, opened 2024-02-07
feat: include wallet priamry wallet address in received search results

domw30:
### What does this do?
- wires up wallet primary address when receiving search results.

- to display the wallet primary address in the UI handled by `getUserHandle`.

### How do I test this?
- search users that do not have a zid set, and have a primary wallet address associated with the account.

#### Things to consider:
  1. Does this change any APIs?

Screenshot 2024-02-08 at 12 40 45


src/store/users/saga.test.ts
```diff
@@ -28,8 +28,20 @@ describe(clearUsers, () => {
...
+      profileImage: 'image-url-1',
+      primaryZID: 'zid-1',
+      wallets: [{ publicAddress: 'address-1' }],
+    };
+    const user2 = {
+      id: 'user-2',
+      name: 'Test User 2',
+      profileImage: 'image-url-2',
+      primaryZID: 'zid-2',
+      wallets: [{ publicAddress: 'address-2' }],
```
ratik21, line 43: if a user has a primary_zid set, with a wallet address. then we'll still display the zid in the `subLabel` right (in the search results in the UI)?

ratik21 (commented):
Could you share a screenshot of how the wallet_address would look like in a subLabel in the UI?

src/store/users/saga.test.ts
domw30: Yeah this is just setting the users record (which I've now changed slightly to set `displaySubHandle` rather than the wallets, to align with the work you had previously done.

```
export function* receiveSearchResults(searchResults) {
  const existingUsers = (yield select((state) => state.normalized.users)) ?? {};
  const existingUserIds = Object.keys(existingUsers);

  const mappedUsers = searchResults
    .filter((r) => !existingUserIds.includes(r.id))
    .map((r) => {
      return {
        userId: r.id,
        firstName: r.name,
        profileImage: r.profileImage,
        matrixId: r.matrixId,
        primaryZID: r.primaryZID,
        displaySubHandle: getUserHandle(r.primaryZID, r.wallets?.[0]),
      };
    });
  yield put(receive(mappedUsers));
}
```

As for what will be displayed, as we use `getUserHandle` this will determine if its displays the zid or address.

Screenshot 2024-02-08 at 12 29 28

src/components/messenger/lib/utils.tsx
```diff
@@ -39,15 +38,13 @@ export const highlightFilter = (text, filter) => {
   return text;
 };
 
-export function getUserHandle(primaryZID: string, wallets: Wallet[]) {
+export function getUserHandle(primaryZID: string, primaryWallet: string) {
```
domw30, line 41: Note: I have updated `getUserHandle` to just take in a single wallet address rather than an array of wallets. As a result, I have updated whats being passed in to `getUserHandle` in other areas it was being used.

src/store/users/saga.test.ts
ratik21: got it. Looks great!

src/components/messenger/lib/utils.tsx
ratik21: a side question but how do we compute the `primaryWallet`? Is it simply the first wallet address in the list of `wallets:[]`? 🤔

ratik21 (approved):

src/components/messenger/lib/utils.tsx
domw30: Yeah first wallet in list. We've never added any logic to determine which wallet would be the one to display (set as `primaryWallet`. We should really copy how we set a default primaryZID for the user on registration, then we'll allow the user to set a default if they wish to do so (like in your edit profile work). The user should be able to do the same with wallets. If that makes sense ?

dalefukami: Doesn't the database have a flag that says if it's the user's primary wallet?

domw30: oh yes there is! Okay nice, maybe we should just include that in the search results instead then

ratik21: makes sense

PR #1487, opened 2024-02-08
Render customIcon first as the conversationImage header (if set)

ratik21:
### What does this do?

This gives priority to the "custom icon" set in the conversation data received from matrix. 

For eg. if there's a `oneOnOne` conversation with a `"icon"` value set, then we'll display that icon instead of the `profileImage` of the "other member".

### Why was this required?
This was required to display the conversationImage in a token-gated-chat, when there's only you & admin in the room. Instead of displaying the admin's profileImage, we want to display the conversationImage set in the "conversation" by the backend.

image

### DIFF
--- a/src/components/messenger/list/conversation-item/index.tsx
+++ b/src/components/messenger/list/conversation-item/index.tsx
@@ -58,28 +58,11 @@ export class ConversationItem extends React.Component<Properties> {
   };
 
   renderAvatar() {
+    let imageUrl;
     if (isCustomIcon(this.props.conversation.icon)) {
-      return (
-        <Avatar
-          size={'regular'}
-          type={'circle'}
-          imageURL={this.props.conversation.icon}
-          statusType={this.conversationStatus}
-          tabIndex={-1}
-          isRaised
-        />
-      );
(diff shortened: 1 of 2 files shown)


src/components/messenger/list/conversation-item/index.tsx
```diff
@@ -58,23 +58,23 @@ export class ConversationItem extends React.Component<Properties> {
   };
 
   renderAvatar() {
-    if (this.props.conversation.isOneOnOne && this.props.conversation.otherMembers[0]?.profileImage) {
+    if (isCustomIcon(this.props.conversation.icon)) {
```
dalefukami, line 61: Looks like we have 3 identical renderings of the Avatar component aside from the url.... bonus points to refactor this to only do it once. lol

dalefukami (approved):
Ah, yeah, the exception to the rule if you're the first one in. haha

src/components/messenger/list/conversation-item/index.tsx
ratik21: done, thanksss

PR #1497, opened 2024-02-09
chore: update package-lock.json

domw30:
chore: update package-lock.json


dalefukami (approved):
Would be nice to know what these changes are and why they're suddenly required.

domw30:
> Would be nice to know what these changes are and why they're suddenly required.

I wasn't able to pinpoint why these are now required, no changes in package.json dev dependencies.

PR #1505, opened 2024-02-12
Display loading indicator while fetching owned_zid's of user (during edit profile)

ratik21:
### What does this do?

image

https://github.com/zer0-os/zOS/assets/33264364/80c819c3-d2e0-454f-9223-819a41253a6a

### DIFF
--- a/src/components/edit-profile/container.test.tsx
+++ b/src/components/edit-profile/container.test.tsx
@@ -14,6 +14,7 @@ describe('Container', () => {
       currentProfileImage: 'profile.jpg',
       currentPrimaryZID: '0://john:doe',
       ownedZIDs: [],
+      loading: false,
       editProfile: () => null,
       startProfileEdit: () => null,
       leaveGlobalNetwork: () => null,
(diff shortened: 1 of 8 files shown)


src/components/edit-profile/container.test.tsx
```diff
@@ -14,6 +14,7 @@ describe('Container', () => {
       currentProfileImage: 'profile.jpg',
       currentPrimaryZID: '0://john:doe',
       ownedZIDs: [],
+      loading: false,
```
domw30, line 17: Just a nit, loading seems a bit vague. I'd have to dig into the saga to check what is actually loading if I didn't have the context.

ratik21: most saga's have this `loading` prop to indicate "loading" state. If not clear, we can also use `EditProfile.State.IS_LOADING` (i.e use the `State` enum)

dalefukami: I think Dom's point is that this flag is not indicating that we're loading "the profile". It's very specific in that we're loading the zids. What if we had different data to load that could be rendered early, for example.

PR #1506, opened 2024-02-13
Render `NONE (wallet address)` in dropdown while editing primary zid

ratik21:
### What does this do?

- allows the user to set primary_zid as "none"

https://github.com/zer0-os/zOS/assets/33264364/cd2df87a-6fc1-4e32-a564-0c56adee118a

### DIFF
--- a/src/components/edit-profile/index.tsx
+++ b/src/components/edit-profile/index.tsx
@@ -47,7 +47,7 @@ export class EditProfile extends React.Component<Properties, State> {
     this.props.onEdit({
       name: this.state.name,
       image: this.state.image,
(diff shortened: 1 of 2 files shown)


src/components/edit-profile/index.tsx
```diff
@@ -113,6 +113,14 @@ export class EditProfile extends React.Component<Properties, State> {
     );
   }
 
+  getNoneOption() {
+    return {
+      id: 'None (wallet address)',
```
dalefukami, line 118: I think some sort of constant is more appropriate for an id. ZID_NONE or something.

PR #1522, opened 2024-02-15
feat: trigger verify id dialog from the messenger lists user header

domw30:
### What does this do?
- opens the verify id dialog from the messenger lists user header `Verify ID` button.

- to provide more information to the user by opening a modal containing more information on verifying their id.

### How do I test this?
- log in with a user via a wallet, and that does not have a primaryZID, enable verify id feature flag and click the `Verify ID` button.

#### Things to consider:
  1. Does this change any APIs?

https://github.com/zer0-os/zOS/assets/39112648/35b1023b-7742-48d1-a8b1-cf2931de25ad


src/components/messenger/list/user-header/styles.scss
```diff
@@ -37,13 +37,22 @@
...
     font-size: 10px;
     line-height: 13px;
   }
 
-  &__link {
-    color: theme.$color-secondary-11;
+  &__verify-id-button {
+    height: fit-content;
+
+    > *:first-child {
```
domw30, line 48: Bit of a quick hack to improve ui until we update 'fiddly' buttons

PR #1527, opened 2024-02-21
feat: add manage secure backup prompt feature flag

domw30:
### What does this do?
- adds feature flag.

- to use in follow up secure backup prompt flow.

#### Things to consider:
  1. Does this change any APIs?


dalefukami (approved):
This is fine, just wondering if there's a need for this. Are any of the changes we're making drastic enough that we can't just deploy in pieces and it doesn't matter?

PR #1533, opened 2024-02-22
feat: wire up ui for rendering backup dialog when user has not performed key backup

domw30:
### What does this do?
- renders backup dialog when a user has not performed key backup and has sent first message on login.

- to convince the user to backup.

### How do I test this?
- login with a user that has not yet secured backup > send a message > wait for 10 second delay > check secure backup dialog is rendered .

#### Things to consider:
  1. Does this change any APIs?

DEMO:

https://github.com/zer0-os/zOS/assets/39112648/9e81bf86-904b-4764-bc9c-cf2587477a11

### DIFF
--- a/src/components/messenger/list/index.tsx
+++ b/src/components/messenger/list/index.tsx
@@ -36,9 +36,12 @@ import { GroupManagementContainer } from './group-management/container';
 import { UserHeader } from './user-header';
 import { getUserSubHandle } from '../../../lib/user';
 import { VerifyIdDialog } from '../../verify-id-dialog';
+import { closeBackupDialog } from '../../../store/matrix';
 
 import { bemClassName } from '../../../lib/bem';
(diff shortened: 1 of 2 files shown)


src/components/messenger/list/index.tsx
```diff
@@ -210,6 +221,14 @@ export class Container extends React.Component<Properties, State> {
     );
   };
 
+  renderSecureBackupDialog = (): JSX.Element => {
```
dalefukami, line 224: Where is the backup dialog currently rendered when you select it from the profile menu? Probably makes sense to remove it from there and just trigger this flag so that it's only rendered in one spot.

That begs the question, is this the right spot?

Anyway, those can be dealt with in follow up PRs but should be considered.

dalefukami (approved):

PR #1534, opened 2024-02-22
Refactor getBackup to use abstracted function rather than client object

dalefukami:
### What does this do?

Refactors some chat client access to simplify the abstraction

### DIFF
--- a/src/store/matrix/saga.ts
+++ b/src/store/matrix/saga.ts
@@ -37,8 +37,7 @@ export function* saga() {
 
 export function* getBackup() {
   yield put(setLoaded(false));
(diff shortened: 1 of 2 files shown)


src/store/matrix/saga.ts
```diff
@@ -37,8 +37,7 @@ export function* saga() {
 
 export function* getBackup() {
   yield put(setLoaded(false));
-  const chatClient = yield call(chat.get);
-  const existingBackup = yield call([chatClient, chatClient.getSecureBackup]);
+  const existingBackup = yield call(getSecureBackup);
```
domw30, line 40: nice 👍

PR #1537, opened 2024-02-23
feat(config): update assets path env variable and config name

domw30:
### What does this do?
- updates the assets path environment variable and config name.

- we are adding a new env variable that contains a path to cloudinary video assets ('video assets path'), as a result, it makes sense to update the existing REACT_APP_ASSETS_PATH and `config.assetsPath` to REACT_APP_IMAGE_ASSETS_PATH and `config.imageAssetsPath`. 

NOTE: I have added the new video path env variable. I shall update the one mentioned above once this is merged.

#### Things to consider:
  1. Does this change any APIs?


dalefukami (approved):
I've added the new env var to the github repo on both development and production.

PR #1540, opened 2024-02-26
Change `unable to decrypt message` text & design

ratik21:
### What does this do?

Before:
image

After:
image

When you click on "help" icon:

https://github.com/zer0-os/zOS/assets/33264364/12a063e8-aff9-479e-94b0-bd32e1ac1e2d

### DIFF
--- a/src/components/message/index.tsx
+++ b/src/components/message/index.tsx
@@ -44,6 +44,8 @@ interface Properties extends MessageModel {
   showSenderAvatar?: boolean;
   showTimestamp: boolean;
   showAuthorName: boolean;
+  isHidden: boolean;
+  onHiddenMessageInfoClick: () => void;
 }
 
 export interface State {
@@ -276,11 +278,17 @@ export class Message extends React.Component<Properties, State> {
   }
 
   renderBody() {
-    const { message } = this.props;
+    const { message, isHidden } = this.props;
 
     return (
       <div {...cn('block-body')}>
-        {message && <ContentHighlighter message={message} />}
+        {message && (
+          <ContentHighlighter
+            message={message}
+            isHidden={isHidden}
+            onHiddenMessageInfoClick={this.props.onHiddenMessageInfoClick}
+          />
+        )}
         {this.renderFooter()}
       </div>
     );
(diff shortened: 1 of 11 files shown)


src/lib/chat/matrix/chat-message.ts
```diff
@@ -58,6 +58,7 @@ export async function mapMatrixMessage(matrixMessage, sdkMatrixClient: SDKMatrix
     },
     parentMessageText: '',
     parentMessageId: parent ? parent['m.in_reply_to']?.event_id : null,
+    isHidden: content?.msgtype === MatrixConstants.BAD_ENCRYPTED_MSGTYPE,
```
dalefukami, line 61: I'm guessing all message related stuff goes through here :crossed_fingers: but I think there's some special logic for message edit events. Did you happen to verify that edit events that we can't decrypt end up with this message? I seem to recall some specialized logic that perhaps we can change to this now. :thinking:

dalefukami (approved):
Just the one question about message edit events. That can be handled in a follow up if there's work to do there.

src/lib/chat/matrix/chat-message.ts
ratik21: yes there's a different logic for capturing the edit events (from the code i am seeing). I had a doubt if we want to display the same message on that event as well. I have asked on the JIRA task, will update in followUp as required. 

Currently there is no change of behaviour with message edit.

src/components/message/index.tsx
```diff
@@ -276,11 +278,17 @@ export class Message extends React.Component<Properties, State> {
...
     return (
       <div {...cn('block-body')}>
-        {message && <ContentHighlighter message={message} />}
+        {message && (
+          <ContentHighlighter
+            message={message}
+            isHidden={isHidden}
+            onHiddenMessageInfoClick={this.props.onHiddenMessageInfoClick}
+          />
+        )}
```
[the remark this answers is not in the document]
ratik21: Makes sense. Yeah will merge this one, and do it in a follow up. Originally..i thought to avoid adding a separate component for this because i thought the code for rendering that message..is too small. design question 😅

PR #1548, opened 2024-02-27
feat(generate-backup-component): add test coverage

domw30:
### What does this do?
- adds tests for the `GenerateBackup` component.

- test coverage.

### How do I test this?
- run tests - `src/components/secure-backup/generate-backup/index.test.tsx`

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/secure-backup/generate-backup/index.test.tsx
+++ b/src/components/secure-backup/generate-backup/index.test.tsx
@@ -1,6 +1,6 @@
 import { shallow } from 'enzyme';
 
-import { buttonLabelled } from '../../../test/utils';
+import { buttonLabelled, pressButton } from '../../../test/utils';
 
 import { GenerateBackup, Properties } from '.';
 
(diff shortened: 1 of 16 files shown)


src/components/secure-backup/generate-backup/index.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+    expect(wrapper.find(c('recovery-key'))).toHaveText('abcd 1234');
+  });
+
+  it('copies the recovery key to the clipboard', function () {
+    const clipboard = { write: jest.fn() };
+    const wrapper = subject({ clipboard, recoveryKey: 'abcd 1234' });
+
+    pressButton(wrapper, 'Copy');
+
+    expect(clipboard.write).toHaveBeenCalledWith(expect.stringContaining('abcd 1234'));
```
dalefukami, line 35: This should be an exact match probably?

PR #1559, opened 2024-02-27
feat(secure-backup): add border error state to verify backup key inputs

domw30:
### What does this do?
- adds error border state as below.

- as per figma design.

Screenshot 2024-02-27 at 11 54 06

### DIFF
--- a/src/components/secure-backup/restore-backup/index.tsx
+++ b/src/components/secure-backup/restore-backup/index.tsx
@@ -44,7 +44,12 @@ export class RestoreBackup extends React.Component<Properties, State> {
           </p>
 
           <div {...cn('input-container')}>
-            <Input placeholder='Enter your recovery key' onChange={this.trackRecoveryKey} value={this.recoveryKey} />
+            <Input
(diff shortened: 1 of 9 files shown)


src/components/secure-backup/restore-backup/index.tsx
```diff
@@ -44,7 +44,12 @@ export class RestoreBackup extends React.Component<Properties, State> {
           </p>
 
           <div {...cn('input-container')}>
-            <Input placeholder='Enter your recovery key' onChange={this.trackRecoveryKey} value={this.recoveryKey} />
+            <Input
+              {...cn('input', `${this.props.errorMessage && 'error'}`)}
```
dalefukami, line 48: Is there not already a way to tell the input that it's in an error state?

domw30: At some point during redesign, the border was removed from `Input` in `zUI`. There is an error state for the component that sets the border color, but I'll need to go into `zUI` and add `border: 1px solid` to that error state. This was more of a quick fix for now.

PR #1572, opened 2024-02-28
refactor: rename restore-prompt props - onVerifyKey to onNext

domw30:
### What does this do?
- renaming props.

- 'onNext' is a better reflection of what happens 'next'. The user will be navigated to the verify stage.


dalefukami (approved):
:heart:

PR #1587, opened 2024-02-28
Remove unused clipboard reference

dalefukami:
### What does this do?

Removes old remnants


domw30 (commented):
thanks for removing this one 👍

PR #1588, opened 2024-02-29
refactor(conversation-header): replace stub user with imported test store stub user

domw30:
### What does this do?
- replaces stub user in conversation header test file with imported test store stub user


src/store/test/store.ts
```diff
@@ -151,6 +151,20 @@ function stubUser(attrs: Partial<User> = {}): Partial<User> {
...
+    isOnline: false,
+    primaryWallet: null,
+    wallets: [],
+    ...attrs,
+  };
+}
+
+export function stubConversation(attrs: Partial<Channel> = {}): Partial<Channel> {
+  return {
+    id: 'channel-id',
```
dalefukami, line 166: Let's update this to use the `stubCount` to ensure unique ids between stubs. Otherwise, some tests actually become invalid.

dalefukami (approved):

PR #1593, opened 2024-03-01
Reset the login stage during logout

dalefukami:
### What does this do?

Sets the LoginStage to the default Web3 when logging out.

There was a bug where if you start on the login page and login via email...then logout... instead of redirecting to the login page you'd remain at the root url. This happened because the completion of email logout sets the stage to Login.Done. So, the login page itself thought it was done and would redirect to the root url but the root url would detect is was not authenticated and thus not render anything.


dalefukami:
Note: this is a quick fix. I think the real fix would be to manage the login stage via the sagas rather than have the Login component doing the redirect.

PR #1599, opened 2024-03-04
refactor: backup faq 

domw30:
- some proposed changes to backup faq implementation


src/components/secure-backup/index.tsx
```diff
@@ -70,7 +69,7 @@ export class SecureBackup extends React.PureComponent<Properties, State> {
   };
 
   renderVideoBanner = () => {
-    if (this.state.faqModalOpen) {
+    if (this.state.isFAQContent) {
```
ratik21, line 72: I think `showFAQContent` makes more sense here.

PR #1610, opened 2024-03-13
feat: update invite clipboard text

domw30:
### What does this do?
- updates the clipboard text when copying invite code

- as per requirements/design
- this gives more context/info to the person who receives the invite code that a current user has copied and sent.

### How do I test this?
- run test as usual.
- checkout branch > run UI > copy invite code > paste and check string has the same structure as below screenshot.

#### Things to consider:
  1. Does this change any APIs?

Screenshot 2024-03-13 at 17 58 14

### DIFF
--- a/src/components/invite-dialog/index.tsx
+++ b/src/components/invite-dialog/index.tsx
@@ -45,7 +45,7 @@ export class InviteDialog extends React.Component<Properties, State> {
   }
 
   get inviteText() {
-    return this.props.inviteCode;
+    return `Use this code to join me on ZERO Messenger: ${this.props.inviteCode} https://zos.zer0.io/get-access`;
   }
 
   writeInviteToClipboard = async () => {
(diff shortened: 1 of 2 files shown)


src/components/invite-dialog/index.tsx
```diff
@@ -45,7 +45,7 @@ export class InviteDialog extends React.Component<Properties, State> {
   }
 
   get inviteText() {
-    return this.props.inviteCode;
+    return `Use this code to join me on ZERO Messenger: ${this.props.inviteCode} https://zos.zer0.io/get-access`;
```
dalefukami, line 48: We definitely shouldn't hard code the url. Do we not have env vars for the root url?

domw30: We'll have to add a new one for the root url

dalefukami: Seems like we have REACT_APP_MESSENGER_INVITE_PATH

domw30: Sure, I'll update.

domw30: Updated to use REACT_APP_MESSENGER_INVITE_PATH & removed the env var I added to github variables. Changes merged :)

PR #1639, opened 2024-03-20
Display rewards tooltip if user has accumulated new rewards (previous day) 

ratik21:
### What does this do?

Displays a rewards tooltip, which shows the user that there are accumulated rewards since the **last day** which (s)he has not seen. 

**NOTE:** As per design, the tooltip disappears after 3s, and we record that the user has seen the rewards (i.e if the user reloads, then the tooltip is NOT displayed again)

image

### DIFF
--- a/src/components/messenger/list/rewards-tooltip/container.tsx
+++ b/src/components/messenger/list/rewards-tooltip/container.tsx
@@ -2,6 +2,8 @@ import * as React from 'react';
 
 import { connectContainer } from '../../../../store/redux-container';
 import { RootState } from '../../../../store/reducer';
+import { rewardsTooltipClosed } from '../../../../store/rewards';
+
 import { RewardsTooltip } from '.';
 import { Container as RewardsItemContainer } from '../../../settings-menu/rewards-item/container';
 export interface PublicProperties {}
@@ -9,6 +11,8 @@ export interface PublicProperties {}
 export interface Properties extends PublicProperties {
   meowPreviousDayInUSD: string;
   isLoading: boolean;
+
+  rewardsTooltipClosed: () => void;
 }
 
 export class Container extends React.Component<Properties> {
@@ -21,7 +25,7 @@ export class Container extends React.Component<Properties> {
   }
 
   static mapActions(_props: Properties): Partial<Properties> {
-    return {};
+    return { rewardsTooltipClosed };
   }
 
   render() {
@@ -29,7 +33,12 @@ export class Container extends React.Component<Properties> {
       return null;
     }
 
-    return <RewardsTooltip meowPreviousDayInUSD={this.props.meowPreviousDayInUSD} children={this.props.children} />;
+    return (
+      <RewardsTooltip
+        meowPreviousDayInUSD={this.props.meowPreviousDayInUSD}
+        onClose={this.props.rewardsTooltipClosed}
+      />
+    );
   }
 }
 export const RewardsToolTipContainer = connectContainer<PublicProperties>(Container);
--- a/src/components/messenger/list/rewards-tooltip/container.tsx
+++ b/src/components/messenger/list/rewards-tooltip/container.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 
 import { connectContainer } from '../../../../store/redux-container';
 import { RootState } from '../../../../store/reducer';
-import { rewardsTooltipClosed } from '../../../../store/rewards';
+import { closeRewardsTooltip } from '../../../../store/rewards';
 
 import { RewardsTooltip } from '.';
 import { formatUSD, calculateTotalPriceInUSDCents } from '../../../../lib/number';
@@ -14,7 +14,7 @@ export interface Properties extends PublicProperties {
   isLoading: boolean;
   meowTokenPriceInUSD: number;
 
-  rewardsTooltipClosed: () => void;
+  closeRewardsTooltip: () => void;
 }
 
 export class Container extends React.Component<Properties> {
@@ -30,7 +30,7 @@ export class Container extends React.Component<Properties> {
   }
 
   static mapActions(_props: Properties): Partial<Properties> {
-    return { rewardsTooltipClosed };
+    return { closeRewardsTooltip };
   }
 
   static totalPriceInUSD(meow: string, meowInUSD: number) {
@@ -45,7 +45,7 @@ export class Container extends React.Component<Properties> {
     return (
       <RewardsTooltip
         meowPreviousDayInUSD={this.props.meowPreviousDayInUSD}
-        onClose={this.props.rewardsTooltipClosed}
+        onClose={this.props.closeRewardsTooltip}
         open={this.props.isOpen}
       />
     );
--- a/src/components/messenger/list/rewards-tooltip/container.tsx
+++ b/src/components/messenger/list/rewards-tooltip/container.tsx
@@ -5,12 +5,14 @@ import { RootState } from '../../../../store/reducer';
 import { rewardsTooltipClosed } from '../../../../store/rewards';
 
 import { RewardsTooltip } from '.';
-import { Container as RewardsItemContainer } from '../../../settings-menu/rewards-item/container';
+import { formatUSD, calculateTotalPriceInUSDCents } from '../../../../lib/number';
 export interface PublicProperties {}
 
 export interface Properties extends PublicProperties {
+  isOpen: boolean;
   meowPreviousDayInUSD: string;
   isLoading: boolean;
+  meowTokenPriceInUSD: number;
 
   rewardsTooltipClosed: () => void;
 }
@@ -18,9 +20,12 @@ export interface Properties extends PublicProperties {
 export class Container extends React.Component<Properties> {
   static mapState(state: RootState): Partial<Properties> {
     const { rewards } = state;
+
     return {
-      meowPreviousDayInUSD: RewardsItemContainer.totalPriceInUSD(rewards.meowPreviousDay, rewards.meowInUSD),
+      meowPreviousDayInUSD: Container.totalPriceInUSD(rewards.meowPreviousDay, rewards.meowInUSD),
       isLoading: rewards.loading,
+      meowTokenPriceInUSD: rewards.meowInUSD,
+      isOpen: rewards.showRewardsInTooltip,
     };
   }
 
@@ -28,8 +33,12 @@ export class Container extends React.Component<Properties> {
     return { rewardsTooltipClosed };
   }
 
+  static totalPriceInUSD(meow: string, meowInUSD: number) {
+    return formatUSD(calculateTotalPriceInUSDCents(meow, meowInUSD ?? 0));
+  }
+
   render() {
-    if (this.props.isLoading || this.props.meowPreviousDayInUSD === '$0.00') {
+    if (this.props.isLoading || this.props.meowTokenPriceInUSD === 0) {
       return null;
     }
 
@@ -37,6 +46,7 @@ export class Container extends React.Component<Properties> {
       <RewardsTooltip
         meowPreviousDayInUSD={this.props.meowPreviousDayInUSD}
         onClose={this.props.rewardsTooltipClosed}
+        open={this.props.isOpen}
       />
     );
   }
--- a/src/components/messenger/list/rewards-tooltip/container.tsx
+++ b/src/components/messenger/list/rewards-tooltip/container.tsx
@@ -25,7 +25,7 @@ export class Container extends React.Component<Properties> {
   }
 
   render() {
-    if (this.props.isLoading) {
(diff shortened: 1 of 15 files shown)


src/components/tooltip-popup/styles.scss
```diff
@@ -13,41 +15,49 @@
...
-    line-height: 24px;
+    font-weight: 400;
+    line-height: 20px;
+    font-size: 14px;
 
-    /* Dark/Greyscale/Grey-12 */
-    color: theme.$color-greyscale-12;
+    @include glass-shadow-and-blur;
+    @include flat-thick;
+    backdrop-filter: blur(32px);
```
domw30, line 32: Do we need this backdrop-filter if its set in `@include glass-shadow-and-blur;` ? Looks like duplication?

src/components/messenger/list/rewards-tooltip/container.tsx
```diff
@@ -0,0 +1,44 @@
...
+      isLoading: rewards.loading,
+    };
+  }
+
+  static mapActions(_props: Properties): Partial<Properties> {
+    return { rewardsTooltipClosed };
+  }
+
+  render() {
+    if (this.props.isLoading || this.props.meowPreviousDayInUSD === '$0.00') {
```
domw30, line 32: Should we use `showRewardsTooltip` here instead of hard coding 0 value? Is the saga handling `showRewardsTooltip` in that way? I haven't checked but would seem thats what the boolean is handling?

src/components/messenger/list/rewards-tooltip/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+  };
+
+  componentDidMount(): void {
+    setTimeout(() => {
+      this.closeTooltip();
+    }, 3000);
+  }
+
+  closeTooltip = () => {
+    this.setState({ isTooltipOpen: false });
```
domw30, line 27: Again not sure if we could use `showRewardsTooltip` here instead of internal state within the component?

src/components/tooltip-popup/styles.scss
```diff
@@ -13,41 +15,49 @@
...
     }
+
     &[data-side='left'][data-align='start'] {
       border-top-right-radius: unset;
     }
   }
 
   &__arrow {
-    fill: theme.$color-primary-3;
+    position: absolute;
```
domw30, line 53: I've not run zos on this branch but something seems off if we're using `position: absolute` and minus position values here for the arrow. If the tooltip needs to be in a different position if used elsewhere will the arrow look out of place?

src/components/tooltip-popup/styles.scss
```diff
@@ -1,10 +1,12 @@
 @use '~@zero-tech/zui/styles/theme' as theme;
 @use '~@zero-tech/zui/styles/colors' as colors;
+@import '../../glass';
 
 .tooltip-popup {
-  display: block;
+  position: fixed;
```
domw30, line 6: Does this break the side and align settings? `data-side='right'][data-align='end'`

domw30 (commented):
I've not got much context here so asked a few questions for some things that stood out to me 👍

src/components/tooltip-popup/styles.scss
ratik21: actually `flat-thick` has a backdrop-filter of `64px` set, i added this here to override..

src/components/messenger/list/rewards-tooltip/index.tsx
ratik21: both are different. `showRewardsTooltip` is global state which represents "has the users seen the last day rewards or not". `isTooltipOpen` is local state which controls the open/close of tooltip rendered (in this case only `close`, since opening is handled automatically).

src/components/messenger/list/rewards-tooltip/container.tsx
ratik21: similar to below comment .., i didn't want to pass this to the actual component so did it here. not sure about moving this to the saga since we have an initial fetch/sync triggered for the rewards when we login..

src/components/tooltip-popup/styles.scss
ratik21: yeah could be, i was just trying to fix the css of the current tooltip. Right now, this is the only place it is used (and the tooltip-popup code was copied directly from zUI), so, .. 🤷‍♂️

src/components/tooltip-popup/styles.scss
ratik21: currently it doesn't, but i also haven't tested a lot of cases. just for the one we need to display. (arrow pointed left, data pointed right)

src/components/messenger/list/rewards-tooltip/container.tsx
dalefukami, line 32: Probably shouldn't do a comparison based on a formatted string. If we need to know the actual numeric value then we should send it in here as a prop.

src/components/messenger/list/rewards-tooltip/container.tsx
```diff
@@ -0,0 +1,44 @@
...
+  isLoading: boolean;
+
+  rewardsTooltipClosed: () => void;
+}
+
+export class Container extends React.Component<Properties> {
+  static mapState(state: RootState): Partial<Properties> {
+    const { rewards } = state;
+    return {
+      meowPreviousDayInUSD: RewardsItemContainer.totalPriceInUSD(rewards.meowPreviousDay, rewards.meowInUSD),
```
dalefukami, line 22: Don't reach into another component. It may work but those aren't meant to be shared.

src/components/messenger/list/rewards-tooltip/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export class RewardsTooltip extends React.Component<Properties, State> {
+  state = {
+    isTooltipOpen: true,
+  };
+
+  componentDidMount(): void {
+    setTimeout(() => {
+      this.closeTooltip();
+    }, 3000);
```
dalefukami, line 23: This is dangerous. What if the component is already destroyed when the timer runs out?

Maybe we should just have this in the saga. It seems a little weird that we have multiple ways of determining whether the tooltip should be visible or not.

src/components/messenger/list/rewards-tooltip/container.tsx
ratik21: ok

src/components/messenger/list/rewards-tooltip/container.tsx
ratik21: The `meowPreviousDayInUSD` here is getting `$0.00` set here 
image

This is because it's taking a while to set the "current meow token price". Until then it is set in the state as `0`. Which is because the rewards are coming out to be `0.00`. 

I'll update this check to be against the `meowTokenPrice`.

src/components/messenger/list/rewards-tooltip/index.tsx
ratik21: okay, removed `isTooltipOpen`, and directly used `rewards.showRewardsInTooltip` prop. We're calling the saga method here, also tested that even if the component is destroyed (i.e user clicks on the `x` icon before the tooltip auto-hides), then also this code will be triggered. In the saga itself we're checking first if the rewards tooltip is closed or not.

ratik21: > Maybe we should just have this in the saga.

Had a doubt about this one, since if we trigger auto close from the "saga", i'm not sure if there would be a time lag.. (since the moment we set the `showRewards` in saga to `true`, and the actual moment the component, i.e, the rewards tooltip which renders might be different., right?)

So, i called `this.props.onClose` from the frontend (after the component is mounted)

src/components/messenger/list/rewards-tooltip/container.tsx
```diff
@@ -0,0 +1,54 @@
+import * as React from 'react';
+
+import { connectContainer } from '../../../../store/redux-container';
+import { RootState } from '../../../../store/reducer';
+import { rewardsTooltipClosed } from '../../../../store/rewards';
```
dalefukami, line 5: I think we should rename this to `closeRewardsTooltip`. It makes more sense in the most contexts.

src/components/messenger/list/rewards-tooltip/index.tsx
ratik21: > When we call put the state is updated and things are rerendered right away

Thanks, didn't knew about this. Updated. Added tests as well.

src/components/messenger/list/rewards-tooltip/container.tsx
```diff
@@ -0,0 +1,54 @@
...
+  static mapActions(_props: Properties): Partial<Properties> {
+    return { rewardsTooltipClosed };
+  }
+
+  static totalPriceInUSD(meow: string, meowInUSD: number) {
+    return formatUSD(calculateTotalPriceInUSDCents(meow, meowInUSD ?? 0));
+  }
+
+  render() {
+    if (this.props.isLoading || this.props.meowTokenPriceInUSD === 0) {
```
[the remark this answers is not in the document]
ratik21: Moved to saga. Essentially added a `yield call(..)` to fetch the current token price first (if it is 0) before checking for new rewards.

src/store/rewards/saga.ts
```diff
@@ -38,13 +38,13 @@ export function* fetch(_action) {
...
       yield put(setMeowPreviousDay(result.response.meowPreviousDay.toString()));
-
-      yield call(checkNewRewardsLoaded);
     } else {
     }
   } catch (e) {
   } finally {
     yield put(setLoading(false));
+
+    yield call(checkNewRewardsLoaded);
```
dalefukami, line 47: Why did we move this into the `finally`? If the `fetchRewards` call fails what good does it do us to then see if your rewards changed?

src/store/rewards/saga.ts
```diff
@@ -68,19 +68,30 @@ export function* syncRewardsAndTokenPrice() {
   if (!featureFlags.enableRewards) {
     return;
   }
-
-  yield spawn(syncFetchRewards);
   yield spawn(syncMEOWPrice);
+  yield spawn(syncFetchRewards);
```
dalefukami, line 72: Just as an FYI, since these are both spawned to run "in the background", it doesn't matter which order they're in.

src/store/rewards/saga.ts
```diff
@@ -68,19 +68,30 @@ export function* syncRewardsAndTokenPrice() {
...
 
 export function* checkNewRewardsLoaded() {
   const { meowPreviousDay, meow } = yield select((state) => state.rewards);
   const isFirstTimeLogin = yield select((state) => getDeepProperty(state, 'registration.isFirstTimeLogin'));
+  const meowTokenPrice = yield select((state) => state.rewards.meowInUSD);
+  if (meowTokenPrice === 0) {
+    yield call(fetchCurrentMeowPriceInUSD);
+  }
 
   if (!isFirstTimeLogin && meowPreviousDay !== '0') {
```
dalefukami, line 88: Can we move the `isFirstTimeLogin` check higher up so we don't actually have to do any of the other stuff if it's your first time logging in?

dalefukami (approved):
Looks pretty close. Handle the comments here or in a follow up since it's feature flagged it's safe.

src/store/rewards/saga.ts
ratik21: got it. I actually wanted to do something like
```
yield call(syncMEOWPrice);
yield spawn(syncFetchRewards);
```

So that when we're doing `syncFetchRewards`, we have the actual `MEOW` token price in state. (so that when we're calculating the total rewards in the UI, it's not 0 for the first time).

But .. it didn't work, so i added the check in `checkNewRewardsLoaded` (to fetch the token price first before showing the rewards)

src/store/rewards/saga.ts
ratik21: yes

src/store/rewards/saga.ts
ratik21: wanted to call `checkNewRewardsLoaded` after Loading has been set to `false`. (so we're able to display the rewards tooltip). Otherwise in the `checkNewRewardsLoaded` we're setting `showRewardsTooltip` to `true`, but we're still in "loading state".

src/store/rewards/saga.ts
```diff
@@ -68,19 +68,33 @@ export function* syncRewardsAndTokenPrice() {
...
   }
-
-  yield spawn(syncFetchRewards);
   yield spawn(syncMEOWPrice);
+  yield spawn(syncFetchRewards);
+}
+
+export function* closeRewardsTooltipAfterDelay() {
+  yield delay(3000);
+  //yield call(closeRewardsTooltip);
```
dalefukami, line 77: Mistake?

ratik21: 😭

PR #1641, opened 2024-03-22
feat: fallback content construction for message replies and handle parsing of plain body

domw30:
### What does this do?
- implements the construction of fallback content for message replies, automatically generating a quoted format of the parent message that precedes the reply content.
- adds util to parse the content body body to filter out fallback content. 

- to align with Mobile (android/ios) content body expectations.
- ensures that the displayed messages contain only the original content added by the user.

### How do I test this?
- run tests > send replies between android and web devices.

#### Things to consider:
  1. Does this change any APIs?
  
  Constructing the fallback
Screenshot 2024-03-25 at 14 11 09

Before:

Screenshot 2024-03-25 at 14 40 56
  

After:
Screenshot 2024-03-25 at 14 39 07

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -455,7 +455,8 @@ export class MatrixClient implements IChatClient {
           event_id: parentMessage.messageId,
         },
       };
-      (content['format'] = 'org.matrix.custom.html'), (content['formatted_body'] = message);
+      content['format'] = 'org.matrix.custom.html';
+      content['formatted_body'] = message;
     }
 
     const messageResult = await this.matrix.sendMessage(channelId, content);
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -455,6 +455,7 @@ export class MatrixClient implements IChatClient {
           event_id: parentMessage.messageId,
         },
       };
+      (content['format'] = 'org.matrix.custom.html'), (content['formatted_body'] = message);
     }
 
     const messageResult = await this.matrix.sendMessage(channelId, content);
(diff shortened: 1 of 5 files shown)


src/lib/chat/matrix-client.ts
```diff
@@ -455,6 +455,8 @@ export class MatrixClient implements IChatClient {
           event_id: parentMessage.messageId,
         },
       };
+      content['format'] = 'org.matrix.custom.html';
+      content['formatted_body'] = message;
```
dalefukami, line 459: I don't think we should be doing this. It's not required and I don't think it's a good idea for the sending client to be telling the receiving client how to render things. That's just coupling things together for no reason.

dalefukami (commented):
I'm assuming we'd also need the rendering piece to this or we'll be rendering all the meta stuff on any replies we create?

src/lib/chat/matrix/chat-message.ts
```diff
@@ -35,9 +36,11 @@ export async function mapMatrixMessage(matrixMessage, sdkMatrixClient: SDKMatrix
   const parent = matrixMessage.content['m.relates_to'];
   const senderData = sdkMatrixClient.getUser(senderId);
 
+  const parsedContentBody = parsePlainBody(content.body);
```
domw30, line 39: @dalefukami - does this look right to you? While debugging the failing test I noticed that if I log `content.body` it sometimes logs the SecureBackup file code!

Screenshot 2024-03-25 at 17 36 33

domw30: Fixed this, but, still very strange that the secure backup file gets logged if you console log `content.body` in `mapMatrixMessage`

PR #1642, opened 2024-03-22
feat: add util to parse content body, filtering out fallback content

domw30:
### What does this do?
- adds util to parse the content body body to filter out fallback content. 

- ensures that the displayed messages contain only the original content added by the user.

### How do I test this?
- run tests as usual > send messages from mobile device to web

#### Things to consider:
  1. Does this change any APIs?


src/lib/chat/matrix/utils.ts
```diff
@@ -58,3 +59,10 @@ export async function getFilteredMembersForAutoComplete(roomMembers: ChannelMemb
 
   return filteredResults;
 }
+
+export function parseFormattedReply(formattedBody) {
```
dalefukami, line 63: We should be parsing the plain body.

PR #1648, opened 2024-03-25
Add unread counts to conversation tabs

dalefukami:
### What does this do?

Adds the unread counts to the All and Favorite tabs

image


domw30 (commented):
Nice work getting through the bulk of the UI 👍

PR #1649, opened 2024-03-26
refactor: remove max invites per user

domw30:
### What does this do?
- removes `maxInvitesPerUser` / `maxUses`.

- no longer require a max value for invites.

### How do I test this?
- run tests as usual.

#### Things to consider:
  1. Does this change any APIs?

NOTE:: I notice the designs still contain the remaining count of invites displayed on the invite dialog. Are we sure we no longer require an invite count?

Screenshot 2024-03-26 at 08 31 55

After deletions:
Screenshot 2024-03-26 at 08 33 04

### DIFF
--- a/src/components/invite-dialog/index.tsx
+++ b/src/components/invite-dialog/index.tsx
@@ -40,6 +40,10 @@ export class InviteDialog extends React.Component<Properties, State> {
     clearTimeout(this.buttonTimeout);
   }
 
+  getInvitesRemaining() {
+    return 5;
+  }
+
   get inviteText() {
     return `Use this code to join me on ZERO Messenger: ${this.props.inviteCode} ${config.inviteUrl}`;
   }
@@ -70,6 +74,15 @@ export class InviteDialog extends React.Component<Properties, State> {
           <Button onPress={this.writeInviteToClipboard} isDisabled={!this.props.inviteCode}>
             {this.state.copyText}
           </Button>
+
+          <div {...cn('remaining-invite-container')}>
+            {!this.props.isLoading && (
+              <>
+                <div {...cn('remaining-invite')}>{this.getInvitesRemaining()}</div>
+                <>Remaining</>
+              </>
+            )}
+          </div>
         </div>
       </div>
     );
--- a/src/components/invite-dialog/index.tsx
+++ b/src/components/invite-dialog/index.tsx
@@ -17,6 +17,7 @@ export interface Clipboard {
 
 export interface Properties {
   inviteCode: string;
+  inviteCount: string;
   invitesUsed: number;
   inviteUrl: string;
   assetsPath: string;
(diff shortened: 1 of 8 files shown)


src/components/invite-dialog/index.tsx
```diff
@@ -75,15 +70,6 @@ export class InviteDialog extends React.Component<Properties, State> {
           <Button onPress={this.writeInviteToClipboard} isDisabled={!this.props.inviteCode}>
             {this.state.copyText}
           </Button>
-
-          <div {...cn('remaining-invite-container')}>
```
dalefukami, line 79: I don't think we want to remove this, do we? We still want to show this it just comes from a different number?

domw30: Yeah mentioned above - `NOTE:: I notice the designs still contain the remaining count of invites displayed on the invite dialog. Are we sure we no longer require an invite count?`

If we're not going to return an `inviteCount` in the response (on the frontend referenced as `maxUses`) when fetching the invite, where is the number going to come from I'm not sure I understand this part. Unless the idea here was to replace `maxUses` with something like `inviteCount` or `remainingCount` or something..

dalefukami: I'm confused, why wouldn't we return an inviteCount from the api? We have the count now, right? We just don't need invitesUsed and maxInvitesPerUser

domw30: lol yeah sorry, some misunderstanding from yesterday 😆 thought it wasn't quite right removing the count in UI. 

Yeah so I'll re-add that part of the invite dialog and get the count from the `inviteCount` prop instead 🙂

domw30: Made the changes and added inviteCount. There is a follow up PR that removes the `invitesUsed` prop.

src/store/create-invitation/saga.test.ts
```diff
@@ -39,12 +38,12 @@ describe('fetchInvite', () => {
         storeState: { createInvitation },
       } = await expectSaga(fetchInvite)
         .withReducer(rootReducer, {
-          createInvitation: { code: 'something', url: 'url', invitesUsed: 2, maxUses: 3, isLoading: false },
+          createInvitation: { code: 'something', url: 'url', invitesUsed: 2, inviteCount: '3', isLoading: false },
```
dalefukami, line 41: is inviteCount from the server actually a string? Should be an integer, I think.

domw30: Double checked and its number. Updated.

dalefukami:
@domw30 I noticed you merged this already. Is this backwards compatible with the current state of the api? I.E., if we deploy this today it'll still work?

domw30:
> @domw30 I noticed you merged this already. Is this backwards compatible with the current state of the api? I.E., if we deploy this today it'll still work?

I will need to deploy the api first. Was planning on doing both after stand up. In hindsight I should have waited to merge this and the follow up PR prior to api being deployed and up to date

PR #1651, opened 2024-03-26
feat: add conversation list item 'more' menu component

domw30:
### What does this do?
- adds the 'more' menu.

-  to open dropdown menu on conversation list items and call actions such as tagging a conversation as a favorite.

### How do I test this?
- run tests as usual.

#### Things to consider:
  1. Does this change any APIs?
  
  
 `isFavorite` = true  -> Unfavorite:

https://github.com/zer0-os/zOS/assets/39112648/35da6119-47c2-4a12-96ee-d32b40cd6219

 `isFavorite` = false  -> Favorite:
  
  

https://github.com/zer0-os/zOS/assets/39112648/20122cab-bf9c-4424-a7a9-1b45847166fa


domw30:
I really don't like `MoreMenu` 🤒  couldn't decide on what we should name this (as it looks like it is actually referenced as the 'more' menu.

dalefukami:
> I really don't like `MoreMenu` 🤒 couldn't decide on what we should name this (as it looks like it is actually referenced as and named the 'more' menu)

Yeah, it definitely shouldn't be MoreMenu...especially if it's only nested under the MessengerList because many different things could have a More menu.

domw30:
> Maybe nest it under conversation-item and then it makes better sense as MoreMenu?

Yeah sounds good to me, shall move :)

PR #1661, opened 2024-03-27
feat: implement removing favorite room tags

domw30:
### What does this do?
- removes room from favorites (matrix)
- handles room unfavorited updates (saga)
- adds and handles action for unfavoriting room
- updates the publishing of room tag updates

- to be able to mark rooms (conversations) NOT as favorites (unfavoriting) for an enhanced user experience

### How do I test this?
- run tests as usual.
- alternatively, you could call the `removeRoomFromFavorites` and check the tag event is logged and the event content tags no longer contain the `m.favorite` tag.

#### Things to consider:
  1. Does this change any APIs?

REMOVE Favorite Demo (once wired up) with rough three dot menu

https://github.com/zer0-os/zOS/assets/39112648/98819d9c-f182-4ca1-aa96-852e45d874c9

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -557,8 +557,6 @@ export class MatrixClient implements IChatClient {
     for (const user of users) {
       await this.matrix.invite(roomId, user.matrixId);
     }
-
-    this.addFavoriteRoomTag(roomId);
   }
 
   async editRoomNameAndIcon(roomId: string, name: string, iconUrl: string): Promise<void> {
@@ -655,7 +653,7 @@ export class MatrixClient implements IChatClient {
     return await Promise.all(matches.map((r) => this.mapConversation(r)));
   }
 
-  async addFavoriteRoomTag(roomId: string): Promise<void> {
+  async addRoomToFavorites(roomId: string): Promise<void> {
     await this.waitForConnection();
 
     try {
@@ -750,6 +748,10 @@ export class MatrixClient implements IChatClient {
         this.receiveDeleteMessage(event);
       }
 
+      if (event.type === EventType.Tag) {
+        this.publishRoomTagUpdate(event);
+      }
+
       this.processMessageEvent(event);
     });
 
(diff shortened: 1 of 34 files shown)


src/lib/chat/matrix-client.ts
```diff
@@ -936,7 +942,7 @@ export class MatrixClient implements IChatClient {
     if (isFavoriteTagAdded) {
       this.events.roomFavorited(roomId);
     } else {
-      console.log('room unfavorited');
+      this.events.roomUnfavorited(roomId);
```
dalefukami, line 945: I'm curious about this TAG event. Is this going to get fired whenever _any_ tag gets changed? Is there a way to know which tag actually changed rather than firing off these Favorite events even if an unrelated tag changes?

src/store/channels/saga.ts
```diff
@@ -110,6 +110,18 @@ export function* onFavoriteRoom(action) {
...
 }
 
+export function* onUnfavoriteRoom(action) {
+  const { roomId } = action.payload;
+
+  const channel = yield select(rawChannelSelector(roomId));
+
+  if (!channel) {
+    return;
+  }
```
dalefukami, line 120: Unnecessary check?

src/store/channels/saga.ts
```diff
@@ -110,6 +110,18 @@ export function* onFavoriteRoom(action) {
...
+
+  yield call(removeRoomFromFavorites, roomId);
```
dalefukami, line 122: We should probably wrap these in try/catch statements. What if Matrix fails to execute this...what should we tell the user?

dalefukami (approved):

src/store/channels/saga.ts
domw30: Removed the unnecessary checks

src/store/channels/saga.ts
domw30: Follow up was going to include some form of basic error state for either of these saga functions failing. For this PR I have now added try/catch statements, simply logging error. Shall implement something for the user as a follow up

src/lib/chat/matrix-client.ts
domw30: Yeah as discussed previously, our current interaction with tags is limited too managing favorites, correct in pointing out this firing whenever there is a tag change - the tag structure doesn't inherently differentiate between the addition or removal of specific tags—it simply presents the current state. This approach would indeed require reevaluation and fine-tuning should our application's tag-related features expand beyond favorites

PR #1666, opened 2024-03-28
refactor(conversation-item/more-menu): handle unfavorite event

domw30:
### What does this do?
- adds unfavorite call/ to `MoreMenu` `unfavorite` menu item

- to fire the `onUnfavorite` event.

### How do I test this?
- run tests as usual.

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/messenger/list/conversation-item/more-menu/index.tsx
+++ b/src/components/messenger/list/conversation-item/more-menu/index.tsx
@@ -7,12 +7,18 @@ import './styles.scss';
 
 export interface Properties {
   isFavorite: boolean;
+
   onFavorite: () => void;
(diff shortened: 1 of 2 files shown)


src/components/messenger/list/conversation-item/more-menu/index.tsx
```diff
@@ -24,16 +30,24 @@ export class MoreMenu extends React.Component<Properties> {
   }
 
   get menuItems() {
-    return [
-      {
+    const menuItems = [];
```
dalefukami, line 33: Yeah, I like this change.

dalefukami (approved):

PR #1668, opened 2024-03-28
feat: add and display `more` menu to conversation list items

domw30:
### What does this do?
- adds and displays the `MoreMenu` to conversation list items.

- to display the favorite/unfavorite dropdown options so the user can control (set/unset) favorite tags on rooms.

### How do I test this?
- run locally and hover over conversation list items.

#### Things to consider:
  1. Does this change any APIs?

Demo:

https://github.com/zer0-os/zOS/assets/39112648/64fa68c2-7fe7-4113-8305-b42e11f26b39

### DIFF
--- a/src/components/messenger/list/conversation-item/index.test.tsx
+++ b/src/components/messenger/list/conversation-item/index.test.tsx
@@ -4,6 +4,7 @@ import { shallow } from 'enzyme';
 import { ConversationItem, Properties } from '.';
 import moment from 'moment';
 import { ContentHighlighter } from '../../../content-highlighter';
+import { MoreMenu } from './more-menu';
 
 describe('ConversationItem', () => {
   const subject = (props: Partial<Properties>) => {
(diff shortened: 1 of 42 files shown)


src/components/messenger/list/conversation-item/index.tsx
```diff
@@ -85,6 +106,21 @@ export class ConversationItem extends React.Component<Properties> {
       />
     );
   }
+  renderMoreMenu() {
+    const stopPropagation = (e) => {
```
domw30, line 110: used to prevent firing 'openConversation' click events.

src/components/messenger/list/conversation-item/index.test.tsx
```diff
@@ -137,6 +138,32 @@ describe('ConversationItem', () => {
     expect(wrapper.find('.conversation-item__timestamp').text()).toEqual(previewDisplayDate);
   });
 
+  it('renders MoreMenu when mouse enters', () => {
```
domw30, line 141: Noted down that this test file is in need of some refactoring as a follow up.

src/components/messenger/list/conversation-item/index.tsx
```diff
@@ -111,8 +147,14 @@ export class ConversationItem extends React.Component<Properties> {
           tabIndex={0}
           role='button'
           is-active={isActive}
+          onMouseEnter={this.toggleHover(true)}
+          onMouseLeave={this.toggleHover(false)}
```
dalefukami, line 151: Is there a reason we can't use CSS to detect the hover? This is a little dicey

dalefukami (approved):

src/components/messenger/list/conversation-item/index.tsx
domw30: Updated to use CSS

PR #1681, opened 2024-04-02
feat: use a right-click interaction to add & remove conversations from favourites

domw30:
### What does this do?
- uses right click to open the more menu dropdown items
- removes the three dot menu trigger for the dropdown (in favor of right click)

- new requirement request

### How do I test this?
- run tests as usual.
- run branch and right click on conversation items to click either 'favorite' or 'unfavorite' menu items

#### Things to consider:
  1. Does this change any APIs?

https://github.com/zer0-os/zOS/assets/39112648/5d88579d-edbb-4c87-bda6-ef5a7de77cb5

### DIFF
--- a/src/components/messenger/list/conversation-item/index.tsx
+++ b/src/components/messenger/list/conversation-item/index.tsx
@@ -117,7 +117,7 @@ export class ConversationItem extends React.Component<Properties, State> {
     };
 
     return (
-      <div {...cn('more-menu-container')} onClick={stopPropagation}>
+      <div onClick={stopPropagation}>
         <MoreMenu
           isFavorite={this.props.conversation.isFavorite}
           onFavorite={this.onFavorite}
--- a/src/components/messenger/list/conversation-item/index.tsx
+++ b/src/components/messenger/list/conversation-item/index.tsx
@@ -111,6 +111,7 @@ export class ConversationItem extends React.Component<Properties, State> {
       />
     );
   }
+
   renderMoreMenu() {
     const stopPropagation = (e) => {
       e.stopPropagation();
--- a/src/components/messenger/list/conversation-item/index.tsx
+++ b/src/components/messenger/list/conversation-item/index.tsx
@@ -29,7 +29,15 @@ export interface Properties {
   onUnfavoriteRoom: (roomId: string) => void;
 }
 
-export class ConversationItem extends React.Component<Properties> {
+export interface State {
+  isContextMenuOpen: boolean;
+}
+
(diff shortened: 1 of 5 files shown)


src/components/messenger/list/conversation-item/index.tsx
```diff
@@ -94,17 +111,20 @@ export class ConversationItem extends React.Component<Properties> {
...
   }
+
   renderMoreMenu() {
     const stopPropagation = (e) => {
       e.stopPropagation();
     };
 
     return (
-      <div {...cn('more-menu-container')} onClick={stopPropagation}>
+      <div onClick={stopPropagation}>
```
dalefukami, line 121: Does this still need the `stopPropogation`?

dalefukami (commented):
I think this is supposed to come with new instructions on the empty favorites list so the user can figure out that they're supposed to right click?

dalefukami (approved):

src/components/messenger/list/conversation-item/index.tsx
domw30: Yeah I tested with/without and `stopPropogation` is still required to prevent the conversation being selected (opened)

domw30:
> I think this is supposed to come with new instructions on the empty favorites list so the user can figure out that they're supposed to right click?
> 
> Edit: that can come as a follow up since this is still feature flagged.

Yeah correct, the copy for the empty state will be updated in a follow up :)

src/components/messenger/list/conversation-item/index.tsx
dalefukami: the conversation opens when you right click? I thought this was just to prevent it from opening when you click the 3 dots but since we don't have the 3 dots now.... it shouldn't matter?

domw30: Not if you right click, but when selecting the menu item `Favorite` or `Unfavorite`, the conversation will be selected and open (without the `stopPropogation`)

dalefukami: That feels more like a bug in the DropdownMenu then.

domw30: Do you think with the current implementation this will be safe enough to remove favorites feature flag and deploy to prod? I'm assuming I shouldn't wait for a QA on this

dalefukami: I think it's good enough to deploy. Whether it should be enabled is Kelly's decision.

PR #1688, opened 2024-04-03
feat: add favorites error state

domw30:
### What does this do?
- adds a basic error alert if favorited room or unfavorited room fails.

- handle errors gracefully / inform user favorite update failed.

### How do I test this?
- run tests as usual.
- throw error when favoriting or unfavoriting.

#### Things to consider:
  1. Does this change any APIs?

https://github.com/zer0-os/zOS/assets/39112648/d12ff873-6fb7-4f83-94e7-891d0a1e480c

### DIFF
--- a/src/components/messenger/list/conversation-item/index.tsx
+++ b/src/components/messenger/list/conversation-item/index.tsx
@@ -16,6 +16,8 @@ import { bemClassName } from '../../../../lib/bem';
 import './conversation-item.scss';
 import '../styles.scss';
 
+import { FeatureFlag } from '../../../feature-flag';
+
 const cn = bemClassName('conversation-item');
 
 export interface Properties {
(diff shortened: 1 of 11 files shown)


ratik21:
To confirm, if if there multiple error states (i.e if i click on multiple conversations to `favourite` them and they all fail), then the error would stack on top of each other, or remain same ?

domw30:
> To confirm, if if there multiple error states (i.e if i click on multiple conversations to `favourite` them and they all fail), then the error would stack on top of each other, or remain same ?

It would remain the same. If error is true, that alert will show in the UI, no stacking. Basic implementation without a design to roll with. What dya think? worth sticking with this or something a bit clearer/nicer ux?

ratik21:
> It would remain the same. If error is true, that alert will show in the UI, no stacking. Basic implementation without a design to roll with. What dya think? worth sticking with this or something a bit clearer/nicer ux?

UI looks good to me!

PR #1704, opened 2024-04-05
feat(mentions): redesign suggestions list and apply highlight filter

domw30:
### What does this do?
- updates suggestions list styles and applies highlight filter to suggestion search results.

- align with figma designs

### How do I test this?
- run tests as usual.
- run UI and manually test mentions.

#### Things to consider:
  1. Does this change any APIs?

Before:
Screenshot 2024-04-05 at 13 00 56
Screenshot 2024-04-05 at 13 02 06

After:
Screenshot 2024-04-05 at 13 00 29
Screenshot 2024-04-05 at 13 01 28

### DIFF
--- a/src/components/message-input/mentions/index.tsx
+++ b/src/components/message-input/mentions/index.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 
 import { UserForMention } from '../utils';
 import { bemClassName } from '../../../lib/bem';
+import { highlightFilter } from '../../messenger/lib/utils';
 import { userMentionsConfig } from './mentions-config';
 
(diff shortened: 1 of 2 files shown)


src/components/message-input/mentions/index.tsx
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 
 import { UserForMention } from '../utils';
 import { bemClassName } from '../../../lib/bem';
+import { highlightFilter } from '../../messenger/lib/utils';
```
domw30, line 5: shall move`highlightFilter` into a higher up lib folder and update import paths where used in a follow up PR towards end of sprint

PR #1801, opened 2024-05-03
chore(background): enable background on auth pages

developerfred:
### What does this do?

* remove BodyClassManager

fix this issue: ZOS-29 

#### Things to consider:
  1. How will this affect performance? yeah


src/main.scss
```diff
@@ -4,6 +4,7 @@
 
 @import 'variables';
 @import 'layout';
+@import './functions';
```
domw30, line 7: `./functions` to `functions`

src/main.scss
```diff
@@ -43,6 +44,29 @@ $border-color-hover: theme-zui.$color-primary-7;
...
+    background-color: #0a0a0a;
+    background-image: transparent;
+    background-size: cover;
+    background-repeat: no-repeat;
+    background-position: center top;
+
+    transition: background-color animation.$animation-duration-double ease-out,
+      background-image animation.$animation-duration-double ease-out;
+
+    --box-radius: 16px;
```
domw30, line 59: You're only using one of these. You don't need the rest.
```
    --animation-duration-double: #{animation.$animation-duration-double};
    ```

**Review by domw30 — COMMENTED** [review]
Okay this appears to fix the issue where we see the app background before the login screen. I think some work still needs to be done for fading the background in when loading the app. 

I think you can also remove 

```
body {
  transition: background 0.5s ease-out, opacity 0.5s ease-in-out;
}
```

That was added a few days ago

PR #1804, opened 2024-05-03
feat(sidekick): close right sidekick when opening a different conversation

domw30:
### What does this do?
- closes right sidekick when opening a different conversations.

- prevents panels being open when they shouldn't be.

### How do I test this?
- run tests as usual.
- run ui and switch conversations whilst the right side kick is open

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/store/channels/saga.ts
+++ b/src/store/channels/saga.ts
@@ -9,6 +9,8 @@ import { mostRecentConversation } from '../channels-list/selectors';
 import { setActiveConversation } from '../chat/saga';
 import { ParentMessage } from '../../lib/chat/types';
 import { rawSetActiveConversationId } from '../chat';
+import { isSecondarySidekickOpenSelector } from '../group-management/selectors';
+import { toggleIsSecondarySidekick } from '../group-management/saga';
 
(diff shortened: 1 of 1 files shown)


src/store/channels/saga.ts
```diff
@@ -52,6 +54,11 @@ export function* openConversation(conversationId) {
 
   yield call(setActiveConversation, conversationId);
   yield spawn(markConversationAsRead, conversationId);
+
+  const isSidekickOpen = yield select(isSecondarySidekickOpenSelector);
+  if (isSidekickOpen) {
+    yield spawn(toggleIsSecondarySidekick);
+  }
```
dalefukami, line 61: I think it might make sense to have a specific `closeSecondarySidekick` and then you can just call that from here. That keeps this method much more scannable (i.e., developers don't have to interpret the `if` statement at all. It's just, "oh, open the conversation, mark messages as read, close the sidekick")

Then, I also don't think `closeSecondarySidekick` has to actually check if it's open... you can just set the boolean to false. If it doesn't change, nothing happens.

PR #1809, opened 2024-05-03
fix(conversation): fix bug open last chating conversation

developerfred:
### What does this do?
When visiting zos without a conversationId specified in the URL, the most recent conversation should open

#### Things to consider:
  1. Does this change any APIs?


src/components/messenger/list/conversation-list-panel/index.tsx
```diff
@@ -53,6 +53,30 @@ export class ConversationListPanel extends React.Component<Properties, State> {
     this.scrollContainerRef = React.createRef();
   }
 
+  componentDidMount() {
+    this.openMostRecentConversation();
+  }
+
+  openMostRecentConversation = () => {
```
domw30, line 60: This won't be required, we shouldn't be handling this sort of functionality at component level. Please take a look at the saga where the logic lives for handling conversations. We already have a `openFirstConversation` saga function that handles the opening of the first conversation. If there is an issue, this is where you need to investigate. It would probably be worth looking into the room events, if they are being handled/received correctly to determine what conversation to open.

`src/store/channels/saga.ts` 
`src/lib/chat/matrix-client.ts`

src/components/messenger/list/conversation-list-panel/conversation-list-panel.scss
```diff
@@ -58,7 +58,7 @@ $side-padding: 16px;
       }
 
       .messages-list__tab-badge {
-        display: none;
+        display: block;
```
domw30, line 61: Is this change related to opening the last conversation bug?

Looks like this change is to fix a different issue, the unread count on favourites tab? If so, this should be fixed in a different pull request/commit. Try to keep commits separate

domw30 (changes requested):

src/components/messenger/list/conversation-list-panel/conversation-list-panel.scss
developerfred: ok, I will do this!

src/store/channels/saga.ts
```diff
@@ -142,6 +146,7 @@ export function* roomUnfavorited(action) {
 }
 
 export function* saga() {
+  yield call(openFirstConversation);
```
ratik21, line 149: @developerfred this won't make this call upon first load. In order to do that, you should probably look into the page-load saga.

ratik21 (commented):
@developerfred @domw30 do we need this? 

I tested on prod & local, if i open `localhost:3000` directly, then it still takes me to the first conversation in the list. What's the issue we're trying to solve w/ this PR?

image

this ^ already happens on prod.

src/store/channels/saga.ts
developerfred: ok, cool!

developerfred:
> @developerfred @domw30 do we need this?
> 
> I tested on prod & local, if i open `localhost:3000` directly, then it still takes me to the first conversation in the list. What's the issue we're trying to solve w/ this PR?
> 
> image
> this ^ already happens on prod.

**To reproduce the error you need the following flow**: 

- start a chat with a new user, write some messages
- then open a new user, write one again

agpra opens the chat in a new tab and checks who will appear first, currently chat A appears and not chat B, which was the last one you talked to. This pull request fixes this!

ratik21:
@developerfred can paste a video here reproducing this? Seems to work fine on mine. 

This is how it looks 

https://github.com/zer0-os/zOS/assets/33264364/c1ec6e21-557f-49cd-b8d7-613b071a2625

This is the correct (expected) behaviour.

PR #1810, opened 2024-05-04
Allow `Reply to a reply`

ratik21:
### What does this do?

Allows to reply to a reply.

https://github.com/zer0-os/zOS/assets/33264364/e317f2e4-17f0-45e6-b4cb-dd6cce407dbd


domw30 (commented):
🔥

PR #1817, opened 2024-05-06
feat(conersation): mandatory-conversation-name

developerfred:
### What does this do?

only allows you to create chat groups with names

*change createGroup
*change conversation saga
*add logic to errorDialog

#### Things to consider:
  1. How will this affect performance? no
  1. Does this change any APIs? yes


src/components/messenger/list/group-details-panel/index.tsx
```diff
@@ -58,7 +65,12 @@ export class GroupDetailsPanel extends React.Component<Properties, State> {
             onChange={this.nameChanged}
             {...cn('name-input')}
             placeholder='Group name...'
+            isRequired={true}
+            aria-invalid={!this.state.isNameValid}
           />
+          {!this.state.isNameValid && (
+            <div style={{ color: 'rgba(221, 50, 50, 0.75)' }}>Please enter a group name.</div>
+          )}
```
ratik21, line 73: I think you should just be able to use the `error` & `errorMessage` field in the `<Input>` component itself. We don't need `area-invalid` & an extra state.

ratik21 (changes requested):
Another reference should be `src/components/edit-profile/index.tsx`

src/components/messenger/list/group-details-panel/index.tsx
developerfred: Thanks, I’m going to change it to this component!

domw30: Yep as Ratik suggested. Also, I'm seeing a number of inline styles being added on tasks. We should be using the scss stylesheets to set css, not the `style` attribute on elements. Thanks!

developerfred: good tip

developerfred: @ratik21 @domw30 ref here: 
#9547793

src/components/messenger/list/group-details-panel/index.tsx
```diff
@@ -14,34 +14,43 @@ const cn = bemClassName('group-details-panel');
...
-    this.setState({ name: value });
-  };
+  nameChanged = (value) => this.setState({ name: value });
+  onImageChange = (image) => this.setState({ image });
+  back = () => this.props.onBack();
 
-  onImageChange = (image) => {
-    this.setState({ image });
-  };
+  get isValid() {
```
ratik21, line 40:
```suggestion
  get isNameValid() {
```

src/components/messenger/list/group-details-panel/index.tsx
```diff
@@ -14,34 +14,43 @@ const cn = bemClassName('group-details-panel');
...
+    return typeof this.state.name === 'string' && this.state.name.trim().length > 0;
+  }
 
-  back = () => {
-    this.props.onBack();
-  };
+  get nameError() {
+    if (!this.isValid) {
+      return { variant: 'error', text: 'Please enter a group name.' } as any;
```
ratik21, line 46: Let's just rename this to `name is required`? So the error message is one-lined?

ratik21 (approved):

src/components/messenger/list/group-details-panel/index.tsx
```diff
@@ -14,34 +14,43 @@ const cn = bemClassName('group-details-panel');
...
-  };
+  nameChanged = (value) => this.setState({ name: value });
+  onImageChange = (image) => this.setState({ image });
+  back = () => this.props.onBack();
 
-  onImageChange = (image) => {
-    this.setState({ image });
-  };
+  get isValid() {
+    return typeof this.state.name === 'string' && this.state.name.trim().length > 0;
```
domw30, line 41: You can remove the `typeof this.state.name === 'string'`

` ``
get isValid() {
    return this.state.name.trim().length > 0;
  }
  ```

domw30 (commented):
This looks fine guys, though I have a couple of questions.

Why don't we just keep things simple here. Disable the submit button if there is no name? There isn't really a need for error handling and doing this means an extra click for the user. Just prevent the user submitting and disable the button if `name.length < 0` or `name === ''` ??

Then add a test to ensure the button is disabled when input is empty, and enabled when it is not.

ratik21:
> Why don't we just keep things simple here. Disable the submit button if there is no name? 

Don't have a strong preference here, but i think it's better UX to display an error state as well along with disabling the button. 

For example i can also create a group without an image. If i just disable the button (on no name), some user might get confused as to "why" the button was disabled. Was it due to no name, or no image?

domw30:
> > Why don't we just keep things simple here. Disable the submit button if there is no name?
> 
> Don't have a strong preference here, but i think it's better UX to display an error state as well along with disabling the button.
> 
> For example i can also create a group without an image. If i just disable the button (on no name), some user might get confused as to "why" the button was disabled. Was it due to no name, or no image?

I think simple is better. Showing the error when immediately clicking into the input because there is no value is not usually normal behaviour.

src/components/messenger/list/group-details-panel/index.tsx
developerfred: ```trim()```  is only available in typescript for strings, while it does not know that it is a string typescript warning

src/components/messenger/list/group-details-panel/index.tsx
developerfred: ok

src/components/messenger/list/group-details-panel/index.tsx
```diff
@@ -54,10 +54,13 @@ export class GroupDetailsPanel extends React.Component<Properties, State> {
...
 
           <Input
-            value={this.state.name}
+            value={name}
             onChange={this.nameChanged}
-            {...cn('name-input')}
             placeholder='Group name...'
+            isRequired={true}
+            error={!this.isNameValid}
+            alert={!this.isNameValid ? { variant: 'error', text: 'Please enter a group name.' } : undefined}
```
ratik21, line 62: I think Dom is right here, I misunderstood the original problem. We can just keep the button disabled and as soon as the user types a name, we can enable it.  this code can be removed

src/components/messenger/list/group-details-panel/index.tsx
```diff
@@ -28,24 +27,25 @@ export class GroupDetailsPanel extends React.Component<Properties, State> {
   state = { name: '', image: null };
 
   createGroup = () => {
-    this.props.onCreate({ name: this.state.name, users: this.props.users, image: this.state.image });
+    if (this.isNameValid) {
+      this.props.onCreate({ name: this.state.name, users: this.props.users, image: this.state.image });
+    }
```
ratik21, line 32: Ideally we shouldn't need this check. `createGroup` should never be called if the name is empty (since the button is disabled)

domw30:
This is all we need : https://github.com/zer0-os/zOS/pull/1846

PR #1822, opened 2024-05-07
Send typing events to matrix homeserver  when user starts typing

ratik21:
### What does this do?

Implements the following logic:

- As soon as you start typing in the <MessageInput/> component, we start publishing a "typing" event (which we'll later process on the client side)
- the above is built into a race saga condition, as we need to publish "stopped typing" when: 
  - it has been 5s & the user hasn't typed anymore
  - the users "sends" a message in chat
  - the user changes the room (opens another conversation). in this case we should publish stop typing event for the previous room.
  - users logs out of the app

- We also store `currentUserLastTypingEventPublishedAt` in redux for each `Channel`, to ensure we only send the next typing event AFTER the timeout (which is 5s). Since the API is rate-limited.

Doc reference: [https://spec.matrix.org/v1.10/client-server-api/#typing-notifications](https://spec.matrix.org/v1.10/client-server-api/#typing-notifications)

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -42,6 +42,7 @@ import { encryptFile } from './matrix/media';
 import { uploadAttachment } from '../../store/messages/api';
 import { featureFlags } from '../feature-flags';
 import { logger } from 'matrix-js-sdk/lib/logger';
+import { USER_TYPING_TIMEOUT } from '../../store/channels/saga';
 
 export class MatrixClient implements IChatClient {
   private matrix: SDKMatrixClient = null;
@@ -685,6 +686,12 @@ export class MatrixClient implements IChatClient {
 
     const result = await this.matrix.getRoomTags(roomId);
     return !!result.tags?.[MatrixConstants.FAVORITE];
+  }
+
+  async sendTypingEvent(roomId: string, isTyping: boolean): Promise<void> {
+    await this.waitForConnection();
+
+    this.matrix.sendTyping(roomId, isTyping, USER_TYPING_TIMEOUT);
   }
 
   arraysMatch(a, b) {
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -42,7 +42,8 @@ import { encryptFile } from './matrix/media';
 import { uploadAttachment } from '../../store/messages/api';
 import { featureFlags } from '../feature-flags';
 import { logger } from 'matrix-js-sdk/lib/logger';
-import { USER_TYPING_TIMEOUT } from '../../store/channels/saga';
+
+export const USER_TYPING_TIMEOUT = 5000; // 5s
 
 export class MatrixClient implements IChatClient {
   private matrix: SDKMatrixClient = null;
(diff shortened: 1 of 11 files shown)


src/lib/chat/matrix-client.ts
```diff
@@ -42,6 +42,7 @@ import { encryptFile } from './matrix/media';
 import { uploadAttachment } from '../../store/messages/api';
 import { featureFlags } from '../feature-flags';
 import { logger } from 'matrix-js-sdk/lib/logger';
+import { USER_TYPING_TIMEOUT } from '../../store/channels/saga';
```
dalefukami, line 45: Probably best not to import saga information here. The matrix client really shouldn't "know" things about the rest of the system.

src/store/channels/saga.ts
```diff
@@ -136,7 +145,63 @@ export function* roomUnfavorited(action) {
...
+    if (!channel) {
+      return;
+    }
+
+    const currentUserLastTypingEventPublishedAt = channel.currentUserLastTypingEventPublishedAt;
+    const now = Date.now();
+    if (currentUserLastTypingEventPublishedAt && now - currentUserLastTypingEventPublishedAt <= USER_TYPING_TIMEOUT) {
+      // not enough time has passed since last typing event
+      return;
+    }
```
dalefukami, line 160: Is this really what we want? Assuming the matrix typing event timeout that we passed in means that it'll clear the typing status after that timeout.... if I were to type non-stop for 4999ms seconds and then pause for a second, the typing status would clear. Should it clear if I _just_ typed something a couple ms ago?

Maybe on line 203 we should just use this instead: https://redux-saga.js.org/docs/api#debouncems-pattern-saga-args

```
yield debounce(1000, SagaActionTypes.UserTypingInRoom, publishUserTypingEvent)
```

Doesn't even really matter if the room changes because the time between typing in one room, switching rooms, and typing in the next room will pretty much cross the 1 second debounce after a few characters anyway.

That removes the complexity of managing this extra state of the "last published" time.

src/store/channels/saga.ts
```diff
@@ -136,7 +145,63 @@ export function* roomUnfavorited(action) {
...
+
+    yield call(matrixSendUserTypingEvent, roomId, true);
+    yield call(receiveChannel, { id: roomId, currentUserLastTypingEventPublishedAt: now });
+  } catch (error) {
+    console.error(`Failed to publish user is typing event in room ${roomId}:`, error);
+  }
+}
+
+export function* publishUserStoppedTypingEvent(roomId) {
```
dalefukami, line 169: Do we actually have to do this? Does the timeout that we sent to the matrix server not just clear it after 5 seconds anyway?

src/store/channels/saga.ts
ratik21: as discussed on call, updated to use `leadingDebounce`

src/store/channels/saga.ts
ratik21: Update to only call this after message is sent

src/store/messages/saga.ts
```diff
@@ -196,6 +201,7 @@ export function* send(action) {
 
   yield call(uploadFileMessages, channelId, rootMessageId, uploadableFiles);
   yield call(publishMessageSent);
+  yield call(publishUserStoppedTypingEvent, channelId);
```
dalefukami, line 204: Small note, I think it probably would have been cleaner to listen for the chatMessageBus event in the other saga file. That way all the typing stuff is together.

ratik21: makes sense! Updated.

PR #1832, opened 2024-05-08
feat(messenger): add user profile container and component

domw30:
### What does this do?
- adds user profile container and component.

- to be able to render user profile panels .
- as per edit profile/profile redesigns

### How do I test this?
- run tests as usual. 
- this has not yet been wired up in the UI to test manually.

#### Things to consider:
  1. Does this change any APIs?
  
  
Screenshot 2024-05-08 at 11 04 36

### DIFF
--- a/src/components/messenger/user-profile/index.tsx
+++ b/src/components/messenger/user-profile/index.tsx
@@ -1,0 +1,58 @@
+import * as React from 'react';
+
+import { bemClassName } from '../../../lib/bem';
+
+import { PanelHeader } from '../list/panel-header';
+import { Image } from '@zero-tech/zui/components';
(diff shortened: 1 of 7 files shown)


src/components/messenger/user-profile/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+export interface Properties {
+  name: string;
+  image: string;
+  subHandle?: string;
+
+  onBack: () => void;
+}
+
+export class UserProfile extends React.Component<Properties> {
+  navigateBack = () => {
```
domw30, line 22: Note : I'll use this in a follow up PR. Currently passing the prop into onBack instead.

PR #1833, opened 2024-05-08
feat(sidekick): wire up user profile - handle and render with open profile state

domw30:
### What does this do?
- wires up the handling and displaying of the user profile in the sidekick.

- as per design.

### How do I test this?
- run tests as usual. 
- you will not be able to render this in the UI just yet as follow up work is required to trigger opening the profile.

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/sidekick/index.tsx
+++ b/src/components/sidekick/index.tsx
@@ -8,6 +8,7 @@ import { GroupManagementContainer } from '../messenger/group-management/containe
 import classNames from 'classnames';
 import { bemClassName } from '../../lib/bem';
 import './styles.scss';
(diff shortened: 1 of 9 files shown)


src/components/sidekick/index.tsx
```diff
@@ -46,7 +49,7 @@ export class Container extends React.Component<Properties> {
     if (this.isSecondary) {
       return <GroupManagementContainer />;
     } else {
-      return <MessengerList />;
+      return this.props.isUserProfileOpen ? <UserProfileContainer /> : <MessengerList />;
```
domw30, line 52: I will improve the handling of these containers as follow ups when continuing work on users profile

PR #1841, opened 2024-05-09
chore(messages): fix message image jump

developerfred:
### What does this do?

There was a glitch where every time we sent a message with an image, now with this fix, it no longer occurs.
So, what we did here was to modify the following functions in the redux message saga, send and performSend, adding a check to see if the image has already been sent to prevent it from being sent twice.

screen movie 
https://youtu.be/2U8xlfX2mR8

### How do I test this?

try send menssage with image and text 

#### Things to consider:
  1. How will this affect performance? yes


src/store/messages/saga.ts
```diff
@@ -253,31 +253,50 @@ export function* createOptimisticPreview(channelId: string, optimisticMessage) {
...
+  }
 }
 
 export function* sendMessage(apiCall, channelId, optimisticId) {
+  let messageSentSuccessfully = false;
+
   try {
+    if (messageSentSuccessfully) {
+      return;
+    }
```
ratik21, line 289: i don't .. follow. will this ever execute? You've just declared `messageSentSuccessfully` to be `false`

developerfred: fixed now!

developerfred:
@domw30 If you could take a look at my pull request, thank you very much!

src/store/messages/saga.ts
```diff
@@ -253,28 +253,32 @@ export function* createOptimisticPreview(channelId: string, optimisticMessage) {
...
-      chatClient.sendMessagesByChannelId,
-    ],
-    channelId,
-    message,
-    mentionedUserIds,
-    parentMessage,
-    null,
-    optimisticId
-  );
-  return yield sendMessage(messageCall, channelId, optimisticId);
```
domw30, line 270: So we no longer want to `sendMessage`? This feels incorrect to remove

domw30 (commented):
Could you explain why you've removed the line I've commented on plz

src/store/messages/saga.ts
domw30: This appears to be working because you are no longer replacing the optimistic message

developerfred: Thanks @domw30, apparently everything is working (Reply and editing messages).
I will return with sendMessage, preventing unnecessary re-rendering.

developerfred: https://github.com/zer0-os/zOS/assets/57037080/ecc44d19-39a3-405f-a6d2-403fb0c47fc3

PR #1844, opened 2024-05-09
fix: right click delete option

developerfred:
### What does this do?
fix right click button
enable right click save image 

### Video Test

https://youtu.be/5-aS2MbS_WQ

#### Things to consider:
  1. How will this affect performance? no

### DIFF
--- a/src/components/message/styles.scss
+++ b/src/components/message/styles.scss
@@ -9,6 +9,7 @@
   gap: 8px;
   overflow: hidden;
   width: 100%;
+  position: relative;
 
   &--owner {
     flex-direction: row-reverse;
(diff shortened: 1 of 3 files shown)


src/components/message/styles.scss
```diff
@@ -144,6 +145,14 @@
     opacity: 0;
     // Sometimes we render an empty menu. This ensures consistency of width of the message bubble.
     min-width: 32px;
+
+    &--hidden {
+      display: none !important;
```
ratik21, line 150: please avoid using `! important` in your css

ratik21 (commented):
I am wondering .. do we really need all that extra code, just to get `Delete` working on right click? I thought we just needed to "trigger" the same code for `Delete` as we do when clicking on `Delete` option from the 3 dot menu.

developerfred:
> I am wondering .. do we really need all that extra code, just to get `Delete` working on right click? I thought we just needed to "trigger" the same code for `Delete` as we do when clicking on `Delete` option from the 3 dot menu.

We were having a scope problem so I needed to change the code. Then we will need the code.

What have I done here? It was to isolate the interactions as much as possible so that the messageMenu would not have any problems when being called.

PR #1873, opened 2024-05-14
Add a function to lower minimum `invite` & `kick` levels for existing groups during initial load

ratik21:
### What does this do?

We need to add a new feature: `Add Moderators`. 

Currently we have minimum `invite, kick` level set to 100 for all users. This means that only the room creator/admin (which has a `power_level` of 100) can invite and kick users. 

We need to bring this down to `50` for existing groups, so that the admins of that groups can set users as moderators (with a `power_level` of `50`), and they can `invite`/`kick` users. Additionally, admins will be able to add/remove moderators.

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -159,6 +159,42 @@ export class MatrixClient implements IChatClient {
     return admins;
   }
 
+  /**
+   * Currently we have minimum invite, kick level set to 100 for all users
+   * this means that only the room creator/admin (which has a power_level of 100) can invite and kick users
+   *
+   * this function will lower the minimum invite and kick levels to 50 for all rooms,
+   * so that the admin can set other users as moderators and they can invite and kick users
+   */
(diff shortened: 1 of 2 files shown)


src/lib/chat/matrix-client.ts
```diff
@@ -159,6 +159,46 @@ export class MatrixClient implements IChatClient {
     return admins;
   }
 
+  /**
+   * Currently we have minimum invite, kick level set to 100 for all users
+   * this means that only the room creator/admin (which has a power_level of 100) can invite and kick users
+   *
+   * this function will lower the minimum invite and kick levels to 50 for all rooms,
```
dalefukami, line 166: Suggestion: Make sure you have a task set for a week(?) from now to remove this. Then, you can just adjust the room when the admin goes to add new moderators if it isn't yet set up for it.

ratik21: noted, thanks

PR #1886, opened 2024-05-15
Update `Edit Group` Flow: Show 3 dot menu instead of `x` icon to remove member

ratik21:
### What does this do?

Updates the members list in the **edit group panel** to use `3 dot` menu instead of the `x` icon. This is to allow more options to manage the members of the group. The admin can now either
a) Remove a member from the group
b) Set/Remove a member as a "moderator" from the group

Part b) is currently in implementation.

Before:

https://github.com/zer0-os/zOS/assets/33264364/abf855ce-2578-4919-ab17-ca3473beb6a7

After:

https://github.com/zer0-os/zOS/assets/33264364/36578d9c-5b24-44dc-b87a-a650b2bd6751

### DIFF
--- a/src/components/messenger/group-management/member-management-menu/container.tsx
+++ b/src/components/messenger/group-management/member-management-menu/container.tsx
@@ -1,0 +1,53 @@
+import * as React from 'react';
+
+import { RootState } from '../../../../store/reducer';
+import { connectContainer } from '../../../../store/redux-container';
+import { openRemoveMember } from '../../../../store/group-management';
+
+import { MemberManagementMenu } from '.';
+
+export interface PublicProperties {
+  user?: any;
+  canRemove?: boolean;
+
+  onOpenChange: (isOpen: boolean) => void;
+}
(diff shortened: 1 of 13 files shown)


src/components/messenger/group-management/member-management-menu/container.tsx
```diff
@@ -0,0 +1,53 @@
...
+      chat: { activeConversationId },
+    } = state;
+
+    return {
+      activeConversationId,
+    };
+  }
+
+  static mapActions(_props: Properties): Partial<Properties> {
+    return { openRemoveMember };
```
ratik21, line 34: I've added this container and moved `onRemoveMember` saga prop over there, since we're also gonna add more functions here: `setModerator`, `removeModerator` etc.

domw30 (commented):
Looks great to me mate 💪 💯 I personally would have just added the member-management-menu in a separate PR then updated the panel/citizen item.

PR #1904, opened 2024-05-20
fix: desktop css mask

thewarman:
#### Things to consider:
  1. Does this change any APIs?


domw30 (commented):
Sorry @thewarman my bad, syntax needs to be like so:

```
  html.todesktop &:before {
    background-image: none;
    mask: none;
  }
  ```

thewarman:
@domw30 updated.

domw30 (approved):

PR #1910, opened 2024-05-21
Render Users with 'Mod' tags, for users who are moderators of group in various panels

ratik21:
### What does this do?

Renders '`Mod`' tag for users who are moderators of the group (`view-members-panel`, `group-info-panel`, `edit-group-panel`)

`Members Panel`
image

`Group Info Panel`
image

`Edit Group Panel`
image

### DIFF
--- a/src/components/group-management/member-management-dialog/index.test.tsx
+++ b/src/components/group-management/member-management-dialog/index.test.tsx
@@ -33,19 +33,11 @@ describe(MemberManagementDialog, () => {
       const wrapper = subject({
         definition: {
           ...mockConfirmitionDefinition,
-          getMessage: () => (
-            <>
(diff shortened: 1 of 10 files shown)


src/components/group-management/member-management-dialog/index.test.tsx
```diff
@@ -33,19 +33,11 @@ describe(MemberManagementDialog, () => {
...
         definition: {
           ...mockConfirmitionDefinition,
-          getMessage: () => (
-            <>
-              Are you sure you want to remove Johnny Cash from Fun Room?
-              <br />
-              Johnny Cash will lose access to the conversation and its history.
-            </>
-          ),
+          getMessage: () => <>Test Message</>,
```
ratik21, line 36: @dalefukami just updated this here since it was a small change

PR #1912, opened 2024-05-22
Send Admin Message when a user is set (or removed) as moderator

ratik21:
### What does this do?

Sends an admin message if a user is 
- set as Moderator (`User X was set as moderator by admin`)
- removed as Moderator (`User X was removed as moderator by admin`)

image

https://github.com/zer0-os/zOS/assets/33264364/62b30516-f44c-424b-891c-5fec2bc2de0b


src/lib/chat/matrix/chat-message.ts
```diff
@@ -129,6 +132,50 @@ function getRoomMemberAdminData(content, targetUserId) {
...
+  // we need to find the user who's power level has changed
+  const usersCurrentPowerLevels: { [userId: string]: number } = content.users;
+  const usersPreviousPowerLevels: { [userId: string]: number } = previousContent.users;
+  const changedUserIds = Object.keys(usersCurrentPowerLevels).filter(
+    (userId) => usersCurrentPowerLevels[userId] !== usersPreviousPowerLevels[userId]
+  );
+
+  if (changedUserIds.length !== 1) {
+    return null;
+  }
```
dalefukami, line 149: Can it never be more than one? Is it not possible, in Matrix, to change 2 users power levels at the same time or possibly would a Matrix server batch the events if it happened quickly?

ratik21: from the UI we only ever "set" (or "remove) a single user as a moderator. Technically you can update multiple user's power_levels at once, but we don't do that (from the UI), we only update a single user's power_level at once. 

Which is why for the "power level changed" events that we get from loading all room events, we get them as single separate events in the list. (for eg. I had tested making multiple users as moderators and removing them consecutively, and when i had reloaded i got the list of events w/ each event as separate).

> or possibly would a Matrix server batch the events if it happened quickly?

not sure about this one 🤔 . is that possible ..? Has that happened w/ any other kind of event? (eg. `member_left`, `member_joined` etc)? Right now i always get separate events (not batched).

dalefukami: Yes, "we" only ever do it one at a time. But, remember, Matrix is an open system and zOS is open source. Anyone can do anything.

ratik21: okay, so if we do, then i guess we can print multiple admin messages in that case. will handle it.

PR #1915, opened 2024-05-22
feat(matrix-client): matrix set receipt preferences and get receipt preferences when reading a room

domw30:
Note: UI updates in follow up.

### What does this do?
- Implements user-configurable read receipt preferences with a default to private if feature flagged, default to public if not, and includes a feature flag for enabling this functionality 
- test coverage

- To give users control over their read receipt visibility, enhancing privacy and user experience.

### How do I test this?
- run tests as usual.
- this is not yet wired up so you can not test manually in the UI without code changes to do so.

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -36,6 +36,7 @@ import {
   IN_ROOM_MEMBERSHIP_STATES,
   MatrixConstants,
   MembershipStateType,
+  ReadReceiptPreferenceType,
 } from './matrix/types';
 import { constructFallbackForParentMessage, getFilteredMembersForAutoComplete, setAsDM } from './matrix/utils';
 import { uploadImage } from '../../store/channels-list/api';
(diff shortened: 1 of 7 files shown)


src/store/channels/saga.ts
```diff
@@ -220,6 +221,15 @@ export function* receivedRoomMemberPowerLevelChanged(action) {
   yield call(receiveChannel, { id: roomId, moderatorIds });
 }
 
+export function* onSetReadReceiptPreference(action) {
+  const { preference } = action.payload;
+  try {
+    yield call(setReadReceiptPreference, preference);
+  } catch (error) {
+    console.error('Failed to set read receipt preference:', error);
```
domw30, line 229: will need to add error state

src/lib/chat/matrix-client.ts
```diff
@@ -678,6 +679,20 @@ export class MatrixClient implements IChatClient {
...
   }
 
+  async setReadReceiptPreference(preference: ReadReceiptPreferenceType) {
+    await this.matrix.setAccountData(MatrixConstants.READ_RECEIPT_PREFERENCE, { readReceipts: preference });
+  }
+
+  async getReadReceiptPreference() {
+    try {
+      const accountData = await this.matrix.getAccountData(MatrixConstants.READ_RECEIPT_PREFERENCE);
+      return accountData?.event?.content?.readReceipts || ReadReceiptPreferenceType.Public;
```
dalefukami, line 689: The default should always be private

src/store/channels/saga.ts
```diff
@@ -220,6 +221,15 @@ export function* receivedRoomMemberPowerLevelChanged(action) {
   yield call(receiveChannel, { id: roomId, moderatorIds });
 }
 
+export function* onSetReadReceiptPreference(action) {
+  const { preference } = action.payload;
+  try {
+    yield call(setReadReceiptPreference, preference);
+  } catch (error) {
```
dalefukami, line 228: Suggestion: This should probably translate between a "Domain" enum and the Matrix enum (or a different option: client wrapper should maybe have "setToPrivate" and "setToPublic" functions). The reason is, you don't really want your UI components to have to import Matrix enumerations in order to do their work. Those should only have knowledge of "Domain" information.

PR #1921, opened 2024-05-24
[FIX] Sometimes the admin messages do not display when you set/remove a `Mod`

ratik21:
### What does this do?

Sometimes the admin messages were not displaying in the UI, after you set/remove a user as a moderator. This was because, sometimes instead of "0" `power_level`, we would just get an "empty"/"undefined" field for the `power_level` in the event content. For eg.

Before
```js
{
  "admin": <100>
  "user-1": <50> // mod
}
```

Now, if we remove this user as mod, then we sometimes get ::

```js
{
  "admin": <100>
  // this is empty
}
```

In normal cases we should get something like:
```js
{
  "admin": <100>
  "user-1": <0> // normal user now
}
```

This PR handles the above case. 

Before (notice one admin message is missing):
image

After:
image

### DIFF
--- a/src/components/group-management/member-management-dialog/styles.scss
+++ b/src/components/group-management/member-management-dialog/styles.scss
@@ -1,7 +1,7 @@
 @use '~@zero-tech/zui/styles/theme' as theme;
 
 .remove-member-dialog {
-  width: 586px;
+  width: 600px;
   line-height: 22px;
 
   & > *:last-child {
--- a/src/lib/chat/matrix/chat-message.test.ts
+++ b/src/lib/chat/matrix/chat-message.test.ts
@@ -28,7 +28,7 @@ describe(getRoomPowerLevelsChangedAdminData, () => {
     expect(result).toBeNull();
   });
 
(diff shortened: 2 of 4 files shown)


src/components/group-management/member-management-dialog/styles.scss
```diff
@@ -1,7 +1,7 @@
 @use '~@zero-tech/zui/styles/theme' as theme;
 
 .remove-member-dialog {
-  width: 586px;
+  width: 600px;
```
dalefukami, line 4: Why keep increasing this? Once the dialog has a size it should pretty much stay that way and whichever text has to wrap around will just wrap around.

ratik21: it is related to this PR, [https://github.com/zer0-os/zOS/pull/1920](https://github.com/zer0-os/zOS/pull/1920]. I hadn't increased it properly so there was still some space at the right of modal, but because of `586px` width the text would wrap around it and it would look weird.

dalefukami: What is providing the actual full size of the modal? Some other place?... if so, don't you just want this text to fill its width? Why hard code a number here?

PR #1923, opened 2024-05-24
Trigger `lowerMinimumInviteAndKickLevels` before setting a user as `Moderator`

ratik21:
### What does this do?

Executes `lowerMinimumInviteAndKickLevels` before setting a user as `Mod` in a room. This is just to "make sure" that all groups have the correct permissions set, for a `Mod` to perform `{ invite, kick }` actions. 

For example, if a room is created via android/iOS, and we don't have the right `power_levels` set for a room in that case, then this should handle it (since we're explicitly correcting the room's power_levels as soon as we set a moderator).

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -191,7 +191,7 @@ export class MatrixClient implements IChatClient {
         }
 
         // just to be safe, above we check if the user is the creator of the room
-        // but also check if this users has a power level of 100,
+        // but we also check if *this* user has a power level of 100,
         // otherwise sendStateEvent will fail
         const users = powerLevelsContent.users || {};
         if (users[this.userId] !== PowerLevels.Owner) {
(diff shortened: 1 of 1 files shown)


src/lib/chat/matrix-client.ts
```diff
@@ -736,6 +736,8 @@ export class MatrixClient implements IChatClient {
 
   async setUserAsModerator(roomId: string, user): Promise<void> {
     await this.waitForConnection();
+
+    this.lowerMinimumInviteAndKickLevels([this.matrix.getRoom(roomId)]);
```
dalefukami, line 740: Needs an `await`? or, are we just assuming it'll eventually catch up and nobody will really notice?

ratik21: > are we just assuming it'll eventually catch up and nobody will really notice?

yeah i mean we don't process that event, & since this is a separate event from `"setUserAsMod"`, didn't feel like this call should be `awaited`. any reason why this should be `awaited`?

dalefukami: The only reason I can think if is what I just asked. If we don't care if the user thinks it work but it isn't done then there's no need to await.

Eg: what if the user sets someone as a mod and then closes the browser prior to the "lower..." finishing? Does it actually complete or does it somehow cancel without finishing? If there's a chance it doesn't finish then the user will _think_ they've done what they wanted but it didn't actually do it.

PR #2004, opened 2024-06-11
feat(reply-card): render media if media is present in parent message

domw30:
### What does this do?
- updates reply card to display media (image) if media url is present.
- message is optional to ensure a reply can contain media only.

- to reply to media messages.

### How do I test this?
- run tests as usual.
- this is not yet wired up to allow replying to media messages so you will not be able to test this yet by running the UI. However, you may enable 'Reply' menu item on the message menu if you'd like to test this (as per below).

#### Things to consider:
  1. Does this change any APIs?

Demo

Uploading imagereply.mov…

### DIFF
--- a/src/components/message-input/index.tsx
+++ b/src/components/message-input/index.tsx
@@ -275,6 +275,8 @@ export class MessageInput extends React.Component<Properties, State> {
               senderIsCurrentUser={this.props.replyIsCurrentUser}
               senderFirstName={reply?.sender?.firstName}
               senderLastName={reply?.sender?.lastName}
+              mediaUrl={reply?.media?.url}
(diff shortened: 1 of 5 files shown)


src/components/chat-view-container/chat-view.test.tsx
```diff
@@ -288,13 +288,14 @@ describe('ChatView', () => {
     });
 
     it('returns the formatted date for the same year', () => {
-      const currentYearDate = moment('2024-06-11'); // Example date within the same year
+      // setting the date here is not ideal as at some point this will fail
+      const currentYearDate = moment('2024-12-11'); // Example date within the same year
```
domw30, line 292: This needs to be improved (follow up)

PR #2020, opened 2024-06-17
chore: upgrade typescript `^4.1.6` -> `^5.0.4`

colbr:
### What does this do?

- Upgrades `typescript` to `^5.0.4`
- Adds `--legacy-peer-deps` to install commands in Github Actions
- Manually add `@emotion/react` as a dependency since it won't be automatically installed (because of `--legacy-peer-deps`)

`wagmi`/`viem` require `typescript` `>=5.0.4`

----

I believe this is working because...

✅ Tests are passing
✅ Build is successful
✅ All Github Actions are working
✅ No new Typescript errors/warnings

Please let me know if there's anything more we should be considering.


colbr:
Ok, so tests aren't working in CI. We will have to alter the scripts so `--legacy-peer-deps` is enabled. I'll change this to a draft PR and remove reviewers until they're fixed!

colbr:
One thing to consider...

Now that we're installing with `--legacy-peer-deps`, it's possible for more unmet peer dependencies to sneak into the codebase. Should we add a CI script to let us know whenever we add another package with unmet peer dependencies?

.github/workflows/production.yml
```diff
@@ -33,7 +33,7 @@ jobs:
 
       - run: echo '//npm.pkg.github.com/:_authToken=${{secrets.GITHUB_TOKEN}}' >> .npmrc
 
-      - run: npm install
+      - run: npm install --legacy-peer-deps
```
ratik21, line 36: could you explain why we need `--legacy..` here? Is it because there are other deps which need to upgraded as well, to actually remove this flag?

ratik21:
> Ok, so tests aren't working in CI. We will have to alter the scripts so `--legacy-peer-deps` is enabled. I'll change this to a draft PR and remove reviewers until they're fixed!

What's the main reason? Are they working locally? Can we update the actions/workflow to actually run the jobs w/ the upgraded typescript version?

.github/workflows/production.yml
colbr: `react-scripts`/`craco` require Typescript 4, so we have to use `--legacy-peer-deps` to install Typescript 5. Unfortunately there's no version of `react-scripts`/`craco` which use Typescript 5

colbr:
> > Ok, so tests aren't working in CI. We will have to alter the scripts so `--legacy-peer-deps` is enabled. I'll change this to a draft PR and remove reviewers until they're fixed!
> 
> What's the main reason? Are they working locally? Can we update the actions/workflow to actually run the jobs w/ the upgraded typescript version?

Mentioned in the comment above, but `react-scripts`/`craco` require Typescript 4. If you pull this branch locally, delete your `node_modules`, and run `npm i` you'll get a peer dependency error.

See the output here:
https://github.com/zer0-os/zOS/actions/runs/9556506335/job/26341905689

ratik21 (approved):

PR #2039, opened 2024-06-25
chore: upgrade to React 18

colbr:
### What does this do?

- React 17 -> React 18. `react`, `react-dom`, and typings, all upgraded.
- Used [the new `root.render` function](https://react.dev/blog/2022/03/08/react-18-upgrade-guide#:~:text=//%20Before,(%3CApp%20tab%3D%22home%22%20/%3E)%3B) at app root.
- Upgraded `testing-library/react` to work with React 18. All of our RTL tests are working as expected.
- Installed and configured an [Enzyme React 18 adapter](https://www.npmjs.com/package/@cfaester/enzyme-adapter-react-18) (`@cfaester/enzyme-adapter-react-18`). All of our Enzyme tests are working as expected.
- Removed `@wojtekmaj/enzyme-adapter-react-17`.
- Upgraded `react-router-dom` to `5.3.2`, [which supports React 18 `StrictMode`](https://github.com/remix-run/react-router/releases/tag/v5.3.2).
- Upgraded `react-redux` to `7.2.8`, [which supports React 18 StrictMode](https://github.com/reduxjs/react-redux/releases/tag/v7.2.8).
- Handles some type issues caused by React 18 not including `children` in intrinsic props (#2040).

### Why are we making this change?

We are taking baby steps getting to the latest versions of all of our tooling. We don't want to be barred out of using a tool because we're on an old version of React.

### How do I test this?

- Run `test` and `test:vitest`.
- Test flows manually.

### Key decisions and Risk Assessment

This one seems high risk, as there are some pretty major dependency changes.

I used [the React 18 migration blog](https://react.dev/blog/2022/03/08/react-18-upgrade-guide) to guide me through the upgrade, so the process was as accurate as possible. Even so, I may have missed something, as I do not yet have complete knowledge of the codebase.

As such, we should make sure to manually test all key features, as we may not be able to completely trust our unit tests.

In the future, we should write e2e tests to help with large migrations such as this.

PR #2040, opened 2024-06-25
chore: handle `children` prop changes for React 18

colbr:
### What does this do?

React 18 (#2039) removes `children` as an intrinsic prop. It needs to be added to a component's interface explicitly.

- Silences type errors caused by external dependencies (`zUI` and `zOS-component-library`).
- Resolves type errors caused by zOS components (adding `children` to their props as needed).

I demarcated all silenced errors with `See: ZOS-115`, which is [the Linear issue for properly handling these errors](https://linear.app/zero-tech/issue/ZOS-115/fix-residual-silenced-type-issues-from-react-18).

So we don't hold up the React 18 migration while updating dependencies.

### DIFF
--- a/src/components/app-bar/more-apps-modal/index.tsx
+++ b/src/components/app-bar/more-apps-modal/index.tsx
@@ -36,8 +36,6 @@ export class MoreAppsModal extends React.Component<Properties> {
 
   render() {
     return (
-      // See: ZOS-115
-      // @ts-ignore
       <Modal title='Explore ZERO Apps' onClose={this.close}>
         <div {...cn()}>
           <div {...cn('header')}>Explore ZERO's suite of dApps, soon to be unified into the ZERO Operating System:</div>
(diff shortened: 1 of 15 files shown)


src/components/app-bar/more-apps-modal/index.tsx
```diff
@@ -36,6 +36,8 @@ export class MoreAppsModal extends React.Component<Properties> {
 
   render() {
     return (
+      // See: ZOS-115
```
ratik21, line 39: could you add a link to the issue as well (only for this comment)

ratik21 (approved):

src/components/chat-view-container/chat-view.tsx
```diff
@@ -274,6 +274,8 @@ export class ChatView extends React.Component<Properties, State> {
             onClose={this.closeLightBox}
           />
         )}
+        {/* See: ZOS-115
+         * @ts-ignore */}
```
dalefukami, line 278: I'm a little confused here. I thought adding children to the pros was supposed to remove the need for the type ignore?

colbr: Yeah you're right - I left those there by accident. Thanks for catching that!

PR #2105, opened 2024-07-17
feat: download and decrypt images only when conversation is active

domw30:
Note :: waypoint could be improved as to when onEnter is triggered if it is a conversation with very few messages, but not a blocker here.

### What does this do?
- The implementation and refactoring for downloading and decrypting images when conversation is active.

-  Prevent download/decrypt happening on initial load for every conversation. 

### How do I test this?
- Run test as usual.
- Run UI and check images media rendering for active conversations. 

#### Things to consider:
  1. Does this change any APIs?

https://github.com/user-attachments/assets/10d95bd9-69c6-41b9-adfd-3e86e4deb39f

### DIFF
--- a/src/store/messages/saga.ts
+++ b/src/store/messages/saga.ts
@@ -602,60 +602,48 @@ function* readReceiptReceived({ payload }) {
 }
 
 const inProgress = {};
-let count = 0;
 export function* loadAttachmentDetails(action) {
-  const { media, messageId } = action.payload;
+  const { media } = action.payload;
 
   console.log('XXX 1: Get Payload', action.payload);
 
-  count++;
-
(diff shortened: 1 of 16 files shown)


src/store/messages/saga.ts
```diff
@@ -599,3 +600,36 @@ function* readReceiptReceived({ payload }) {
...
+    if (!url) {
+      return;
+    }
+
+    yield put(
+      receiveMessage({
+        id: messageId,
+        media: { ...media, url },
+        image: { ...media, url },
+        rootMessageId: media?.info?.rootMessageId || '',
```
domw30, line 627: We shouldn't really need to set root message id here

PR #2122, opened 2024-07-25
feat(matrix-client/channels-saga): add room label management

domw30:
### What does this do?
- This change adds room label management using the RoomLabels enum and updates the getRoomTags function. 
- For now, we will filter out favorite and muted tags, ensuring the labels property only includes valid labels. I will refactor this in follow up.

- This change improves the organization and management of room labels by standardizing label handling, enhancing code clarity and maintainability.

### How do I test this?
- run tests as usual.
- no ui to run at this stage.

#### Things to consider:
  1. Does this change any APIs?

Uploading labels.mov…

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -22,7 +22,7 @@ import {
 } from 'matrix-js-sdk';
 import { RealtimeChatEvents, IChatClient } from './';
 import { mapEventToAdminMessage, mapMatrixMessage, mapToLiveRoomEvent } from './matrix/chat-message';
-import { ConversationStatus, Channel, User as UserModel } from '../../store/channels';
+import { ConversationStatus, Channel, User as UserModel, RoomLabels } from '../../store/channels';
 import { EditMessageOptions, Message, MessagesResponse } from '../../store/messages';
 import { FileUploadResult } from '../../store/messages/saga';
 import { ParentMessage, PowerLevels, User } from './types';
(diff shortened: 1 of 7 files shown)


src/lib/chat/matrix-client.ts
```diff
@@ -856,14 +856,31 @@ export class MatrixClient implements IChatClient {
     const tags = conversations.map(async (conversation) => {
       featureFlags.enableTimerLogs && console.time(`xxxgetRoomTags${conversation.id}`);
       const result = await this.matrix.getRoomTags(conversation.id);
+
+      // should move this to saga level
```
domw30, line 860: Will refactor this in a follow up

src/lib/chat/matrix-client.ts
```diff
@@ -856,14 +856,31 @@ export class MatrixClient implements IChatClient {
...
       conversation.isFavorite = !!result.tags?.[MatrixConstants.FAVORITE];
       conversation.isMuted = !!result.tags?.[MatrixConstants.MUTE];
+
+      const filteredLabels = Object.keys(result.tags).filter(
```
domw30, line 864: once refactored this filtering won't be required

src/store/channels/index.ts
```diff
@@ -35,6 +35,13 @@ export enum ConversationStatus {
   ERROR,
 }
 
+export enum RoomLabels {
+  WORK = 'm.work',
+  FAMILY = 'm.family',
+  SOCIAL = 'm.social',
```
dalefukami, line 41: Why are we forcing users to adhere to a restricted set of labels?

domw30: Yeah if we want to allow the user to create custom labels we shouldn't force them to adhere to a restricted set. Was only planning for requirements now rather than future. Removed restriction & renamed enum to reflect defaults: https://github.com/zer0-os/zOS/pull/2125

PR #2128, opened 2024-07-26
feat: app switcher

colbr:
### What does this do?

- Creates a top-level `App` component, which houses all "global UI" (providers, app bar), and renders the `AppSwitcher`.
- Creates an `AppRouter`, which is basically just a `Router` which renders different apps to a container based on `Route`.
- Moves all "global UI" from `Main` to `App`.
- Adds a feature flag (`enableExplorer`) for testing the app switcher.

So we can add arbitrary "sub apps" to zOS, as per the initial vision of the product.

### How do I test this?

- These changes should deliver 0 perceivable difference to the user. So, all flows should work exactly the same as before.
- Run `FEATURE_FLAGS.enableExplorer = true` in your browser console, and click the `Explorer` option in the app bar. Note: you will have to hit `Back` in your browser to go back to Messenger

---

Please note:
- I used function components because both of these components are very small and don't contain much business logic. I'm happy to rewrite them as class components if need be, 
- I used `@testing-library/react` for unit tests, as we are meant to be moving away from Enzyme.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -33,17 +33,14 @@ export const App = () => {
 };
 
 const useAppMain = () => {
-  const { isAuthenticated, background } = useSelector((state: RootState) => ({
-    isAuthenticated: !!state.authentication.user?.data,
-    background: state.background.selectedMainBackground,
-  }));
+  const isAuthenticated = useSelector((state: RootState) => !!state.authentication.user?.data);
+  const background = useSelector((state: RootState) => state.background.selectedMainBackground);
+  const videoBackgroundSrc = getMainBackgroundVideoSrc(background);
 
   const mainClassName = classNames('main', 'messenger-full-screen', getMainBackgroundClass(background), {
     'sidekick-panel-open': isAuthenticated,
     background: isAuthenticated,
   });
-
-  const videoBackgroundSrc = getMainBackgroundVideoSrc(background);
 
   return {
     isAuthenticated,
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,6 +7,7 @@ import { RootState } from './store';
 import classNames from 'classnames';
 import { getMainBackgroundClass } from './utils';
 import { AppBar } from './components/app-bar';
+import { DialogManager } from './components/dialog-manager/container';
 
 export const App = () => {
   const { mainClassName } = useAppMain();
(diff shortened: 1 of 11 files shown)


src/apps/app-router.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { Redirect, Route, Switch } from 'react-router-dom';
+import { MessengerMain } from '../messenger-main';
+import { FeatureFlag } from '../components/feature-flag';
+import { featureFlags } from '../lib/feature-flags';
+
+const redirectToRoot = () => <Redirect to={'/'} />;
+
+export const AppRouter = () => {
+  return (
+    <Switch>
```
colbr, line 16: I'm using `react-router-dom` to decide which app to render. IIRC this was handled in state previously, but I can't remember how. Do we need to keep the selected app in state? If yes, I can work that out, but it currently doesn't seem like there's a need for it.

src/Main.tsx
```diff
@@ -68,8 +36,7 @@ export class Container extends React.Component<Properties> {
             </FeatureFlag>
```
colbr, line 36: There's some additional clean-up I'd like to do here. I would like to move all of the children to `messenger-main.tsx`, and delete this file. In my opinion, it's just creating clutter now 🧹

I would also like to move `messenger-main.tsx` to `apps/messenger/index.tsx`. In general, I think storing all app entry points in `apps/***` will keep things tidy.

I can do this in a follow-up PR if y'all agree with the idea 🕵️‍♀️

src/App.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { AppRouter } from './apps/app-router';
+import { ZUIProvider } from '@zero-tech/zui/ZUIProvider';
+import { RootState } from './store';
+import classNames from 'classnames';
+import { getMainBackgroundClass, getMainBackgroundVideoSrc } from './utils';
+import { AppBar } from './components/app-bar';
+import { DialogManager } from './components/dialog-manager/container';
+import { ThemeEngine } from './components/theme-engine';
+
+export const App = () => {
```
colbr, line 13: Function components as these components are very simple. But, happy to change to class components if need be.

domw30 (approved):

PR #2156, opened 2024-08-10
fix: error handling for link preview

domw30:
### What does this do?
- adds error handling for link preview

- resolve error thrown when there is a bad preview url

### How do I test this?
- run tests as usual.
- run ui and enter a bad url

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/message/index.tsx
+++ b/src/components/message/index.tsx
@@ -420,7 +420,16 @@ export class Message extends React.Component<Properties, State> {
 
   renderLinkPreview() {
     const { preview, hidePreview, media, parentMessageText } = this.props;
-    if (!preview || hidePreview || media || parentMessageText) {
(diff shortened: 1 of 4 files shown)


src/components/message/index.tsx
```diff
@@ -420,7 +420,16 @@ export class Message extends React.Component<Properties, State> {
 
   renderLinkPreview() {
     const { preview, hidePreview, media, parentMessageText } = this.props;
-    if (!preview || hidePreview || media || parentMessageText) {
+    console.log('XXXpreview', preview);
+    if (
+      !preview?.title ||
```
domw30, line 425: The preview will always have a url as it is extracted from the message that is typed and sent. Therefore I have added to checks for the other required fields.

PR #2160, opened 2024-08-12
refactor: move Messenger files to `apps/messenger/`

colbr:
This is a very minor change.

- Clears up the `src` directory.
- Is consistent with putting all apps in `apps/` - this lets us keep all app entry-points in the same place.


domw30 (commented):
👍

PR #2215, opened 2024-08-28
feat: move conversation header higher up the DOM tree

colbr:
Header now spans the full length of the Messenger container


domw30 (approved):
Looks good mate, nice one for sorting that, not the easiest of components to lift up 🥵 💯

PR #2248, opened 2024-09-09
feat(post-input): add emoji picker and handle sending emojis

domw30:
### What does this do?
- adds emoji picker and handles sending emojis

- improve post input with emoji feature

### How do I test this?
- run tests as usual.
- run ui > open/create social channel > click emoji icon > select emoji from picker > post the 'post' as per demo below.

#### Things to consider:
  1. Does this change any APIs?

https://github.com/user-attachments/assets/afc053e5-d413-4fdb-aa85-5a16d509ffe5

### DIFF
--- a/src/components/messenger/feed/components/post-input/index.tsx
+++ b/src/components/messenger/feed/components/post-input/index.tsx
@@ -4,9 +4,11 @@ import Dropzone from 'react-dropzone';
 import { config } from '../../../../../config';
 import { Key } from '../../../../../lib/keyboard-search';
 import { motion, AnimatePresence } from 'framer-motion';
-import { Avatar, Button } from '@zero-tech/zui/components';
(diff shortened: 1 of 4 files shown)


src/components/messenger/feed/components/post-input/index.tsx
```diff
@@ -15,12 +17,14 @@ import './styles.scss';
 import Menu from '../../../../message-input/menu/menu';
 import { MediaType } from '../../../../../store/messages';
 import { Media, addImagePreview, dropzoneToMedia, windowClipboard } from '../../../../message-input/utils';
+import { EmojiPicker } from '../../../../message-input/emoji-picker/emoji-picker';
```
domw30, line 20: When we have updated posts implementation i.e. irys, we can work on refactoring the locations of these imports

PR #2290, opened 2024-09-20
feat(rewards): add meow transfer functionality with error handling and state management updates

domw30:
### What does this do?
- This adds functionality to handle MEOW token 'transfers' between users, including API integration and state management in the frontend.

- We are implementing the ability for users to transfer MEOW tokens as part of the app's reward and interaction system.

### How do I test this?
- run tests as usual.
- Note:: this is not yet wired up to UI

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/store/rewards/saga.ts
+++ b/src/store/rewards/saga.ts
@@ -124,6 +124,8 @@ export function* closeRewardsTooltip() {
 }
 
 export function* transferMeow(action) {
+  yield put(setTransferError({ error: null }));
+
   const { meowSenderUserId, meowRecipientUserId, amount } = action.payload;
 
   if (meowSenderUserId === meowRecipientUserId) {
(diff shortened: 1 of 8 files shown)


src/store/rewards/saga.ts
```diff
@@ -116,6 +122,22 @@ export function* closeRewardsTooltip() {
...
 }
 
+export function* transferMeow(action) {
+  const { senderUserId, recipientUserId, amount } = action.payload;
+
+  try {
+    const result = yield call(transferMeowAPI, senderUserId, recipientUserId, amount);
+
+    if (result.success) {
+      yield put(setMeow(result.response.senderBalance));
```
domw30, line 132: If we set Meow, this should handle live updates of the users total rewards. Then when rewards are fetched again, i.e. on reload, the same amount of meow should be displayed based on the db updates when hitting transferMeowApi

PR #2339, opened 2024-10-07
feat(messages): add download status state to media object to prevent repeated decryption attempts on re-renders

domw30:
### What does this do?
- This change adds a downloadStatus to the media object to prevent repeated decryption attempts during re-renders.

- We are making this change to improve performance and avoid unnecessary processing by ensuring that failed or already processed media files are not re-attempted for decryption.

### How do I test this?
- run tests as usual
- run UI and check redux state - ensure that loadAttachmentDetails is not run on re-render

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/store/messages/saga.ts
+++ b/src/store/messages/saga.ts
@@ -1,7 +1,16 @@
 import { currentUserSelector } from './../authentication/saga';
 import getDeepProperty from 'lodash.get';
 import { takeLatest, put, call, select, delay, spawn, takeEvery } from 'redux-saga/effects';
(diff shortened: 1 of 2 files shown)


src/store/messages/saga.ts
```diff
@@ -615,31 +624,41 @@ const inProgress = {};
 export function* loadAttachmentDetails(action) {
   const { media, messageId } = action.payload;
 
-  if (inProgress[messageId] || media.url) {
+  // Check conditions to skip processing
+  if (inProgress[messageId] || media.url || media.downloadStatus === MediaDownloadStatus.Failed) {
```
domw30, line 628: In follow up PR, we won't even call this `loadAttachmentDetails` if status === failed, but adding check here to double up on handling the errors

PR #2351, opened 2024-10-11
feat(media): update media handling for unencrypted rooms

domw30:
### What does this do?
- updates media (payload) handling for unencrypted rooms by ensuring Matrix URLs are processed and skipping unnecessary file decryption for non-encrypted media.

- payload update to work when receiving and sending to mobile devices, as a result some updates here required to handle the changes.

### How do I test this?
- run tests as usual.
- run UI and send images in encrypted/non-encrypted rooms.

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/store/messages/saga.ts
+++ b/src/store/messages/saga.ts
@@ -624,8 +624,11 @@ const inProgress = {};
 export function* loadAttachmentDetails(action) {
   const { media, messageId } = action.payload;
 
-  // Check conditions to skip processing
-  if (inProgress[messageId] || media.url || media.downloadStatus === MediaDownloadStatus.Failed) {
+  if (
+    inProgress[messageId] ||
+    (media.url && !media.url.startsWith('mxc://')) ||
+    media.downloadStatus === MediaDownloadStatus.Failed
+  ) {
     return;
   }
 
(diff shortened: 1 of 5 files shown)


src/store/messages/saga.ts
```diff
@@ -624,8 +624,11 @@ const inProgress = {};
 export function* loadAttachmentDetails(action) {
   const { media, messageId } = action.payload;
 
-  // Check conditions to skip processing
-  if (inProgress[messageId] || media.url || media.downloadStatus === MediaDownloadStatus.Failed) {
+  if (
+    inProgress[messageId] ||
+    (media.url && !media.url.startsWith('mxc://')) ||
```
ratik21, line 629: I am assuming, when you download the media, you're considering that the media url is updated from `mxc://` -> `blob://..`? Which is why you've added the `mxc://` condition here?

ratik21 (approved):

src/store/messages/saga.ts
domw30: Yeah thats it. The matrix url is transformed into a blob, so this will exit early and not process again

PR #2638, opened 2025-02-09
fix: handle `undefined` `firstName`

colbr:
### What does this do?

- Adds handling for `undefined` `firstName` in `sortMembers`.
- Members with `undefined` `firstName` are sorted to the top of the list.
- Adds tests for the above case.

Note: I'm not sure why `firstName` could ever be `undefined`.


ratik21:
> Note: I'm not sure why firstName could ever be undefined.

Yeah, I am not sure either. Must be some user the bridge hasn't been able to parse specifically. Doesn't happen for me though.

PR #2681, opened 2025-02-20
fix: unable to mark channel chat notifications as read

colbr:
### What does this do?

- Moves App Bar notification count state to App Bar (prevents re-rendering the whole app).
- Adds `isFeedApp` to "mark as read" conditional in Matrix client.
- Calls "mark as read" whenever a chat is validated (i.e. when it's opened).


domw30 (approved):
Reviewed & tested. Looking good 💪 Shall merge and deploy!

PR #2693, opened 2025-02-24
Trigger `recordMessageSent` with `type == 'social'` in a social (token gated) channel

ratik21:
### What does this do?

Updates the `sendMessage` logic to also include the channel type (if it is a social channel or not). If yes, then we pass on this info to `recordMessageSent` fn, which calls the respective API to record that event.

This is done so that in the rewards calculation, we can filter out the users who have sent messages in social channels.


domw30 (commented):
All looks good to me!

PR #2750, opened 2025-03-06
Proposal for zApp communication

RickyRoller:
### What does this do?

This change introduces a communication system between zOS and external applications (zApps) using MessageChannel. The system allows third-party applications to be embedded within zOS while maintaining strict control over their capabilities and communication.
Key features include:
- Secure message passing between zOS and external apps using MessageChannel
- Manifest-based feature declaration for external apps
- Whitelisted app authentication with token management
- Configurable UI integration options through `features` (e.g. fullscreen mode)

We have a new app being built outside of zOS that'll need new capabilities

### How do I test this?
Currently no apps in prod take advantage of this. The readme has instructions for interacting with this system if you want to setup a simple app.

#### Things to consider:
    - Sharing the original access token across apps isn't an appropriate solution for apps that are written by third parties, so we'll need to use a strict whitelist for what apps can use this functionality. In the future we can implement a new token system where apps can request certain permissions and the user can agree to those permissions.


domw30 (commented):
I think this is a good approach. The manifest concept provides a good foundation for future expansion while addressing immediate needs for AURA.

The authentication approach makes sense as a starting point, though I agree a proper scopes/claims system would be better long-term. For now, passing the access token to trusted apps is pragmatic.

The full-viewport iframe feature will definitely help make integrated apps feel more native within zOS.

RickyRoller:
I decided that it'd be best to keep the manifest inside of zOS for now instead of having the app pass it in. Things are still very coupled at this point so no reason to complicate it.



There is one more change to make around getting the token and passing it, but getting close to having this ready

src/apps/external-app/index.tsx
```diff
@@ -7,65 +7,118 @@
...
+          port: channel.port2,
+        },
+        { targetOrigin: event.origin, transfer: [channel.port2] }
+      );
+    }
+  };
+
+  onMessage = (event: MessageEvent<IncomingMessage>) => {
+    // Legacy message handler for backwards compatibility
+    // Remove this once we're sure all apps have updated to the new MessageChannel handler
```
domw30, line 98: 👍

domw30 (approved):

PR #2784, opened 2025-03-13
feat: Add new functionality to feature flag system

RickyRoller:
### What does this do?

This adds a few new features to the feature flag system and makes it a bit easier to maintain.
- Flags can now be controlled separately in development vs production
- Flags can be locked, preventing tampering in the console
- Flags can be restricted to individual users
  - Flags with this restriction are considered locked
- Flags can be reset to their defaults with `FEATURES_FLAGS.reset()`

This should allow us more control over rolling out new features.
For instance, we can enable something that only the team can test out internally for a while before going to staging, and then production.

Locking flag
```
{ enableFeedChat: { defaultValue: true, locked: true } }
```
Screenshot 2025-03-13 at 3 44 08 PM

Enabled for specific users
```
{ enableFeedChat: { defaultValue: true, allowedUserIds: ['12345'] } }
```
### How do I test this?

I kept all of the flags with their current value so everything should work the same.

### DIFF
--- a/src/lib/feature-flags/flags.ts
+++ b/src/lib/feature-flags/flags.ts
@@ -1,5 +1,4 @@
-import { store } from '../../store';
-import { currentUserSelector } from '../../store/authentication/saga';
+import window from 'global/window';
 
 export type FeatureFlagKey =
   | 'channelsApp'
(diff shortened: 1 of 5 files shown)


src/lib/feature-flags/flags.ts
```diff
@@ -0,0 +1,87 @@
...
+let getCurrentUserId = () => null;
+export const setGetCurrentUserId = (fn: () => string | null) => {
+  getCurrentUserId = fn;
+};
+if (process.env.NODE_ENV !== 'test') {
+  (async () => {
+    const { store } = await import('../../store');
+    getCurrentUserId = () => store.getState().authentication?.user?.data?.id ?? null;
+  })();
+}
```
RickyRoller, line 50: We don't have any tests around feature flags themselves (outside of the FeatureFlag component), so I'm bypassing this saga dependency when we're in tests.
This means we won't be able to write tests around `allowedUserIds` as the current user id will always be null, but that seems like a reasonable tradeoff instead of mocking this in all tests that use it.

src/lib/feature-flags/flags.ts
```diff
@@ -0,0 +1,87 @@
...
+  | 'enableMeows'
+  | 'enableTokenGatedChat'
+  | 'enableNotificationsApp'
+  | 'enableNotificationsReadStatus'
+  | 'enableLoadMore'
+  | 'enableComments'
+  | 'enableFeedApp'
+  | 'enableLinkedAccounts'
+  | 'enableZeroWalletSigning'
+  | 'enableFeedChat';
```
RickyRoller, line 26: I'd be happy to clean up any of these if we no longer need them

src/lib/feature-flags/flags.ts
RickyRoller: For reference, there were 50 failed tests due to pulling in the store

src/lib/feature-flags/flags.ts
domw30: Yeah we could definitely do with cleaning these up, most are no longer required

domw30 (approved):
Nice improvements, looks good to me

PR #2792, opened 2025-03-14
Add Aura zApp

RickyRoller:
### What does this do?

Adds the Aura zApp to zOS for internal testing


domw30 (approved):
Looking good! Nice work 🚀

PR #2859, opened 2025-03-25
Refactor and optimize message component

RickyRoller:
### What does this do?

Handful of changes in this one in order to get it out.

Refactored the message component. It was very hard to read and maintain, plus it was using `any` all over the place.
I've broken out a lot of the render methods into their own components and cleaned up the typing.

Also, calling the cloudinary provider for the avatar url inline was causing some extra memory usage, so that's been moved to a memo.

There was a bug where we were calling a saga function in the main render path.

The tests ended up being a decent rewrite. A lot of them are no longer testable for various reasons, but we should be in a fairly comparable position.

### DIFF
--- a/src/components/message/index.tsx
+++ b/src/components/message/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useMemo } from 'react';
+import React, { useState } from 'react';
 import { createPortal } from 'react-dom';
 import classNames from 'classnames';
 import {
@@ -9,7 +9,6 @@ import {
   MessageAttachment,
 } from '../../store/messages';
 import { LinkPreview } from '../link-preview';
-import { getProvider } from '../../lib/cloudinary/provider';
 import { MessageInput } from '../message-input/container';
 import { MessagesFetchState, User } from '../../store/channels';
 import { ParentMessage as ParentMessageType } from '../../lib/chat/types';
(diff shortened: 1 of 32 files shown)


src/components/message/index.tsx
```diff
@@ -59,85 +58,82 @@ interface Properties extends MessageModel {
...
+  const [isMessageMenuOpen, setIsMessageMenuOpen] = useState(false);
+  const [isAttachmentPreviewOpen, setIsAttachmentPreviewOpen] = useState(false);
+  const [attachmentPreview, setAttachmentPreview] = useState(null);
+  const [isReactionPickerOpen, setIsReactionPickerOpen] = useState(false);
+  const [isDropdownMenuOpen, setIsDropdownMenuOpen] = useState(false);
+
+  const isMenuTriggerAlwaysVisible = sendStatus === MessageSendStatus.FAILED;
+
+  const senderAvatarUrl = useMemo(() => {
+    return getProvider().getSourceUrl(sender.profileImage);
```
RickyRoller, line 120: @ratik21 you mentioned that we shouldn't be getting avatars this way. Can you point me to another usage where we're getting the avatar from the correct location?

ratik21: as discussed on call, i think we can just remove this.

domw30 (approved):
Really nice refactoring here. I think this is good to merge in. If we spot any issues that we may have missed we can address.

PR #2864, opened 2025-03-26
Migrating to latest Matrix sdk version using rust crypto

RickyRoller:
### What does this do?

Upgrades us to the latest Matrix SDK version and integrates the new rust crypto methods (using wasm bindings)

This should unblock us implementing sliding sync

### How do I test this?

Pull the branch, and load it up.
It'll take an extra second to load conversations as it'll be running a migration on your local crypto store, but I ran it on prod with 23k migrations and it only took a few seconds

#### Things to consider:
  1. Does this change any APIs?


src/lib/chat/matrix-client.ts
```diff
@@ -1800,47 +1810,6 @@ export class MatrixClient implements IChatClient {
...
-    const members = (await room?.getEncryptionTargetMembers()) || [];
-    if (members.length > 0) {
-      await this.matrix.crypto?.ensureOlmSessionsForUsers(
-        members.map((m) => m.userId),
-        true
-      );
-    }
-    this.debug('done resetting the olm session', roomId);
-  }
-
```
RickyRoller, line 1843: These were all using removed methods, so you'll see them pulled out throughout the codebase.
They were just used for a dev panel, so shouldn't affect any users.

src/lib/chat/matrix-client.ts
```diff
@@ -1064,13 +1066,13 @@ export class MatrixClient implements IChatClient {
...
-    await this.matrix.setAccountData(MatrixConstants.READ_RECEIPT_PREFERENCE, { readReceipts: preference });
+    localStorage.setItem(MatrixConstants.READ_RECEIPT_PREFERENCE, preference);
   }
 
   async getReadReceiptPreference() {
     try {
-      const accountData = await this.matrix.getAccountData(MatrixConstants.READ_RECEIPT_PREFERENCE);
-      return accountData?.event?.content?.readReceipts || ReadReceiptPreferenceType.Private;
+      const readReceiptPreference = localStorage.getItem(MatrixConstants.READ_RECEIPT_PREFERENCE);
+      return readReceiptPreference || ReadReceiptPreferenceType.Private;
```
RickyRoller, line 1075: This wasn't an official setting and isn't supported in storing in the account data anymore, so I switched to just store it locally instead.

src/lib/chat/matrix-client.ts
domw30: yeah correct this should be fine

domw30 (commented):
💪 looks like tests need a couple of tweaks but nothing major? I'll keep an eye out for changes 🤝

domw30 (approved):
Hey mate, spent quite a bit of time testing things locally for this. All looks stable to me. I can't see anything in the code that looks out of place or doesn't make sense. As far as I can tell the changes are pretty straight forward. I think it would be good for us to merge and carry out some further testing on staging.

src/lib/chat/matrix-client.ts
```diff
@@ -1435,20 +1427,25 @@ export class MatrixClient implements IChatClient {
...
       const opts: any = {
+        cryptoStore,
         baseUrl: config.matrix.homeServerUrl,
         cryptoCallbacks: { getSecretStorageKey: this.getSecretStorageKey },
-        ...(await this.getCredentials(userId, ssoToken)),
+        ...userCreds,
       };
 
       this.matrix = this.sdk.createClient(opts);
+      await this.matrix.initRustCrypto();
```
ratik21, line 1445: 🎉

ratik21 (approved):
Looks good. 

+ I think this needs to be tested pretty heavily before merging (creating multiple backups on different devices, with different sessions, restoring them ..etc). 

+ Just to confirm, my current backup key will still work after this goes into production right?

RickyRoller:
> Looks good.
> 
> * I think this needs to be tested pretty heavily before merging (creating multiple backups on different devices, with different sessions, restoring them ..etc).
> * Just to confirm, my current backup key will still work after this goes into production right?

Yes, your current key will work.

PR #2865, opened 2025-03-27
feat: wip revoke image blob urls

domw30:
#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/message/hooks/useLoadAttachmentEffect.ts
+++ b/src/components/message/hooks/useLoadAttachmentEffect.ts
@@ -1,18 +1,22 @@
 import { useEffect } from 'react';
 import { MessagesFetchState } from '../../../store/channels';
-import { MediaDownloadStatus } from '../../../store/messages';
+import { Media, MediaDownloadStatus } from '../../../store/messages';
 
 export const useLoadAttachmentEffect = (
-  media: any,
+  media: Media,
   messageId: string,
   loadAttachmentDetails: (payload: { media: any; messageId: string }) => void,
   messagesFetchStatus: MessagesFetchState
 ) => {
   useEffect(() => {
+    // Images are handled by the useMatrixImage hook
+    const isImage = media?.mimetype?.startsWith('image/');
+
     const isLoading = media?.downloadStatus === MediaDownloadStatus.Loading;
     const hasFailed = media?.downloadStatus === MediaDownloadStatus.Failed;
     if (
       media &&
+      !isImage &&
       (!media.url || media.url.startsWith('mxc://')) &&
       !isLoading &&
       !hasFailed &&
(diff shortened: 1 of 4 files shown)


src/components/message/hooks/useLoadAttachmentEffect.ts
```diff
@@ -1,18 +1,22 @@
...
 ) => {
   useEffect(() => {
+    // Images are handled by the useMatrixImage hook
+    const isImage = media?.mimetype?.startsWith('image/');
+
     const isLoading = media?.downloadStatus === MediaDownloadStatus.Loading;
     const hasFailed = media?.downloadStatus === MediaDownloadStatus.Failed;
     if (
       media &&
+      !isImage &&
```
RickyRoller, line 19: This hook modifies the media object, which we don't want since we need the original mxc url in order to manage the blobs.

I actually think we should kill this whole thing eventually and move all media management to react query

src/lib/hooks/useBlobMemoryManagement.ts
```diff
@@ -0,0 +1,151 @@
...
+  const timeoutRef = useRef<number | null>(null);
+  const [isUrlRevoked, setIsUrlRevoked] = useState(false);
+  const originalUrlRef = useRef<string | undefined>(url);
+
+  useEffect(() => {
+    originalUrlRef.current = url;
+  }, [url]);
+
+  // Use the Matrix image hook for Matrix URLs
+  const { data: effectiveUrl, isPending: isMatrixLoading } = useMatrixImage(url, {
```
RickyRoller, line 39: isPending is recommended over isLoading for most cases
https://github.com/TanStack/query/discussions/6297#discussioncomment-7467010

src/lib/hooks/useBlobMemoryManagement.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  useEffect(() => {
+    console.log(`xxxx [Memory Management] URL changed to ${url?.substring(0, 30)}...`);
+    setIsUrlRevoked(false);
+  }, [url]);
+
+  useEffect(() => {
+    if (isInView && isUrlRevoked) {
+      console.log(`xxxx [Memory Management] URL was revoked, refreshing image for ${url?.substring(0, 30)}...`);
+      queryClient.invalidateQueries({ queryKey: ['matrix', 'file', { url, isThumbnail }] });
```
RickyRoller, line 51: The counter that was in place was changing the query key, which meant the useQuery cache would keep the old image url around and make a new one.

Instead, we can just invalidate the query and that'll cause it to refetch, which will get us an updated url.

An alternative is that the useQuery returns a `refetch` method you can use.

PR #2871, opened 2025-03-28
feat: revoke blob urls when changing conversations

domw30:
### What does this do?
- We're improving the useMatrixImage hook to properly clean up blob URLs when components unmount and setting React Query's garbage collection time to zero.

- To prevent memory leaks during long sessions by ensuring blob URLs are properly revoked and removed from React Query's cache when they're no longer needed.

### How do I test this?
- run tests as usual
- run ui and switch conversations > check image blob urls are revoked and then updated when switching back to the original conversation you were on.
- test sending image messages and receiving image messages

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/message/hooks.test.tsx
+++ b/src/components/message/hooks.test.tsx
@@ -12,7 +12,7 @@ import { useContextMenu } from './hooks/useContextMenu';
 describe('loadAttachmentEffect', () => {
   it('calls loadAttachmentDetails if no media url and messagesFetchStatus is success', () => {
     const loadAttachmentDetails = jest.fn();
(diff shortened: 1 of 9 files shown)


domw30:
@RickyRoller - fixed up the PR and marked ready for review. Not sure if we even want to merge this as we discussed, re the urls already being put into indexdb and browser cache, and then it's getting cleaned up on a schedule already. Could do with a review/manual testing on this if we do decide to go forward and merge. Looks good for me, but worth checking light box, loading & other forms of media such as video/files.

RickyRoller (commented):
@domw30 Agreed, your other 2 PRs look like they'll make this one a bit obsolete.
We can monitor and see if we want to turn on the gcTime in the useMedia hook after that's in

PR #2878, opened 2025-04-01
refactor(components): migrate components to new media loading system using the useMatrixMedia hook

domw30:
### What does this do?
- Updates all components to use the new useMatrixMedia hook for handling media, including MessageMedia, ParentMessage, ReplyCard, Lightbox, and related components, ensuring consistent media handling across the application.

- To standardize media handling across components, improve type safety, and leverage the built-in caching provided by the new hook, while preparing for the removal of the legacy media loading system.

### How do I test this?
- run tests as usual
- run UI and render media in conversations

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/components/chat-view-container/chat-view-container.tsx
+++ b/src/components/chat-view-container/chat-view-container.tsx
@@ -8,8 +8,6 @@ import {
   editMessage,
   Message,
   EditMessageOptions,
-  loadAttachmentDetails,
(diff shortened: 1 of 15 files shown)


RickyRoller (dismissed):
Nice clean up!
Just did some testing and this looks good to me

PR #2899, opened 2025-04-06
feat(maintenance-mode): add maintenance mode detection to show maintenance page when server is unavailable

domw30:
### What does this do?
We're adding a maintenance mode detection system that checks the server status and displays a maintenance page when the server is in maintenance mode.

- To provide users with clear feedback when the server is undergoing maintenance instead of showing cryptic errors or a blank screen.

### How do I test this?
- run tests as usual

#### Things to consider:
  1. Does this change any APIs?


domw30:
closing pull request - we will manually add the maintenance page when necessary

PR #2916, opened 2025-04-10
fix(matrix-client): prevent duplicate avatar updates by properly comparing MXC and blob URLs in profile sync

domw30:
### What does this do?
- We're modifying the URL comparison logic in verifyMatrixProfileIsSynced to properly handle both MXC and blob URLs when comparing avatars.

- We're fixing a bug where the system was incorrectly detecting avatar differences between blob and MXC URLs, causing unnecessary avatar update messages to be sent to every conversation on page refresh.

### How do I test this?
- run tests as usual
- run UI and check for avatar admin messages

#### Things to consider:
  1. Does this change any APIs?

### DIFF
--- a/src/lib/chat/matrix-client.test.ts
+++ b/src/lib/chat/matrix-client.test.ts
@@ -2014,6 +2014,116 @@ describe('matrix client', () => {
     consoleWarnSpy.mockRestore();
   });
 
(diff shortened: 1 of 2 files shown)


RickyRoller:
Fixed here: https://github.com/zer0-os/zOS/pull/2917

PR #2918, opened 2025-04-10
Fix avatar change message when joining a channel

RickyRoller:
### What does this do?

When joining a channel you have no `previousContent`, so it's always going to display an avatar change message.
Now checking for `previousContent` as well

### DIFF
--- a/src/lib/chat/matrix/chat-message.ts
+++ b/src/lib/chat/matrix/chat-message.ts
@@ -179,7 +179,7 @@ function getRoomReactionAdminData(content, sender) {
 }
 
 function getRoomMemberAdminData(content, targetUserId, previousContent?) {
-  if (previousContent && previousContent?.avatar_url !== content.avatar_url) {
+  if (previousContent?.avatar_url && previousContent?.avatar_url !== content.avatar_url) {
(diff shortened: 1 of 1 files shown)


src/lib/chat/matrix/chat-message.ts
```diff
@@ -179,7 +179,7 @@ function getRoomReactionAdminData(content, sender) {
 }
 
 function getRoomMemberAdminData(content, targetUserId, previousContent?) {
-  if (previousContent?.avatar_url !== content.avatar_url) {
+  if (previousContent?.avatar_url && previousContent?.avatar_url !== content.avatar_url) {
```
ratik21, line 182: why not add an empty check for `content.avatar_url` also? Well, i guess you can't do that from our UI 🤔

ratik21 (approved):

ratik21:
BDW, @RickyRoller I do get these often when i am "creating a new conversation" (and not just adding members to a group)
image

Is this related to this PR as well?

RickyRoller:
> BDW, @RickyRoller I do get these often when i am "creating a new conversation" (and not just adding members to a group) image
> 
> Is this related to this PR as well?

No, doesn't look like it.
This is what I see on creating a new conversation
Screenshot 2025-04-11 at 7 31 32 AM

PR #2986, opened 2025-04-29
Fix: First pass on strict null checks

RickyRoller:
### What does this do?
This is my first pass on enabling `strictNullChecks` in TypeScript.
There are still another 117 files that need to be updated before we can turn this on

With this flag turned off, we're completely ignoring null values. In this example, the function clearly states that it's return type is `Room | null` but TypeScript isn't checking that and says the value is `Room`. This means we have landmines all throughout the codebase where we think things are valid when they could be null.

Screenshot 2025-04-29 at 8 47 15 AM
Screenshot 2025-04-29 at 8 47 22 AM

It also allows you to null anything and it won't throw any errors.
For example, it doesn't see any issues with this code
Screenshot 2025-04-29 at 8 51 51 AM


src/components/messenger/list/utils/utils.tsx
```diff
@@ -1,18 +1,5 @@
-import { monthsSince, fromNow } from '../../../../lib/date';
 import { User } from '../../../../store/channels';
 
-export function lastSeenText(user): string {
```
RickyRoller, line 4: unused function

RickyRoller:
I'll revisit this and clean up the merge conflicts when I get some time next week

PR #2995, opened 2025-05-01
Overhaul message data flow from Matrix to Redux

RickyRoller:
### What does this do?
Moves the logic for updating optimistic messages from the timeline event over to the local echo update event.
In the process of fixing optimistic messages I noticed that we had a handful of race conditions that could break messages in the chat feed. After digging into it I landed on breaking the message flows into 4 distinct operations

**Update active channel messages on new message:**
Matrix message event -> Current active channel -> Sync room messages with channel

**Update messages on viewing channel:**
Open channel -> initial messages not loaded -> fetch messages
Open channel -> initial messages loaded -> sync room messages with channel

**Update channel last message:**
Matrix message event -> update channels last message
Sliding sync event -> update channels last message
Finish decrypting messages with recovery key -> update channels last message

**Optimistic message updating:**
Matrix local echo event -> Update optimistic version of message with full message object

In general, this changes the flow of redux manipulation to be more of a pull data flow instead of a push data flow.
Matrix events don't modify the redux store directly, they instead trigger sagas that sync with Matrix to get the latest data.
This removes issues from race conditions, removes reliance on the accuracy of the data in the event, and allows batching/debouncing since we don't need the data from the event.
It also ensures our data accuracy since we're not relying on events to populate redux

The timeline event was flakey on when it would fire so we were still seeing issues with optimistic messages updating to full message objects.

We had issues with race conditions due to multiple matrix events all triggering updates on the same data structures even though they had different functionality and accuracy.
For instance, new messages don't have the context of whether they've been edited since the edit event is a separate event. So if we replace a message in the chat view with a message from a matrix event, it won't have the context of the timeline and will remove/reset valuable fields.

### How do I test this?
Send a message and make sure you can edit it and send a reaction to it shortly after.
Receive messages and see last message update in the channel list
Receiving new message in the active channel should appear
When visiting a channel you should see the latest messages
Scrolling up in a channel should fetch the next set of messages


src/components/message/index.tsx
RickyRoller, line 1: Just adding useCallbacks to prevent extra renders

src/lib/chat/matrix/matrix-adapter.ts
```diff
@@ -15,6 +15,7 @@ export class MatrixAdapter {
     const createdAt = Matrix.client.getRoomCreatedAt(room);
     const labels = Object.keys(room.tags || {});
     const [admins, mods] = Matrix.client.getRoomAdminsAndMods(room);
+    const isSocialChannel = Matrix.client.getRoomGroupType(room) === 'social';
```
RickyRoller, line 18: Fix social channel flag not being present on initial sync

PR #3030, opened 2025-05-12
Fix: Optimize re-renders for messenger list

RickyRoller:
### What does this do?
Refactors the messenger list and it's sub-components to better structure redux data flow and to optimize for re-renders.

Small updates to a single prop in a redux object case every component to re-render.

### How do I test this?
Using the react-scan browser extension, make a change to a channel and check the updates in the list
| Before | After|
|---|---|
|  |  |

### Key decisions and Risk Assessment:
All tests have been re-written in RTL + vitest. This is because the shallow testing method has quite a lot of drawbacks and in general isn't very compatible with function components.
The tests were the largest re-write of all of this, so I'd appreciate any thoughts on how they're done.


domw30 (approved):
Nice work, lgtm!

PR #3044, opened 2025-05-16
Refactor messenger components to optimize for re-renders

RickyRoller:
### What does this do?
Refactors the main messenger / chat view components to optimize them for re-rendering.
Since we normalize data, I've added new selectors for handling converting the normalized data as needed. The current methods for denormalizing everything in one pass causes a new reference to be created on every render, which means we can't optimize the components.
All of these hooks have been added to a store/hooks folder

Any changes to any channel cause the current messenger components to be re-rendered.

### How do I test this?
Test out all of the regular messenger functionality, everything should work as expected.


src/components/message/index.tsx
```diff
@@ -375,7 +375,7 @@ export const Message: React.FC<Properties> = ({
   const renderBody = () => {
     return (
       <div {...cn('block-body')}>
-        {media?.body && (
```
RickyRoller, line 378: Fix for showing file names as the message body

PR #3046, opened 2025-05-19
Fix: New conversation state syncing

RickyRoller:
### What does this do?
3 Separate issues in here:
- Properly update state in redux for newly created conversations
  - Regular Matrix event handlers don't trigger for new conversations, so I've added state syncing to the sliding sync events to make sure we update the member list and the messages for these new conversations.
- Fix inviter conversation creation for new users
- Fix opening existing 1:1 conversation if it exists

The member list wasn't fully synced from Matrix to Redux when creating a new conversation.
The messages for the room weren't fully synced from Matrix to redux when creating a new conversation.
`any` TypeScript was hiding an issue with checking for existing 1:1
Welcome conversation wasn't using the updated method for loading user data and putting it in Redux

### How do I test this?
Create a new conversation, you should see all of the members in the side panel and you should see all of the "member was added to conversation" messages.
When trying to create a 1:1 with someone you already have a conversation with, it should open that conversation instead.
When creating a new user, they should have a newly created conversation with the person who invited them

### DIFF
--- a/src/lib/chat/matrix-client.ts
+++ b/src/lib/chat/matrix-client.ts
@@ -203,7 +203,9 @@ export class MatrixClient {
       return {
         otherMembers,
         memberHistory,
-        totalMembers: room.getInvitedAndJoinedMemberCount(),
+        // Even if a member leaves they stay in the member list so this will provide the correct count
+        // for determining if a conversation is 1 on 1
+        totalMembers: room.getMembers().length,
       };
     }
   }


src/lib/chat/matrix-client.ts
```diff
@@ -203,7 +203,9 @@ export class MatrixClient {
       return {
         otherMembers,
         memberHistory,
-        totalMembers: room.getInvitedAndJoinedMemberCount(),
+        // Even if a member leaves they stay in the member list so this will provide the correct count
+        // for determining if a conversation is 1 on 1
+        totalMembers: room.getMembers().length,
```
RickyRoller, line 208: This was a regression when implementing sliding sync. I've moved back to the old logic and added the comment from the old logic as well

src/store/registration/saga.test.ts
```diff
@@ -508,33 +507,30 @@ describe('authorizeAndCreateWeb3Account', () => {
...
+    initialState.withUsers(inviter);
 
-    await subject(createWelcomeConversation, 'new-user-id', { id: 'inviter-id', matrixId: 'inviter-matrix-id' })
-      .provide([[call(getZEROUsersAPI, ['inviter-matrix-id']), [{ userId: 'inviter-id', firstName: 'The inviter' }]]])
-      .withReducer(rootReducer, initialState.build())
-      .call(createConversation, ['inviter-id'], '', null)
-      .run();
-  });
-
-  it('stores the inviter in state', async () => {
```
RickyRoller, line 525: Removing this test as this is no longer responsible for adding user to Redux. That's handled by getUserByMatrixId and that has it's own tests

PR #3211, opened 2025-07-23
wallet: Small cleanup items in tokens list

RickyRoller:
### What does this do?
Better token skeleton
Better transaction skeleton
Better decimal handling for dollar values
Preloading token images to prevent pop in

### DIFF
--- a/src/apps/wallet/tokens/token-skeleton.tsx
+++ b/src/apps/wallet/tokens/token-skeleton.tsx
@@ -1,0 +1,22 @@
+import styles from './token-skeleton.module.scss';
+import { Skeleton } from '@zero-tech/zui/components';
+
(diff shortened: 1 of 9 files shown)


src/apps/wallet/tokens/token-skeleton.tsx
```diff
@@ -0,0 +1,22 @@
+import styles from './token-skeleton.module.scss';
+import { Skeleton } from '@zero-tech/zui/components';
+
+export const TokenSkeleton = () => {
+  return (
+    <div className={styles.tokenSkeleton}>
```
domw30, line 6: @colbr - guys shall we decide on camelcase / capitalize for classes? very minor but worth aligning

PR #3407, opened 2025-09-30
Fix some bugs with standalone image messages, and image w/ text messages

ratik21:
### What does this do?

Fixes the following bugs

1. There is a similar issue when sending images / different types of images messages, due to optimistic updates. When sending a regular image it sends, and then once it registers on the server it blinks. This should be invisible to the user.

2. When you send an image with a message, it completely disappears, and then re-renders once the server updates.

3. When you send an image or link preview, it adds it to the dom with one size, and then quickly resizes the message abruptly to fit the contents. It should just be the proper size with no change from when it is added.

https://github.com/user-attachments/assets/91e945e2-50d4-4298-b7c5-ebcf3f7926c2

### DIFF
--- a/src/store/messages/saga.ts
+++ b/src/store/messages/saga.ts
@@ -150,7 +150,7 @@ export function* fetchMessages(action) {
     yield call(receiveChannel, {
       id: channelId,
       messages,
-      hasMore: messagesResponse.hasMore,
+      hasMore: messagesResponse.hasMore && messages.length > 0,
       hasLoadedMessages: true,
       messagesFetchStatus: MessagesFetchState.SUCCESS,
     });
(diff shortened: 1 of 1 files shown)


src/store/messages/saga.ts
```diff
@@ -243,7 +244,32 @@ export function* createOptimisticMessage(channelId, message, parentMessage, file
...
+  let enrichedFile = file;
+  if (
+    file &&
+    file.mediaType === MediaType.Image &&
+    file.nativeFile &&
+    (file.width === undefined || file.height === undefined)
+  ) {
+    try {
+      const { width, height } = yield call(getImageDimensions, file.nativeFile);
+      enrichedFile = { ...file, width, height };
```
ratik21, line 256: We essentially copy the dimensions of the actual image in the optimistic message object, that the placeholder image has the same dimensions as the actual image, which prevents it from re-sizing after the file is uploaded

